PR #8, opened 2025-12-01
adding gemm overlap config for single node

oyazdanb:
This configuration replicates the issue we have for overlapping GEMM  computation with RCCL collectives.


IanNod (approved):
lgtm

PR #12, opened 2025-12-01
Gemm report analysis

oyazdanb:
This PR adds visualization, overlap analysis, and reporting tools for analyzing GEMM kernel performance data generated from sweep profiling.
**Changes**

Adds scripts/gemm_analysis/ visualization and reporting tools:
_Core Pipeline (Steps 5-7):_

- plot_gemm_variance.py - Generate comprehensive variance visualizations (box plots, violin plots, interaction plots)

- enhance_gemm_variance_with_timestamps.py - Add kernel execution timestamps from trace files

- gemm_report_with_collective_overlap.py - Identify NCCL collective operations overlapping with GEMM kernels

- create_embeded_html_report.py - Generate side-by-side HTML comparison reports for two sweeps

_Additional Analysis Tools:_

- process_gpu_timeline.py - Aggregate GPU timeline data across ranks into multi-sheet Excel reports

- process_comms.py - Extract and aggregate NCCL collective communication data into Excel/CSV

_Documentation:_

- README.md - Complete usage documentation for all visualization and reporting tools


scripts/gemm_analysis/create_embeded_html_report.py
```diff
@@ -0,0 +1,426 @@
...
+        b64 = image_to_base64(path)
+        if b64:
+            image_data[f'{key}_sweep2'] = f"data:image/png;base64,{b64}"
+            print(f"    [OK]")
+        else:
+            image_data[f'{key}_sweep2'] = ""
+            print(f"    [MISSING] {path}")
+
+    # Create HTML with embedded images
+    html_content = f"""<!DOCTYPE html>
```
IanNod: This block of html_content is taking up a large chunk of this script, can it be moved to some util file?

scripts/gemm_analysis/enhance_gemm_variance_with_timestamps.py
```diff
@@ -0,0 +1,270 @@
...
+
+    # Sort by duration
+    kernel_instances.sort(key=lambda x: x['duration_us'])
+
+    # Option 1: Use all instances (current behavior - RECOMMENDED)
+    # Get the actual minimum and maximum instances
+    min_instance = kernel_instances[0]  # Shortest duration
+    max_instance = kernel_instances[-1]  # Longest duration
+
+    # Option 2: Match TraceLens top 100 (uncomment to use)
```
IanNod: Better to enable through an input arg instead of relying on the user to hack at the code

scripts/gemm_analysis/enhance_gemm_variance_with_timestamps.py
```diff
@@ -0,0 +1,270 @@
...
+            'min_timestamp_ms': None,
+            'max_timestamp_ms': None,
+            'min_duration_found_us': None,
+            'max_duration_found_us': None
+        }
+
+    # Sort by duration
+    kernel_instances.sort(key=lambda x: x['duration_us'])
+
+    # Option 1: Use all instances (current behavior - RECOMMENDED)
```
IanNod: why is this the recommended option?

scripts/gemm_analysis/enhance_gemm_variance_with_timestamps.py
```diff
@@ -0,0 +1,270 @@
...
+
+    result = {
+        'min_timestamp_ms': min_instance['timestamp_ms'],
+        'max_timestamp_ms': max_instance['timestamp_ms'],
+        'min_duration_found_us': min_instance['duration_us'],
+        'max_duration_found_us': max_instance['duration_us']
+    }
+
+    # Print warnings if mismatch
+    if abs(min_instance['duration_us'] - min_duration_us) > min_tolerance:
```
IanNod, line 91: I think actually calculating a variance instead of comparing min/max durations would be more robust. There may be some kernels that are expected to have higher variability than a 1% tolerance in min/max times.

scripts/gemm_analysis/enhance_gemm_variance_with_timestamps.py
```diff
@@ -0,0 +1,270 @@
...
+    """Parse command line arguments."""
+    parser = argparse.ArgumentParser(
+        description="Enhance GEMM variance CSV with timestamp information",
+        formatter_class=argparse.RawDescriptionHelpFormatter
+    )
+
+    parser.add_argument(
+        '--input-csv',
+        type=Path,
+        default=Path("/home/oyazdanb/aorta/experiments/sweep_20251121_155219/tracelens_analysis/top5_gemm_kernels_time_variance.csv"),
```
IanNod: Should avoid specific user/file paths for default arguments

scripts/gemm_analysis/enhance_gemm_variance_with_timestamps.py
```diff
@@ -0,0 +1,270 @@
...
+        '--output-csv',
+        type=Path,
+        default=None,
+        help='Output CSV file (default: input_file_with_timestamps.csv)'
+    )
+
+    parser.add_argument(
+        '--base-path',
+        type=Path,
+        default=Path("/home/oyazdanb/aorta/experiments/sweep_20251121_155219"),
```
IanNod: same as above

scripts/gemm_analysis/plot_gemm_variance.py
```diff
@@ -0,0 +1,389 @@
...
+                mean_val = sum(thread_channel_data[thread][channel]) / len(thread_channel_data[thread][channel])
+                means.append(mean_val)
+            else:
+                means.append(0)
+
+        marker = 'o' if thread == 256 else 's'
+        label = f'{thread} threads'
+        ax.plot(channels, means, marker=marker, linewidth=2, markersize=10, label=label)
+
+    ax.set_xlabel('Channel Count', fontsize=14, fontweight='bold')
```
IanNod: There seems to be a lot of overalap in the axes across these different functions/plots. Could we create a utility function to set these up instead essentially repeating much of this code?

scripts/gemm_analysis/process_gpu_timeline.py
```diff
@@ -0,0 +1,297 @@
...
+
+def geometric_mean(values):
+    """Calculate geometric mean, handling zeros."""
+    values = np.array(values)
+    # Replace zeros with small value to avoid log(0)
+    values = np.where(values == 0, 1e-10, values)
+    return np.exp(np.mean(np.log(values)))
+
+
+def process_gpu_timeline_data(sweep_dir, use_geo_mean=False):
```
IanNod, line 344: This function is doing a lot. Can you split it up into components?

scripts/gemm_analysis/enhance_gemm_variance_with_timestamps.py
oyazdanb: I removed the other option because it filtered to only the top 100 slowest kernel instances, which excludes the fastest instances. This mean we never find the min duration kernel.

scripts/gemm_analysis/enhance_gemm_variance_with_timestamps.py
oyazdanb: Good catch - I think there is a mixup here. The tolerance we have here is for validating the min/max we find in traces and not handing kernel exectuion variance. Kernel variance can be consider next feature.

IanNod: Yes, I understand this tolerance was set for the current config/workload being tested. I just think that a true variance calculation would be more robust to more workloads. Makes sense to add as a later feature, maybe add a todo comment for it

scripts/gemm_analysis/plot_gemm_variance.py
prosenjitdhole: All the plots are now refactored to reuse the same code

scripts/gemm_analysis/process_gpu_timeline.py
prosenjitdhole: Refactored.

scripts/gemm_analysis/create_embeded_html_report.py
```diff
@@ -0,0 +1,181 @@
...
+
+    parser.add_argument(
+        '--sweep1',
+        type=Path,
+        required=True,
+        help='Path to first sweep directory'
+    )
+
+    parser.add_argument(
+        '--sweep2',
```
IanNod, line 57: Is there a reason we are providing 2 sweep args instead of one comma separated one? Doing that would allow comparing a larger sample of sweeps.

scripts/gemm_analysis/enhance_gemm_variance_with_timestamps.py
```diff
@@ -0,0 +1,268 @@
...
+        print(f"  Warning: No instances of kernel {kernel_name[:50]}... found")
+        return {
+            'min_timestamp_ms': None,
+            'max_timestamp_ms': None,
+            'min_duration_found_us': None,
+            'max_duration_found_us': None
+        }
+
+    # Sort by duration
+    kernel_instances.sort(key=lambda x: x['duration_us'])
```
IanNod, line 72: Is duration_us always guaranteed to exist? If not this will fail, and we should add a check if it exists

scripts/gemm_analysis/gemm_report_with_collective_overlap.py
```diff
@@ -0,0 +1,343 @@
...
+        '--output-csv',
+        type=Path,
+        default=None,
+        help='Output CSV file (default: input_file_with_collective_overlap.csv)'
+    )
+
+    parser.add_argument(
+        '--tracelens-path',
+        type=Path,
+        default=Path("/home/oyazdanb/aorta/experiments/sweep_20251121_155219/tracelens_analysis"),
```
IanNod: remove user dir for default path

scripts/gemm_analysis/create_embeded_html_report.py
prosenjitdhole: Right now this script only support 2 sweep. But generalising to n sweep could be a good feature to add. Right now the code is structured for 2 sweeps only.

scripts/gemm_analysis/enhance_gemm_variance_with_timestamps.py
oyazdanb: fixed.

scripts/gemm_analysis/create_embeded_html_report.py
oyazdanb: we have added a Note in [Readme](https://github.com/ROCm/aorta/pull/12/files#diff-e1a6be8580ff6e78b4839e9852b0143e618488a0ed31bae25b4eb6dbffa19127R233) as well.

IanNod (approved):

PR #14, opened 2025-12-03
Generate merged report and plots for rccl runs

prosenjitdhole:
This generates the final html report. 
Need to fill in : 
1. Percentage change plots
2. Modify the README


scripts/tracelens_single_config/generate_merged_html.py
```diff
@@ -0,0 +1,182 @@
...
+def get_image_data(image_path):
+    try:
+        with open(image_path, "rb") as f:
+            return base64.b64encode(f.read()).decode("utf-8")
+    except Exception as e:
+        print(f"Error getting image data: {e}")
+        return None
+
+
+def create_final_html(plot_file_path, output_path):
```
oyazdanb, line 16: Could we extend/modify one of the existing HTML report generators instead of creating a new one? We already have:
 - [scripts/gemm_analysis/create_embeded_html_report.py](https://github.com/ROCm/aorta/blob/main/scripts/gemm_analysis/create_embeded_html_report.py)
 - [scripts/gemm_analysis/html_template.py](https://github.com/ROCm/aorta/blob/main/scripts/gemm_analysis/html_template.py)
>
 This would help reduce code duplication and make maintenance easier. If there are specific features needed that aren't in those scripts, perhaps we could enhance one of them instead?

PR #24, opened 2025-12-08
[WIP][scripts] Plumb rocprofv3 profile handling

vinayakdsci:
Initial draft for the integration of `rocprofv3` into the data summarization pipeline.

Currently we generate `pftrace` files from `rocprofv3` and then convert them to a reliably merged JSON files through the `traceconv` tool provided by `perfetto`, as other ways of merging JSON output files are not reliable.

[WIP]: Still requires wiring in `TraceLens` and final report generation.

### DIFF
--- a/scripts/convert_pftrace_to_json.sh
+++ b/scripts/convert_pftrace_to_json.sh
@@ -7,7 +7,7 @@ log() {
     local timestamp=$(date '+%Y-%m-%d %H:%M:%S')
     echo "[${timestamp}] ${message}" 2>&1
 }
-            
+
 usage() {
     echo "Usage: $0 [OPTIONS]"
     echo "Options:"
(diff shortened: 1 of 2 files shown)


oyazdanb:
we could define the output format as json in the yaml input passed to rocprof. why do we need to convert to perfetto first? https://github.com/ROCm/aorta/blob/main/scripts/gemm_analysis/README.md#simple-yaml-based-tracing-recommended

vinayakdsci:
> we could define the output format as json in the yaml input passed to rocprof. why do we need to convert to perfetto first? https://github.com/ROCm/aorta/blob/main/scripts/gemm_analysis/README.md#simple-yaml-based-tracing-recommended

As per the documentation at [Rocprof Section](https://github.com/ROCm/aorta?tab=readme-ov-file#rocm-rocprofv3-capture), I and @amd-vivekag were under the impression that we need to generate a single merged JSON file from multiple rocprof traces. We tried merging json files, but it was not working as expected, so we used pftrace approach to merge these traces. Let us know if you see any issue in this approach. 
Also, we wanted to know if we really need to merge the JSON files, because tracelens should be able to convert individual json files as well into html. So, we can avoid merging step if it is not required.

PR #23, opened 2025-12-08
Initial check in for single config tarcelense support

prosenjitdhole:
This PR is related to #15. 
This checkin is for support of 
- Analyzing single config tracelens runs
- Compare two single config tracelens run

This checkin includes : 
- run_tracelens_single_config.sh - To generate tracelense reports on a single config run 
- process_gpu_timeline.py - To creat a single gpu mean/geomean timeline excel for a single config run. _(This extra script is required as gpu time line data is dumped for each individual run. We will not require such a script for nccl data as they are collective in nature)_
- combine_reports.py - To combine collective (nccl channel data over all ranks) and gpu_timeline summary of different runs
- add_comparison_sheet.py - To add gputime comparison on the combined gpu timeline excles
- add_collective_comparison_sheets.py - To add nccl comparison data on combined nccl runs 
- create_final_report.py - To generate a combined report of a) combined and b) comparison data for both gpu timeline and nccl collective data
- create_final_plots.py - To create plots from the final report excel
- create_final_html.py - To create the html report from the plots 
- html_report_config.py - Contains the template html and configuration of what plots to print
- run_full_analysis.py - Wrapper script To run all the above scripts in due order.

Note : 
- These are the base scripts on which rccl_test will depend (future check in) 
- Few script here might be similar to gemm_analysis scripts, but these are the base scripts and later we will modify gemm scripts to use these scripts. 
- #15 will be closed as there are hard conflicts in rebasing to gemm script updates.

PR #26, opened 2025-12-08
Added process gpu timeline and combining different reports

prosenjitdhole:
Orignal PR : #23 : Split across mutiple prs
subpr : 2
Files :
- scripts/tracelens_single_config/combine_reports.py
- scripts/tracelens_single_config/process_gpu_timeline.py

### DIFF
--- a/scripts/tracelens_single_config/combine_reports.py
+++ b/scripts/tracelens_single_config/combine_reports.py
@@ -28,7 +28,7 @@ def combine_collective_reports(baseline_path, test_path, output_path):
             test_df = pd.read_excel(test_path, sheet_name=sheet_name)
 
             baseline_df["source"] = "baseline"
-            test_df["source"] = "saleelk"
+            test_df["source"] = "test"
 
             combined = pd.concat([baseline_df, test_df], ignore_index=True)
 
--- a/scripts/tracelens_single_config/combine_reports.py
+++ b/scripts/tracelens_single_config/combine_reports.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+import pandas as pd
+import argparse
+from pathlib import Path
+
+
+def combine_collective_reports(baseline_path, test_path, output_path):
+    """
+    Combine two collective reports into a single Excel file by adding a source column to the data.
+    """
+
+    print(f"Loading baseline: {baseline_path}")
+    baseline_xl = pd.ExcelFile(baseline_path)
+
+    print(f"Loading test: {test_path}")
+    test_xl = pd.ExcelFile(test_path)
+
+    print(f"\nBaseline sheets: {baseline_xl.sheet_names}")
+    print(f"Test sheets: {test_xl.sheet_names}")
+
+    with pd.ExcelWriter(output_path, engine="openpyxl") as writer:
+        for sheet_name in baseline_xl.sheet_names:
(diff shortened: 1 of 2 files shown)


scripts/tracelens_single_config/combine_reports.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
```
IanNod, line 1: How is combine_reports related to process_gpu_timeline? Should they be separate PR's?

scripts/tracelens_single_config/process_gpu_timeline.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
+import pandas as pd
+import numpy as np
+import argparse
+from pathlib import Path
+
+
+def geometric_mean(values):
```
IanNod, line 8: This function already exists [here](https://github.com/ROCm/aorta/blob/795dfd4ee7066226ecb470c712bfc73fd023460a/scripts/gemm_analysis/process_gpu_timeline.py#L28)

scripts/tracelens_single_config/process_gpu_timeline.py
```diff
@@ -0,0 +1,100 @@
...
+    values = np.array(values)
+    values = np.where(values == 0, 1e-10, values)
+    return np.exp(np.mean(np.log(values)))
+
+
+def process_gpu_timeline(reports_dir, use_geo_mean=False):
```
IanNod, line 14: This looks very similar in functionality to https://github.com/ROCm/aorta/blob/795dfd4ee7066226ecb470c712bfc73fd023460a/scripts/gemm_analysis/process_gpu_timeline.py#L344. Is there a reason we don't just reuse or modify that?

prosenjitdhole: Hi Ian, these two functions are different, other than reading an excel and grouping by type, nothing is almost common. Referred one read a different directory structure and produce an excel with thread and channel information, whereas the one in the PR is a lot simplified version, it reads a directory structure without thread and channel information and hence produce a lot simpler excel. We can refactor them to use same base, but the common part is just reading excel and group by type along with the mean calculation (mean/geometric mean). Rest all will be different.

prosenjitdhole: A sample implementation of what I explained : https://github.com/ROCm/aorta/pull/33

IanNod: Is there a reason we need to have a different directory structure for gemm_analysis? Seem to me that we could better abstract and combine the two single config and gemm analysis to use the same overarching function and pass in the types that you want to group along as an arg and the gemm analysis can loop through the larger set of configs with a maybe slightly modified file structure that it sweeps through. 

For sake of time this could be done later if you add a TODO comment of how we could better merge these two

prosenjitdhole: Essentially GEMM is done for sweep run which is 8 single config (2 thread x 4 channels as of now).
Tracelens dumps comes in different format .. But we can create an abstraction to read the excel file in single config and iteration in sweep analysis (GEMM). Something what we already did in #33 but with some more abstraction.

IanNod (approved):

PR #29, opened 2025-12-08
Adding top level scripts to run the full flow and generate reports

prosenjitdhole:
Orignal PR : https://github.com/ROCm/aorta/pull/23 : Split across mutiple prs
subpr : 5
Files :
- scripts/tracelens_single_config/create_final_report.py
- scripts/tracelens_single_config/run_full_analysis.py

### DIFF
--- a/scripts/tracelens_single_config/run_full_analysis.py
+++ b/scripts/tracelens_single_config/run_full_analysis.py
@@ -1,0 +1,387 @@
+#!/usr/bin/env python3
+"""
+Master script for complete TraceLens analysis pipeline.
+Runs analysis on baseline and test traces, then performs all comparisons.
+"""
+import argparse
+import subprocess
+import os
+import sys
+from pathlib import Path
+
+
+def run_command(cmd, description):
+    """Execute a command and handle errors."""
+    print(f"\n{'='*80}")
+    print(f"{description}")
+    print(f"{'='*80}")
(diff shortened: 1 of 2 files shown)


scripts/tracelens_single_config/run_full_analysis.py
```diff
@@ -0,0 +1,387 @@
+#!/usr/bin/env python3
```
IanNod, line 1: This script looks like it is essentially just a series of run commands. Would it be better as a bash script?

prosenjitdhole: This script is kept as python as going forward we can : 
1. Keep a single top level full_analysis script for all the flows
2. Add more functionality (which could be difficult in bash)
3. Add more design complexity.

IanNod: what further functionality (that would be difficult in bash) or design complexity are we looking to add?

prosenjitdhole: We plan to have a container injection on this run_full_analysis, where the steps are predefined (using a contractor) but the filler will come from an implementation specific to a flow. 

Also, in a general guideline, we are converting our existing scripts to python as well for aorta (like @oyazdanb is converting all our bash files to python).

IanNod (approved):

PR #35, opened 2025-12-09
updated screenshots in docs

mycpuorg:


mycpuorg:
looks like the pre-commit check is failing on the updated report and doc. Will override and merge since no code was changed.

PR #36, opened 2025-12-11
[scripts] Resolve FileNotFoundError for trace files during TraceLens collective reporting

vinayakdsci:
`scripts/gemm_analysis/run_tracelens_analysis.sh` assumes during collective reporting that each rank folder will have the same basename for the Pytorch trace profile.
This is not guaranteed, and when we run `TraceLens` with the `--trace_pattern` flag (which allows only one `*` wildcard in the whole path), the code raises a `FileNotFoundError` because of differing tracefile names between different ranks.

This path moves the tracefile to a canonical location in the rank folder under `trace/pt.trace.json` so that we can use the `--trace_pattern` flag to consolidate multiple ranks.


oyazdanb (approved):

amd-vivekag (approved):
LGTM

PR #54, opened 2025-12-17
Add support to compare multiple test run with a single base run

prosenjitdhole:
This is a change from rccl 3 run visulaization. 

Now, we can invoke run_full_analysis with mutiple test run with a single baseline run to generate the comparison html directly.

### DIFF
--- a/scripts/tracelens_single_config/compare_all_runs.py
+++ b/scripts/tracelens_single_config/compare_all_runs.py
@@ -1,0 +1,454 @@
+import argparse
+from pathlib import Path
+import subprocess
+import pandas as pd
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+def run_command(cmd, description):
+    """Execute a command and handle errors."""
+    print(f"\n{'='*80}")
+    print(f"{description}")
+    print(f"{'='*80}")
+    print(f"Command: {' '.join(cmd)}")
+
+    result = subprocess.run(cmd, capture_output=True, text=True)
+
+    if result.returncode != 0:
+        print(f"Error: {description} failed!")
+        print(f"Stderr: {result.stderr}")
+        return False
+
+    print(result.stdout)
+    return True
+
+
+def plot_nccl_data_per_msg(df, labels, output_dir: Path):
+    """
+    Plot comm_latency_mean for each message size from NCCL data.
+    """
+    output_path = Path(output_dir)
+    output_path.mkdir(parents=True, exist_ok=True)
+
+    # Get unique index values (Collective_MsgSize)
+    indices = df["index"].values
+
+    x = np.arange(len(indices))
+    width = 0.8 / len(labels)
+    # Vibrant color palette
+    vibrant_colors = [
+        "#E63946",
+        "#2A9D8F",
+        "#E9C46A",
+        "#264653",
+        "#F4A261",
+        "#8338EC",
+        "#06D6A0",
+        "#FF006E",
+    ]
+
+    plot_items = {
+        "NCCL Communication Latency": {
+            "x_label": "Collective Operation (Message Size)",
+            "y_label": "Communication Latency (ms)",
+            "y_col": "comm_latency_mean",
+        },
+        "NCCL Algorithm Bandwidth": {
+            "x_label": "Collective Operation (Message Size)",
+            "y_label": "Algorithm Bandwidth (GB/s)",
+            "y_col": "algo bw (GB/s)_mean",
+        },
+        "NCCL Bus Bandwidth": {
+            "x_label": "Collective Operation (Message Size)",
+            "y_label": "Bus Bandwidth (GB/s)",
+            "y_col": "bus bw (GB/s)_mean",
+        },
+        "NCCL Total Communication Latency": {
+            "x_label": "Collective Operation (Message Size)",
+            "y_label": "Total Communication Latency (ms)",
+            "y_col": "Total comm latency (ms)",
+        },
+    }
+
+    for plot_item in plot_items.keys():
+        fig, ax = plt.subplots(figsize=(14, 7))
+        for i, label in enumerate(labels):
+            col_name = f"{plot_items[plot_item]['y_col']}_{label}"
+            print(f"Plotting {col_name}")
+            if col_name in df.columns:
+                values = df[col_name].values
+                color = vibrant_colors[i % len(vibrant_colors)]
+                offset = (i - len(labels) / 2 + 0.5) * width
+                ax.bar(
+                    x + offset,
+                    values,
+                    width,
+                    label=label,
+                    color=color,
+                    alpha=0.85,
+                    edgecolor="black",
+                    linewidth=0.5,
+                )
+            else:
+                print(f"Column {col_name} not found in dataframe")
+
+        ax.set_xlabel(plot_items[plot_item]["x_label"], fontsize=12, fontweight="bold")
+        ax.set_ylabel(plot_items[plot_item]["y_label"], fontsize=12, fontweight="bold")
+        ax.set_title(f"{plot_item} per Message Size", fontsize=14, fontweight="bold")
+        ax.set_xticks(x)
+        ax.set_xticklabels(indices, rotation=45, ha="right", fontsize=9)
+        ax.legend(loc="upper left")
+        ax.grid(True, alpha=0.3, axis="y")
+
+        plt.tight_layout()
+        output_file = output_path / f'{plot_item.replace(" ", "_")}_comparison.png'
+        plt.savefig(output_file, dpi=150, bbox_inches="tight")
+        plt.close()
+        print(f"Saved: {output_file}")
+    print("Completed plotting NCCL data per message size")
+
+
+def plot_all_types_per_rank(df, labels, output_dir: Path):
+    """
+    Plot data for every rank, where every unique type is a different file.
+
+    Parameters:
+    -----------
+    df : DataFrame
+        Merged gpu_time_per_rank_df with columns like 'type', 'rank0_label1', 'rank0_label2', etc.
+    labels : list
+        List of configuration labels (e.g., ['32cu_512threads', '37cu_384threads'])
+    output_dir : str
+        Directory to save plots
+    """
+    output_path = Path(output_dir)
+    output_path.mkdir(parents=True, exist_ok=True)
+
+    unique_types = df["type"].unique()
+
+    # Find rank columns (extract rank numbers from column names)
+    # Columns are like: rank0_32cu_512threads, rank1_32cu_512threads, etc.
+    sample_label = labels[0]
+    rank_cols = [col for col in df.columns if col.endswith(f"_{sample_label}") and col != "type"]
+    ranks = [col.replace(f"_{sample_label}", "") for col in rank_cols]
+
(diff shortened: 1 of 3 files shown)


scripts/tracelens_single_config/run_full_analysis.py
```diff
@@ -54,10 +52,26 @@ def process_gpu_timeline(reports_dir):
...
+    script_path = Path(__file__).parent / "compare_all_runs.py"
+    cmd = [
+        "python3",
+        str(script_path),
+        "--inputs",
+        [str(p) for p in all_paths],
+        "--output",
+        str(output_path),
+    ]
+    return run_command(cmd, "Comparing all runs")
```
amd-vivekag, line 69: any particular reason that instead of calling the python API, complete python process is being invoked here. Shouldn't be better to call the relevant API from here?

scripts/tracelens_single_config/run_full_analysis.py
```diff
@@ -82,9 +96,7 @@ def combine_reports(baseline_file, test_file, output_file, baseline_label=None,
 def add_comparison_sheets(input_file, output_file, baseline_label=None, test_label=None):
     """Add comparison sheets for GPU timeline."""
     script_path = Path(__file__).parent / "add_comparison_sheets.py"
-    cmd = ["python3", str(script_path),
-           "--input", input_file,
-           "--output", output_file]
+    cmd = ["python3", str(script_path), "--input", input_file, "--output", output_file]
```
amd-vivekag, line 102: Although this was existing code, but I think all these python code should be called through API to keep the same process instead of a separate process.

scripts/tracelens_single_config/run_full_analysis.py
```diff
@@ -95,9 +107,7 @@ def add_comparison_sheets(input_file, output_file, baseline_label=None, test_lab
 def add_collective_comparison(input_file, output_file, baseline_label=None, test_label=None):
     """Add comparison sheets for collective operations."""
     script_path = Path(__file__).parent / "add_collective_comparison.py"
-    cmd = ["python3", str(script_path),
-           "--input", input_file,
-           "--output", output_file]
+    cmd = ["python3", str(script_path), "--input", input_file, "--output", output_file]
```
amd-vivekag, line 113: Same comment: Although this was existing code, but I think all these python code should be called through API to keep the same process instead of a separate process.

scripts/tracelens_single_config/run_full_analysis.py
```diff
@@ -106,17 +116,29 @@ def add_collective_comparison(input_file, output_file, baseline_label=None, test
...
     """Create comprehensive final report with all data."""
     script_path = Path(__file__).parent / "create_final_report.py"
     cmd = [
-        "python3", str(script_path),
-        "--gpu-combined", gpu_combined,
-        "--gpu-comparison", gpu_comparison,
-        "--coll-combined", coll_combined,
-        "--coll-comparison", coll_comparison,
-        "--output", output_file,
+        "python3",
```
amd-vivekag, line 133: Same comment: Although this was existing code, but I think all these python code should be called through API to keep the same process instead of a separate process.

scripts/tracelens_single_config/run_full_analysis.py
```diff
@@ -195,36 +214,55 @@ def main():
...
+    parser.add_argument(
+        "--compare-all-runs",
+        action="store_true",
+        help="Merge all TraceLens analysis into a single report",
+    )
+
     args = parser.parse_args()
 
-    # Handle --all flag
     if args.all:
```
amd-vivekag, line 240: what is the difference between `all` and `compare_all_runs`? What if both options were passed by the user? Since under these options, some of the arguments are being overridden, so it might result into unexpected run results.

scripts/tracelens_single_config/run_full_analysis.py
```diff
@@ -195,36 +214,55 @@ def main():
...
         args.gpu_timeline = True
         args.collective = True
         args.final_report = True
         args.generate_plots = True
 
+    if args.compare_all_runs:
+        args.gpu_timeline = True
```
amd-vivekag, line 246: Is it possible to tie up these arguments with argparse compare_all_runs argument? That way you can make it mutually exclusive.

scripts/tracelens_single_config/run_full_analysis.py
```diff
@@ -246,75 +284,101 @@ def main():
...
         return 1
 
-    if not test_analysis.exists():
-        print(f"Error: Test analysis not found: {test_analysis}")
-        print("Run without --skip-tracelens flag first")
-        return 1
+    for test_path in test_paths:
+        test_analysis = test_path / "tracelens_analysis"
+        if not test_analysis.exists():
+            print(f"Error: Test analysis not found: {test_analysis}")
```
amd-vivekag, line 313: Instead of print statement, use python logger so that the messages gets logged in a file. This applies to all print statement

scripts/tracelens_single_config/run_full_analysis.py
```diff
@@ -246,75 +284,101 @@ def main():
...
+            print(f"Processing {test_path.name} GPU timeline...")
+            if not process_gpu_timeline(str(individual_report)):
+                return 1
+
+        if args.compare_all_runs:
+            all_paths = [baseline_path] + test_paths
+            if not run_compare_all_runs(all_paths, output_path):
+                return 1
+            return 0
+        # if not baseline_reports.exists() or not test_reports.exists():
```
amd-vivekag, line 359: please remove commented code

scripts/tracelens_single_config/run_full_analysis.py
```diff
@@ -246,75 +284,101 @@ def main():
...
+            test_analysis / "individual_reports"
+        )  # should correctly point to the last path if multiple test os not provided.
+
+        temp_test_paths = test_paths.copy()
+        temp_test_paths.append(baseline_path)
+        for test_path in temp_test_paths:
+            individual_report = test_path / "tracelens_analysis" / "individual_reports"
+            collective_report = test_path / "tracelens_analysis" / "collective_reports"
+            if not individual_report.exists() or not collective_report.exists():
+                print(f"Error: individual/collective reports not found: {test_reports}")
```
amd-vivekag: It seems, message is incomplete. It is printing only individual_reports directory and not collective_reports.

scripts/tracelens_single_config/compare_all_runs.py
```diff
@@ -0,0 +1,454 @@
...
+
+
+def run_command(cmd, description):
+    """Execute a command and handle errors."""
+    print(f"\n{'='*80}")
+    print(f"{description}")
+    print(f"{'='*80}")
+    print(f"Command: {' '.join(cmd)}")
+
+    result = subprocess.run(cmd, capture_output=True, text=True)
```
amd-vivekag, line 16: Any particular reason of running it as subprocess? python to python API should be more efficient.

scripts/tracelens_single_config/compare_all_runs.py
```diff
@@ -0,0 +1,454 @@
...
+        ax.set_xticklabels(indices, rotation=45, ha="right", fontsize=9)
+        ax.legend(loc="upper left")
+        ax.grid(True, alpha=0.3, axis="y")
+
+        plt.tight_layout()
+        output_file = output_path / f'{plot_item.replace(" ", "_")}_comparison.png'
+        plt.savefig(output_file, dpi=150, bbox_inches="tight")
+        plt.close()
+        print(f"Saved: {output_file}")
+    print("Completed plotting NCCL data per message size")
```
amd-vivekag, line 109: print calls should be replaced by python logger

scripts/tracelens_single_config/compare_all_runs.py
```diff
@@ -0,0 +1,454 @@
...
+    ax.legend()
+    ax.grid(True, alpha=0.3, axis="y")
+
+    plt.tight_layout()
+    plt.savefig(output_dir / "abs_time_comparison.png")
+    plt.close()
+
+
+"""
+def plot_improvement_chart(df, output_path):
```
amd-vivekag: why is it commented out? If not required, then the code can be removed.

scripts/tracelens_single_config/compare_all_runs.py
```diff
@@ -0,0 +1,454 @@
...
+    args = parser.parse_args()
+
+    args.output.mkdir(parents=True, exist_ok=True)
+
+    labels = []
+    summary_dfs = []
+    gpu_time_per_rank_dfs = []
+    nccl_dfs = []
+
+    for directory in args.inputs:
```
amd-vivekag: Can you please breakdown main function into smaller functions? It will help us to understand the overall algo of this code.

amd-vivekag (changes requested):

scripts/tracelens_single_config/run_full_analysis.py
prosenjitdhole: --all enable all the internal phases, and compare a report between baseline and test, 
where as --compare-all-runs take a single baseline and mutiple test runs and prepare a comparison report between them .. 

if --comapare-all-runs are not given, then only one test directory is allowed, with this option we can give multiple test directory .. 

These two are unrelated options. Naming could be better, if you can suggest.

scripts/tracelens_single_config/run_full_analysis.py
prosenjitdhole: This is the pattern we are following in this script. We can wrap the main function into another function and call the same, but we are following this process from the begining.

scripts/tracelens_single_config/run_full_analysis.py
prosenjitdhole: Same as above

scripts/tracelens_single_config/run_full_analysis.py
prosenjitdhole: Same as above

scripts/tracelens_single_config/run_full_analysis.py
prosenjitdhole: Sure, We can keep it as a TODO.

scripts/tracelens_single_config/run_full_analysis.py
prosenjitdhole: Thanks, it should print test_path

scripts/tracelens_single_config/run_full_analysis.py
amd-vivekag: I think we should convert them into API calls. That will be much faster and will be sharing same memory because of a single process. There won't be any overhead as well. You can raise separate PR for this if you don't want to mix those changes in this PR.

prosenjitdhole: Sure, I have added a TODO, and will replace this ina  seperate PR later.

scripts/tracelens_single_config/run_full_analysis.py
amd-vivekag: Those are not unrelated from run point of view, because if both options will be on, the value set to the args variables will be overriden by if block of compare-all-runs. This will impact the behavior for run with "all" option.

amd-vivekag (approved):

amd-vivekag (approved):
LGTM

PR #57, opened 2026-01-15
AORTA-5 - Multi-Node worklflow

oyazdanb:
Adding multi-node workload for Aorta. The instruction on how to run is mentioned in README. Related to AORTA-5


scripts/multi_node/set_env_variables.sh.bak
amd-vivekag, line 1: Why do we need a bak file in the git repo?

scripts/multi_node/start_docker_all_nodes.sh
amd-vivekag, line 1: Ideally, we should be running slurm commands from the login machine instead of compute machine. Some of the checks would have been removed in that case. 
For now, we can keep this but please add TODO to convert these commands to SLURM commands to avoid ssh to individual machines to run the code.

amd-vivekag (approved):
LGTM

PR #76, opened 2026-01-28
AORTA-17 : Process command implementation

prosenjitdhole:
Process Implementation.
Dev Spec : https://github.com/ROCm/aorta/blob/399ade34606fbf318abe330b3055ff062b51b37e/src/aorta/report/PROCESS_CMD_DEV_DOCS.md

### DIFF
--- a/src/aorta/report/processing/gpu_timeline_single.py
+++ b/src/aorta/report/processing/gpu_timeline_single.py
@@ -1,0 +1,143 @@
+"""
+GPU Timeline Processing - Single Configuration Mode.
+
+Processes GPU timeline data from TraceLens individual reports for a single
+configuration (no thread/channel variations). Aggregates across ranks.
+
+Source: scripts/tracelens_single_config/process_gpu_timeline.py
+"""
+
+from pathlib import Path
+from typing import Optional
+
+import numpy as np
+import pandas as pd
+
+
+def geometric_mean(values: np.ndarray) -> float:
+    """
+    Calculate geometric mean, handling zeros.
+
+    Args:
+        values: Array of values
+
+    Returns:
+        Geometric mean value
+    """
+    values = np.array(values)
+    values = np.where(values == 0, 1e-10, values)
+    return float(np.exp(np.mean(np.log(values))))
+
+
+def process_single_config(
+    reports_dir: Path,
+    use_geo_mean: bool = False,
+    output_path: Optional[Path] = None,
+    verbose: bool = False,
+) -> Optional[Path]:
+    """
+    Process GPU timeline from single config individual reports.
+
+    Reads gpu_timeline sheet from each perf_rank*.xlsx file and aggregates
+    across all ranks using mean or geometric mean.
+
+    Args:
+        reports_dir: Path to individual_reports directory containing perf_rank*.xlsx
+        use_geo_mean: Use geometric mean instead of arithmetic mean
+        output_path: Custom output path (default: parent/gpu_timeline_summary_{method}.xlsx)
+        verbose: Print verbose output
+
+    Returns:
+        Path to output Excel file or None if processing failed
+    """
+    reports_path = Path(reports_dir)
+
+    if not reports_path.exists():
+        raise FileNotFoundError(f"Directory not found: {reports_dir}")
+
+    agg_method = "Geometric Mean" if use_geo_mean else "Arithmetic Mean"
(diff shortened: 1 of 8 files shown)


src/aorta/report/processing/gpu_timeline_single.py
```diff
@@ -0,0 +1,143 @@
...
+"""
+
+from pathlib import Path
+from typing import Optional
+
+import numpy as np
+import pandas as pd
+
+
+def geometric_mean(values: np.ndarray) -> float:
```
vmnit: same function in 2 different files. Please create a utils from where you can reuse the same function

src/aorta/report/processing/gpu_timeline_sweep.py
```diff
@@ -0,0 +1,435 @@
...
+    """
+    Parse performance filename to extract channel config and rank.
+
+    Args:
+        filename: e.g., 'perf_28ch_rank0.xlsx'
+
+    Returns:
+        Tuple of (channel_config, rank) e.g., ('28ch', 0)
+    """
+    parts = filename.replace("perf_", "").replace(".xlsx", "").split("_")
```
vmnit, line 30: please add assert if the filename contains these strings that you want to replace. If those are there, this function will run silently without letting user know about it.

src/aorta/report/processing/gpu_timeline_sweep.py
```diff
@@ -0,0 +1,435 @@
...
+    channel_config = parts[0]  # e.g., "28ch"
```
vmnit, line 31: please also put assert on len(parts)

src/aorta/report/processing/gpu_timeline_sweep.py
```diff
@@ -0,0 +1,435 @@
...
+        List of DataFrames with rank column added
+    """
+    rank_data = []
+    for rank, file_path in rank_files:
+        try:
+            df = pd.read_excel(file_path, sheet_name="gpu_timeline")
+            df["rank"] = rank
+            rank_data.append(df)
+        except Exception as e:
+            if verbose:
```
vmnit: while reading excel, if there is any exception, then it should be reported.

src/aorta/report/processing/gpu_timeline_sweep.py
```diff
@@ -0,0 +1,435 @@
...
+                print(f"    Warning: Could not read {Path(file_path).name}: {e}")
```
vmnit: all the print statements should be converted to logger, so that appropriate tagging would be there.

src/aorta/report/processing/process_comms.py
```diff
@@ -0,0 +1,291 @@
...
+            channel_config = filename.replace("collective_", "").replace(".xlsx", "")
+            channels_num = int(channel_config.replace("ch", ""))
+            threads_num = int(thread_config.replace("thread", ""))
+
+            if verbose:
+                print(f"  Reading: {filename}")
+
+            try:
+                # Read the nccl_summary_implicit_sync sheet
+                df = pd.read_excel(file_path, sheet_name="nccl_summary_implicit_sync")
```
vmnit, line 108: please use config or something instead of hard coding of the filenames in the code. If there is any change in the name, it would be hard to replace everywhere.

vmnit (changes requested):

src/aorta/report/processing/gpu_timeline_sweep.py
prosenjitdhole: That check is handled upper in the flow, if it does not find files of the same pattern flow will not work ..

src/aorta/report/processing/gpu_timeline_sweep.py
prosenjitdhole: Will add a TODO for that to be handled as a part of AORTA-9

src/aorta/report/processing/process_comms.py
prosenjitdhole: We will start using a config file in AORTA-9

src/aorta/report/processing/gpu_timeline_sweep.py
prosenjitdhole: Done, converted to error.

amd-vivekag (approved):
LGTM

amd-vivekag (approved):
LGTM

PR #113, opened 2026-02-23
AORTA-21: Add comms-compute overlap workload with distributed support

vivekkhandelwal1:
Introduce a configurable comm-compute overlap workload to hw_queue_eval that runs GEMM compute on dedicated streams while collective communication runs on a separate stream. This is the simplest workload for measuring overlap efficiency and GPU queue scheduling quality.

Key features:
- Three modes: compute_only, comms_only, comms_compute
- Real NCCL/RCCL collectives via torchrun (--real-collectives)
- Async non-blocking collectives with deferred wait (--async-op)
- Arbitrary process group topologies (--process-groups)
- Independent compute stream count (--compute-streams)
- Separate data types for compute and comm (--comp-dtype, --comm-dtype)
- Configurable GEMM dimensions, iteration counts, and message sizes
- Rank-aware profiler trace filenames in distributed mode

Signed-off-by: Vivek Khandelwal <vivek.khandelwal@amd.com>

### DIFF
--- a/src/aorta/hw_queue_eval/workloads/distributed/comms_compute_overlap.py
+++ b/src/aorta/hw_queue_eval/workloads/distributed/comms_compute_overlap.py
@@ -289,6 +289,16 @@ class CommsComputeOverlapWorkload(MultiGPUMixin, DistributedWorkload):
         if self.mode != "compute_only":
             elem_size = torch.tensor([], dtype=self.comm_data_type).element_size()
             num_elements = max(1, self.comm_size_bytes // elem_size)
+            # Two buffers (comm + scratch) are allocated
+            alloc_bytes = num_elements * elem_size * 2
+            if alloc_bytes > 4 * 1024 ** 3:
+                logger.warning(
+                    "Comm tensor allocation is %.1f GiB (comm_size_bytes=%d, "
+                    "dtype=%s). This may cause out-of-memory errors.",
+                    alloc_bytes / 1024 ** 3,
+                    self.comm_size_bytes,
+                    self.comm_data_type,
+                )
             comm_device = self._get_device_for_stream(self._comm_stream_idx)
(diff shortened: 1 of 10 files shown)


src/aorta/hw_queue_eval/workloads/distributed/comms_compute_overlap.py
```diff
@@ -0,0 +1,453 @@
...
+        - **comms_compute**: collectives on stream 0, GEMMs on streams 1..N
+        """
+        with record_function(f"comms_compute_overlap::{self.mode}"):
+            if self.mode == "comms_only":
+                self._run_comms(streams)
+            elif self.mode == "compute_only":
+                self._run_compute(streams, compute_stream_offset=0)
+            else:
+                self._run_comms(streams)
+                self._run_compute(streams, compute_stream_offset=1)
```
amd-vivekag, line 316: do we need synchronization between these 2 calls?

amd-vivekag (commented):
Just one minor comment, everything else looks good to me.

src/aorta/hw_queue_eval/workloads/distributed/comms_compute_overlap.py
vivekkhandelwal1: Hi Vivek, no synchronization is needed here. The two calls dispatch work onto independent hip streams and return immediately to the CPU. The comms work on stream 0 and the GEMM work on streams 1..N execute concurrently on the GPU, which is the overlap we're measuring. Adding a sync between them would serialize the execution and defeat the purpose of the `comms_compute` mode. Iteration-level synchronization is handled downstream by `_complete_pending_ops()` and the harness's `sync_all_streams`.

amd-vivekag (approved):
LGTM

Issue #151, opened 2026-04-27
B2: aorta triage run -mode matrix


oyazdanb:
## Goal

Productize a contingency-table style **triage matrix**: for a given workload, run a fixed set of `(mitigation × environment)` cells, count failures and step-times per cell, and produce a `matrix.md` table that flags cells where a mitigation may be working only via a speed confound.

Driven by either a checked-in recipe file (primary) or ad-hoc flags (shim) — both modes converge on the same `run_recipe(recipe)` execution path.

## Why this matters — speed-confound detection

Investigations into mitigation effectiveness routinely run into the same trap: a change that eliminates a numeric failure also slows down GPU execution. Without measuring step-time alongside pass/fail, the matrix lies about which mitigations "fix" the bug versus which ones merely delay it long enough to not observe.

Concrete pattern from prior internal RCAs:
- A mitigation that disables TF32 eliminates a NaN — but is ~25% slower.
- A mitigation that toggles XNACK eliminates the same NaN — and is *not* measurably slower.

The first reads "speed confound suspected; verify with profiler before drawing conclusions"; the second reads "trust this cell." Encoding this distinction in the matrix output is the core value-add over a hand-maintained spreadsheet.

The lesson:
- Capture step-time per trial alongside pass/fail.
- Compute baseline step-time from the (no-mitigation × baseline-environment) cell.
- Flag any cell where `cell_step_time / baseline_step_time > 1.15` as a potential speed confound.

## Design shape — recipe primary, flags secondary

A triage matrix run is two pieces of information:

1. **Which cells to run** — the cartesian product (or a hand-picked subset) of mitigation × environment, plus per-cell trial/step counts.
2. **What investigation it belongs to** — drives output grouping; ties weeks of `matrix.md` artifacts back to the originating ticket.

A YAML/JSON **recipe file** captures both, lives in version control, and gives users a one-line invocation: `aorta triage run --recipe recipes/<name>.yaml`. Flags are kept for ad-hoc one-shots and as an escape hatch — internally they construct an in-memory recipe and reuse the same execution path.

Recipe names resolve through `aorta.registry` (B3): `mitigations: [tf32_off, xnack]` is a list of registered names (built-in or contributed via `aorta.mitigations` entry-points). Same for `environment: <name>`. Unknown names raise B3's `UnknownMitigationError` / `UnknownEnvironmentError`.

## Files to create / modify

```
[...]

PR #160, opened 2026-05-05
B2: triage run -- mitigation x environment matrix runner

amd-vivekag:
Implements task B2 (issue #151): the `aorta triage run` command that sweeps a (mitigation x environment x trials) matrix in-process via B1's `run_trials`, aggregates per-cell stats, classifies each cell against a baseline with speed-confound detection, and emits matrix.md / matrix.json / recipe.resolved.yaml under a ticket-grouped output directory.

Scope (matrix mode only per design D11; optimize mode deferred to P1):

* `triage/recipe.py` -- schema v1, YAML/JSON loader, and the `build_recipe_from_flags` shim. `Recipe`/`Cell`/`InlineEnv`/ `ConfoundCfg` are frozen dataclasses; loader rejects unknown schema_version, unknown keys, duplicate cell names, and resolves every `(mitigation, environment)` reference through B3's registry (built-ins + entry-point plugins + B3.1 sidecars). Inline-docker envs auto-name as `_inline_<8char-blake2b>` so two cells with the same image ref collapse to one environment probe.

* `triage/runner.py` -- `run_recipe()` is the single execution path both modes converge on. Resolves `<output>/<ticket>/<workload>/<ts>/`, writes inline-docker sidecar JSON for B1's resolver to consume, captures host + per-environment `collect_env` snapshots, dispatches every cell via in-process `run_trials`. Per-cell exceptions are caught and surfaced as an `error` row so a single bad cell doesn't kill the matrix. Strictly no `subprocess` -- a grep test under `tests/triage/` enforces it.

* `triage/matrix.py` -- `aggregate_cell()` collapses per-trial results into median/p90/min/max step time, NaN counts, exit-status histogram, and the per-cell label set.

* `triage/confound.py` -- `resolve_baseline()` (explicit override -> first `baseline-*` cell -> first `mitigations: [none]` cell) and `classify_all()` (cell_step_time / baseline_step_time > threshold -> `speed (+N%)` flag). Threshold default 1.15 per D11.

* `triage/output.py` -- writes the three artifacts. matrix.md is human-first (table with per-row label badges + run summary); matrix.json is machine-first (full per-cell stats and the resolved recipe round-trips into recipe.resolved.yaml so a matrix can be re-run byte-identically).

* `cli/triage.py` -- two equivalent entry points: `--recipe <file>` (primary) and `--mode matrix --workload ... --mitigation-axis ... --environment-axis ...` (flag shim). Mixing the two raises a UsageError up-front instead of silently letting one win. Adds `list-mitigations` / `list-environments` discovery subcommands (delegates to B3's resolver, prints source_package so users can see whether a mitigation came from `aorta`, a plugin, or a sidecar).

Recipes:

* `recipes/README.md` -- full schema reference + worked examples.
* `recipes/example-fsdp-smoke.yaml` -- minimal end-to-end example.

Tests (1364 lines, 5 files):

* `test_recipe.py` -- schema validation: unknown version, unknown top-level keys, duplicate cell names, missing required fields, inline-env hash determinism, sidecar resolution.
* `test_runner_b1_api.py` -- per-cell error isolation, in-process guarantee (greps `triage/runner.py` for `subprocess`), inline-env sidecar round-trip, per-environment snapshot caching.
* `test_matrix.py` -- aggregation against synthetic TrialResult fixtures incl. the all-NaN and mixed-exit-status edges.
* `test_confound.py` -- baseline resolution priority and the threshold classifier (boundary at exactly 1.15).
* `test_output_layout.py` -- artifact paths, matrix.md table shape, matrix.json schema, resolved-recipe round-trip.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


src/aorta/triage/matrix.py
```diff
@@ -0,0 +1,258 @@
...
+            isinstance(iters, int)
+            and iters > 0
+            and isinstance(elapsed, (int, float))
+            and elapsed > 0
+        ):
+            return [float(elapsed) / iters * 1000.0]
+    wall = getattr(trial, "wall_clock_sec", 0.0) or 0.0
+    if wall > 0 and effective_steps > 0:
+        return [float(wall) / effective_steps * 1000.0]
+    return []
```
oyazdanb: `_step_times_from_trial` falls back through three different timing signals:

1.` result["step_times_ms"]`
2. `elapsed_sec / total_iterations`
3. `wall_clock_sec / effective_steps`

The third branch folds setup/teardown into the “step time”. If the baseline uses branch 1 but another cell only reaches branch 3, the Confound column is comparing different kinds of numbers. That will mislead as soon as we mix workloads like fsdp (per-step timing) with meta_recom_repro (only wall-clock available).

record which branch supplied the timing (e.g. `step_time_source: "per_step" | "elapsed_per_iter" | "wall_clock_total"` on CellStats, surfaced in matrix.json). Optional but nice: if the baseline and the current cell come from different sources, classify as n/a instead of pretending the ratio is meaningful. Not a release blocker, but landing this before we run a mixed matrix would save the first NaN investigation a lot of confusion.

oyazdanb (approved):

Issue #182, opened 2026-05-19
A1.4-platform-buck-tier


amd-vivekag:
## Goal

Add a first-class `buck_target` field to `aorta.registry.Environment` so any workload (public or private) can opt into Buck-tier subprocess invocation the same way today's workloads opt into Docker-tier via `Environment.docker`.

This is the platform half of the A1.4 Buck-tier arc. Workload-side adoption lands in `aorta-internal#TBD` (recom_repro pilot) and the regression-gate side lands in `aorta-internal#TBD`. This issue is independently useful — once it ships, any workload author can adopt Buck-tier by reading one config key.

Issue #163 (`A1.2-env-probe-buck`) taught `aorta env probe` to *read* Buck state; A1.4 teaches workloads to *run under* Buck. This is the platform contract for that.

## Why this matters

Today the dispatcher threads the resolved `Environment` into the workload via `config["_aorta_environment"] = asdict(env_descriptor)` (see `src/aorta/run/dispatcher.py` line 308). Workloads that can isolate themselves read the dict and branch their subprocess invocation. `recom_repro` does this today with `Environment.docker`:

```python
image = (self.config.get("_aorta_environment") or {}).get("docker")
argv = _build_argv(image, env, self._entry)  # docker run vs python
```

Adding a `buck_target` peer field follows the same pattern. The dispatcher needs no change (it already round-trips the whole dataclass). The wrapper contract gains one new key any workload can opt into.

## Deliverables

- `src/aorta/registry/types.py` — add `buck_target: str | None = None` to `Environment` as a peer of `docker` / `venv`. Update the docstring to note the field is interpreted by Buck-aware workload wrappers (analogous to how `docker` is interpreted by docker-aware wrappers); the platform itself does not launch Buck.
- `src/aorta/registry/environments.py` — accept the new field in the registry loader; existing built-in `Environment` entries (`local`, etc.) leave it as the default `None`.
- `src/aorta/run/dispatcher.py` — no logic change required (the existing `asdict(env_descriptor)` call already round-trips). Add a note in the inline comment around line 291-308 listing `buck_target` alongside `docker` as a recognized tier hint.
- `docs/workloads.md` (or wherever the workload-author guide lives) — add a "Buck-tier subprocess invocation" section showing the `config["_aorta_environment"]["buck_target"]` read pattern. Reference the recom_repro pilot (`aorta-internal#TBD`) once it lands as the canonical example.
- Tests in `tests/registry/test_environments.py` (or equivalent):
[...]

Issue #163, opened 2026-05-06
A1.2-env-probe-buck


oyazdanb:
## Goal

Extend `aorta env probe` to produce a meaningful env.json snapshot when invoked inside a [Buck2](https://buck2.build/) build environment. Today the probe assumes ROCm and supporting libraries are discoverable via system package managers (apt / dnf / `pkg-config`) or via Docker image digests. Inside a Buck2 monorepo, those signals are absent — libraries are Buck targets, the runtime root is a Buck repo, and source revisions live in hg or git via Buck's repo root, not in `/etc`.

Three sub-features:

1. **Build-system metadata block** — detect Buck2 presence and capture buck2 version, repo root, and source revision into a new `build_system` block in env.json. Coexists with the existing Docker / library-introspection blocks; populated when in Buck2, `{kind: "none"}` otherwise.
2. **Buck-aware library introspection** — when in a Buck2 env, resolve hipBLASLt / PyTorch / RCCL via `buck2 audit dependencies` against a user-supplied target. Produces the same `library_introspection` shape A1 already defines, with `source: "buck"` instead of `source: "pkg-config"` / `source: "elf"`.
3. **Buck recipe emitter** — `aorta env recipe --format buck <env.json>` emits a `BUCK` / `TARGETS` file fragment best-effort reproducing the captured env. Best-effort, not exact reconstruction (env.json captures observed state, not a complete build recipe).

```bash
# Sub-feature 1 + 2: probe inside a Buck2 env
aorta env recommended-tools          # buck2 listed if not on PATH
aorta env probe --buck-target //my:trainer
# → env.json with populated build_system block + library_introspection sourced from buck2 audit

# Sub-feature 3: emit a Buck recipe from a captured env.json
aorta env recipe --format buck env.json > BUCK.aorta-recipe
# Output is a BUCK file fragment (target declarations like `prebuilt_cxx_library(...)`),
# NOT a .bzl macro/rule file. Append the fragment to your project's BUCK file or
# load it with `include_defs("//path:BUCK.aorta-recipe")` in a parent BUCK.
```

## Why this matters

[...]

PR #42, opened 2025-12-12
Initial workflow file for gemm-sweep-analysis

prosenjitdhole:
This is an initial version of GH action for gemm-sweep analysis

Note : YML uses requirements.txt, but I have added this file in this review as it will cause conflict. 

TODO : 
- Cretae rccl run action (rccl related changes are still on the way : #41 )

- Find out a way to dump comparing html
   - Need to figure out a way to compare the baseline run
   - Create a single config tracelens run workflow

- Need a runner machine and update the yml accordingly

PR #191, opened 2026-05-25
cli/run: add --buck-target overlay for the resolved Environment (#182 follow-up)

amd-vivekag:
## Summary

Adds `--buck-target` to `aorta run`, the run-side peer of the existing `aorta env probe --buck-target` (#163, A1.2b). #182 made `buck_target` a first-class peer of `docker` / `venv` on the `Environment` dataclass, and the registry / sidecar / dispatcher already thread it through to `config["_aorta_environment"]["buck_target"]` where Buck-aware workload wrappers read it. The piece this PR fills in: a CLI flag to **overlay** `buck_target` per-run, so operators don't have to register a one-shot named environment per Buck pin.

The overlay is a single-axis pin: it overrides only the `buck_target` field on the resolved `Environment`; `docker`, `venv`, and `source_package` from the named `--environment` are preserved. `--buck-target=<unset>` (the default `None`) is a no-op so every existing invocation continues to behave exactly as before.

## Why now

aorta-internal #42 (regression-gate Buck tier, currently open) ships a dispatcher that emits the argv:

- BUCK_ONLY: `aorta run --workload <wl> --buck-target <target>`
- BUCK_IN_DOCKER: `docker run --rm <digest> aorta run --workload <wl> --buck-target <target>`

Before this PR both tiers failed at exec time with `Error: No such option '--buck-target'` -- the exact "loud at the wrong layer" failure mode that PR is designed to prevent for the schema. The dispatcher cites #182 as the field source and expects `--buck-target` on `aorta run`; the field landed, the CLI flag didn't.

End-to-end after this change (against a `buck2 build //:aorta` binary):
- `buck2 run //:aorta -- run --help` shows the new option.
- `buck2 run //:aorta -- run --workload <real> --buck-target //:aorta` parses successfully; failure (if any) now surfaces at workload discovery / setup, not at CLI parse.

## Changes

| File | Change |
|---|---|
| `src/aorta/run/dispatcher.py` | `RunRequest` gains `buck_target: str \| None = None` (between `environment` and `mitigations`). `run_trials` applies the override via `replace(env_descriptor, buck_target=request.buck_target)` immediately after `get_environment(...)` resolves the named env. |
| `src/aorta/cli/run.py` | New `--buck-target TEXT` click option, threaded into `RunRequest`. Handler body grows by 1 line; `TestCliHandlerIsThinShell::test_body_is_short` (the ~30-line shell budget) still passes. |
| `tests/run/test_cli_parsing.py` | New `TestBuckTargetOption` class: 5 tests (option appears in `--help`; default is no-op; canonical label parses; root-package label parses; missing value rejected). |
| `tests/run/test_dispatcher.py` | New `TestBuckTargetOverride` class: 4 tests (override sets value on empty-pin env; override replaces named env's value; `None` preserves named env's value; override preserves all other `Environment` fields). Plus extended `TestRunRequest::test_default_values` to pin `buck_target is None` default. |

128/128 tests in `tests/run/` pass (was 119 before; +9 new).

## Test plan

- [ ] Reviewer runs `PYTHONPATH=$PWD/src python3 -m pytest tests/run/ -q` and confirms 128 passing.
- [ ] Reviewer runs `buck2 build //:aorta && buck2 run //:aorta -- run --help` and confirms `--buck-target` is documented in the option list.
- [ ] Reviewer runs `buck2 run //:aorta -- run --workload <existing-workload> --buck-target //:aorta` and confirms the flag is accepted (failure, if any, is workload-discovery / setup, not `No such option '--buck-target'`).
- [ ] Reviewer confirms backward-compat: any existing invocation that omits `--buck-target` produces identical behavior to pre-PR (the new param defaults to `None`, the `replace(...)` overlay is gated on `is not None`).

## Cross-repo follow-up

Unblocks aorta-internal #42's BUCK_ONLY and BUCK_IN_DOCKER regression-gate tiers, which can now actually execute end-to-end against this binary instead of just generating argv. No aorta-internal changes required by this PR -- the gates module's dispatcher already emits the correct argv shape.


tests/run/test_dispatcher.py
```diff
@@ -966,6 +970,216 @@ def cleanup(self):
...
+        assert results[0].execution_env["buck_target"] == "//:aorta"
+
+    def test_none_override_preserves_named_env_buck_target(self, tmp_path):
+        """``buck_target=None`` (the default) is a no-op.
+
+        Backward-compat guarantee: a pre-existing invocation that
+        relied on a named env's declared ``buck_target`` must
+        continue to see that value flow into
+        ``_aorta_environment``. Otherwise this CLI addition would
+        silently broken every existing buck-aware named env.
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in 6477071. Changed "would silently broken" -> "would silently break".

src/aorta/run/dispatcher.py
```diff
@@ -110,6 +110,22 @@ class RunRequest:
     workload: str
     trials: int
     environment: str = "local"
+    # Runtime override for the resolved :class:`Environment`'s
+    # ``buck_target`` field (#182 made it a first-class peer of
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in 6477071. Added a `buck_target` entry to the `Attributes:` section, inserted between `environment` and `mitigations` to match the source-order grouping (env-tier overlay first, then mitigation set). The entry covers the override/no-op semantics and includes an explicit **Keyword-only** callout.

While addressing this I also proactively applied the `kw_only=True` fix from the symmetric --image PR (#193, comment [3298058572](https://github.com/ROCm/aorta/pull/193#discussion_r3298058572)) to `buck_target` here -- same backward-compat principle (declaring `buck_target` before `mitigations` in source would shift `mitigations` from positional slot 4 to 5, silently breaking external positional callers of `RunRequest`). Pinned by a new `TestBuckTargetIsKeywordOnly` class (3 tests) so a future refactor can't silently regress it.

Test suite: 128 -> 131 passing (+3 kw_only pins, no regressions). End-to-end against `buck2 run //:aorta` still parses `--buck-target` cleanly.

tests/run/test_dispatcher.py
```diff
@@ -966,6 +970,289 @@ def cleanup(self):
...
+
+    def test_mitigations_is_still_positional_at_slot_4(self):
+        """Sanity: ``mitigations`` remains positionally addressable
+        as the 4th argument despite ``buck_target`` being declared
+        before it in the source. The whole point of ``kw_only=True``
+        on ``buck_target`` is to keep this true.
+
+        Pre-PR-#191 behavior: ``RunRequest("wl", 1, "env",
+        ("none",))`` constructed with ``mitigations=("none",)``.
+        Post-PR-#191 behavior MUST be identical.
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in b9f4244. Re-phrased to "Behavior before the `buck_target` field was added" / "Behavior after this field was added (this PR, the #182 follow-up that introduced the `--buck-target` overlay)". Names the actual change being pinned (the field addition) rather than the PR number, while still anchoring to this PR for context.

tests/run/test_cli_parsing.py
```diff
@@ -344,3 +344,95 @@ def test_unknown_mitigation_error(self):
...
+            [
+                "--workload",
+                "nonexistent_workload",
+                "--buck-target",
+                "//:aorta",
+            ],
+        )
+        assert "Unknown option" not in result.output
+        assert "No such option" not in result.output
+
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in b9f4244. Added the matching positive signal (`exit_code != 0` AND `"not found" in output`) that the sibling `test_buck_target_accepts_canonical_label` test already pins. The two tests now have identical assertion shapes, so a click-plumbing regression trips both consistently.

src/aorta/cli/run.py
```diff
@@ -39,6 +39,28 @@
...
+    type=str,
+    default=None,
+    help=(
+        "Buck2 target label (e.g. '//workloads/recom_repro:recom_repro') "
+        "to overlay onto the resolved environment's ``buck_target`` "
+        "field at run time (#182). Peer of the docker / venv tier hints: "
+        "the platform threads it, Buck-aware workload wrappers consume "
+        "it via ``config['_aorta_environment']['buck_target']`` and "
+        "decide whether to shell out via ``buck2 run <label>`` (see "
+        "registry/README.md). Other fields of the named --environment "
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in b9f4244. Corrected to `src/aorta/registry/README.md` (the actual path in the source tree). Re-phrased to "in the source tree" so a reader running `aorta run --help` from an installed wheel doesn't try to look for a `src/` directory under their site-packages.

src/aorta/cli/run.py
```diff
@@ -39,6 +39,29 @@
...
+        "``src/aorta/registry/README.md`` in the source tree). Other "
+        "fields of the named --environment "
+        "(docker, venv, source_package) are preserved -- the override "
+        "is a pin on the Buck axis only. When omitted, the named "
+        "environment's existing ``buck_target`` (if any) is used as-is. "
+        "Symmetric to ``aorta env probe --buck-target`` (#163, A1.2b), "
+        "which enriches the env snapshot via ``buck2 cquery``; this "
+        "flag instead pins the runtime recipe so the workload can act "
+        "on it. Required by aorta-internal #42's BUCK_ONLY and "
+        "BUCK_IN_DOCKER regression-gate tiers."
```
oyazdanb: This is long for --help and has reference to issue. It should be short and has the info user need to use this flag.

src/aorta/run/dispatcher.py
```diff
@@ -209,6 +256,18 @@ def run_trials(request: RunRequest) -> list[TrialResult]:
...
+    #     resolving the named environment, so the named env's
+    #     ``docker`` / ``venv`` / ``source_package`` fields are
+    #     preserved.  ``None`` means "no override" -- a named env that
+    #     already declares ``buck_target`` keeps its value.  This is
+    #     the dispatcher side of the ``--buck-target`` CLI flag
+    #     symmetric to ``aorta env probe --buck-target``; see the
+    #     ``RunRequest.buck_target`` docstring for the cross-repo
+    #     motivation (aorta-internal #42).
+    if request.buck_target is not None:
+        env_descriptor = replace(env_descriptor, buck_target=request.buck_target)
```
oyazdanb: replace it with:

```
if request.buck_target:
          env_descriptor =replace(env_descriptor,buck_target=request.buck_target)
```

so "" silently does not continue.

src/aorta/run/dispatcher.py
```diff
@@ -51,6 +51,24 @@ class RunRequest:
...
+            and every pre-existing ``RunRequest`` invocation behaves
+            unchanged. Symmetric peer of ``aorta env probe
+            --buck-target`` (#163). Threaded into
+            ``config['_aorta_environment']['buck_target']`` for the
+            workload's wrapper to consume. **Keyword-only**: this
+            field is declared with ``kw_only=True`` so adding it
+            before existing positional fields like ``mitigations``
+            does NOT shift the positional ``__init__`` signature
+            (positional callers continue to interpret the 4th arg as
+            ``mitigations``).
```
oyazdanb, line 80: Add one sentence to the docstring: "Note: TrialResult.execution_env records the effective recipe (declared env + any runtime overlays from CLI flags). Replays should read execution_env field-by-field, not look up the named environment, when overlays were used."

src/aorta/cli/run.py
amd-vivekag: Fixed in df1fe47. Trimmed to 5 lines focused on what the user needs to USE the flag: example label, what it overlays, what it preserves, what omitting it does. Removed all four cross-repo issue refs (#182 / #163 / A1.2b / aorta-internal #42) -- the design rationale they pointed to already lives in the `RunRequest.buck_target` docstring and the inline source comment at the call site, so `--help` was duplicating it. The new shape is consistent with the sibling options in the same file (`--workload`, `--trials`, `--mitigations`, etc.). Rendered `aorta run --help` confirms the new block is sibling-shaped.

src/aorta/run/dispatcher.py
amd-vivekag: Fixed in df1fe47. Switched to the truthy check `if request.buck_target:` so both `None` and `""` are no-ops -- empty string is never a valid Buck2 label, and silently overlaying `buck_target=""` onto the resolved env would land downstream as a `buck2 run ""`-style failure that's hard to attribute back to the flag.

Pinned by a new `test_empty_string_override_preserves_named_env_buck_target` test that mirrors the existing `None`-no-op test (131 -> 132 passing, no regressions on the existing 131). Inline source comment updated to call out the falsy guard explicitly.

I'll mirror this change to `--image` on the symmetric PR #193 when it rebases -- same pattern, same reasoning.

src/aorta/run/dispatcher.py
amd-vivekag: Fixed in df1fe47. Added the replay note to the `RunRequest` class-level docstring (placed above the `Attributes:` block rather than under the `buck_target` attribute specifically) so it also covers the upcoming `image` overlay on #193 -- the same reasoning applies to every CLI-supplied env-axis pin, present or future.

Wording stays close to your suggested sentence:

> Note:
>     `TrialResult.execution_env` records the *effective* recipe (the declared environment plus any runtime overlays from CLI flags such as `--buck-target`). Replays should read `execution_env` field-by-field, not re-resolve the named environment, when overlays were used -- otherwise the replay silently drops the overlay and runs against a different recipe than the original trial.

PR #210, opened 2026-06-04
race/fsdp: shared-weight transformer with per-kernel int16 checksum verification

mycpuorg:
## Summary

The AINIC GDR flush SDC recipe (`ainic-gdr-flush-sdc.yaml`) used `compute_type: gemm` with independently-random weight matrices per layer.  This makes cross-layer comparison meaningless — gradient/activation differences across layers could be numerical divergence rather than corruption.

This PR switches to a **synthetic identical-layer transformer model**: all layers share one weight matrix (fixed seed) and receive the same fixed reference input, so every layer's forward output is analytically identical.  Any divergence across layers is a corruption signal, not numerical noise.

### Per-kernel int16 checksums

Following the bitwise checksum design from the onsite whiteboard (Jun 2 2026): each tensor is reinterpret-cast from bf16 to int16 and summed in int64 — zero information loss, no float rounding, catches any bit flip.

Four checksums are recorded per layer, wrapping both kernels:

| Checkpoint | What it checksums |
|---|---|
| `comm_input` | param shard **before** all_gather |
| `comm_output` | full_param **after** all_gather |
| `compute_input` | reference_input **before** GEMM |
| `compute_output` | activation **after** GELU |

`_verify_layer_checksums()` compares all four across layers.  The diagnostic pinpoints **where** corruption entered:
- `comm_output` diverges, `comm_input` matches → corruption in the collective (RCCL/NIC path)
- `compute_output` diverges, `comm_output` matches → corruption in GPU compute (ALU)

### Changes

- **`src/aorta/race/base.py`**: Add `bf16`, `fp16`, `fp32` as dtype short-form aliases
- **`src/aorta/race/config.py`**: Add `shared_layer_weights: bool` field to `ReproducerConfig`
- **`src/aorta/race/modes/fsdp.py`**:
  - `_checksum()`: `tensor.view(int16).to(int64).sum()` — bitwise, lossless
  - `setup_buffers()`: shared weight matrix (seed=0) + fixed reference input (seed=1) when `shared_layer_weights=True`
  - `_forward_layer()`: records comm and compute input/output checksums per layer
  - `_verify_layer_checksums()`: cross-layer comparison of all four checksum keys
- **`recipes/ainic-gdr-flush-sdc.yaml`**: Switch to `compute_type: transformer`, `shared_layer_weights: true`, `dtype: bf16`; drop dead `gemm_size`/`gemm_layers` fields

## Test plan

- [ ] `aorta triage run --recipe recipes/ainic-gdr-flush-sdc.yaml --dry-run` passes with no "ignoring unknown" warnings
- [ ] Single-node smoke run confirms all layer checksums match (zero corruption on clean hardware)
- [ ] Inject synthetic corruption (flip bits in one layer's all_gather output) and verify `LAYER_CHECKSUM_MISMATCH (comm_output)` is logged with correct layer index
- [ ] Existing `_verify_all_gather` / `_verify_reduce_scatter` pattern checks still fire independently

### DIFF
--- a/src/aorta/race/modes/fsdp.py
+++ b/src/aorta/race/modes/fsdp.py
@@ -77,6 +77,10 @@ class FSDPModeReproducer(BaseReproducer):
         self.activation: Optional[torch.Tensor] = None
         self.grad_buffer: Optional[torch.Tensor] = None
 
+        # Shared-weight transformer: fixed reference input + per-layer checksums
+        self.reference_input: Optional[torch.Tensor] = None
+        self.layer_checksums: List[Optional[dict]] = []
+
     def _setup_compute(self) -> None:
         """
         Override base compute setup -- FSDP manages its own per-layer compute.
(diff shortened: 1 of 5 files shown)


src/aorta/race/modes/fsdp.py
```diff
@@ -166,30 +187,78 @@ def _fill_patterns(self) -> None:
...
+    def _checksum(tensor: torch.Tensor) -> int:
+        """
+        Bitwise checksum: reinterpret-cast to int16 and sum.
+
+        bf16 (or any 16-bit dtype) is viewed as int16 so every bit pattern
+        contributes to the checksum with zero information loss -- no float
+        rounding, no abs(), and NaN / denorm bit patterns are included.
+        Accumulation is done in int64 to avoid overflow.
+        """
+        return tensor.view(torch.int16).to(torch.int64).sum().item()
```
Alkaid-Benetnash, line 200: Just to point out that overflow is expected. And since this checksum only cares about the binary value, not integer semantics. int16 or int64 affects more of "hash-collision" rather than "overflow".

PR #225, opened 2026-06-12
probe: discard latched tier2:hang flag on clean subprocess exit

### LINKED ISSUE #224: aorta probe: false-positive tier2:hang on wrapper-delegated (docker/sudo) workloads
## Summary

`aorta probe` can emit a **false-positive `tier2:hang`** verdict for a
wrapped command that delegates its real work to a child process tree
(e.g. `sudo → bash launch.sh → docker run → container → python3`). The
trial exits `0`, does not time out, and the workload reports success —
yet it is classified `fail / tier2:hang`, and the reproduction matrix
rolls that up into an authoritative-looking `100% hang` cell.

A run that voluntarily exits `0` within its timeout cannot have been
hung. The latched hang flag is never reconciled against the process
exit, so a clean baseline and a real failure render identically.

- **Component:** `aorta probe` — Tier-2 hang classifier (`src/aorta/probe/classifier/tier2_hang.py`) + subprocess workload (`src/aorta/workloads/_subprocess.py`)
- **Severity:** high — corrupts the primary deliverable (the reproduction matrix)

---

## Reproduction (representative `result.json`)

```json
{
  "verdict": "fail",
  "exit_code": 0,
  "walltime_sec": 2561.95,
  "timed_out": false,
  "failure_detectors_fired": ["tier2:hang"]
}
```

`exit 0`, `timed_out: false`, the workload's own stdout reports a clean
run — yet `tier2:hang` fired. The same command run standalone completes
cleanly with no hang.

---

### 1. The hang monitor is blind to wrapper-delegated workloads

`HangMonitor` decides "hung" from a **two-of-three** predicate:

- `stdout_silent` — no writes to the trial's `stdout.log` for `hang_window_sec`
- `io_idle` — `/proc/<pid>/io` `rchar+wchar` unchanged for the window
- `gpu_idle` — `amd-smi` reports the busiest GPU idle

`_subprocess.py` wires the monitor to the **wrapper** PID (`proc.pid`)
and the wrapper's `stdout.log`. When the wrapper delegates everything to
a child process tree (docker/sudo), the wrapper prints a few summary
lines then goes silent while the descendants do the work
(`stdout_silent = True`), and the wrapper PID itself does almost no I/O
(`io_idle = True`, since the work happens under a different PID tree
invisible to `/proc/<wrapper_pid>/io`). That is already **2 of 3**, so
`tier2:hang` fires regardless of GPU state — even though the GPUs were
busy.

### 2. The latched hang flag is never reconciled against exit

`_subprocess.py` reads `hang_detected = bool(hang_monitor and
hang_monitor.hang_detected)` and passes it straight into
`classify_trial(...)`. There is **no check** that the process actually
exited abnormally. `result.json` confirms `exit_code: 0, timed_out:
false` coexisting with `tier2:hang`. A hang flag latched mid-run should
be discarded when the process later exits `0` within the timeout.

---

## Proposed fix

**Reconcile the hang flag against exit (smallest, highest-value fix).**
In `_subprocess.py`, after `proc.wait(...)`, gate the latched flag:

```python
# A process that voluntarily exited 0 within the timeout was not hung,
# no matter what the in-flight monitor latched. The monitor's stdout/io
# legs are blind to work delegated to a child tree (docker/sudo), so a
# mid-run latch is advisory only.
hang_detected = bool(hang_monitor and hang_monitor.hang_detected)
if exit_code == 0 and not timed_out:
    hang_detected = False
```

This turns a clean baseline from `tier2:hang` back into a correct `pass`,
while genuine hangs (`timed_out=True`) and crashes (`exit_code != 0`)
keep the flag.

### Follow-ups (out of scope for the first fix, tracked here)

- Make the hang predicate robust to wrapper-delegated workloads
  (monitor the descendant process tree, or require the GPU leg to
  participate rather than letting the two cheap legs trip it alone).
- Distinguish matrix outcome classes (reproduced / hang / workload
  failure / infrastructure failure / incomplete artifacts) so a clean
  run never renders as a reproduction.

---

## Acceptance criteria

- [ ] A wrapped command that exits `0` within the timeout is **never** classified `tier2:hang`.
- [ ] A genuine hang (timed out) or a non-zero exit still preserves the hang signal.
- [ ] Regression tests pin the reconciliation behavior.

amd-vivekag:
## Summary

Fixes #224.

`aorta probe` could emit a **false-positive `tier2:hang`** for a wrapped
command that delegates its real work to a child process tree
(`sudo → bash launch.sh → docker run → container → python3`). The trial
exited `0`, did not time out, and the workload reported success — yet it
was classified `fail / tier2:hang`.

- The Tier-2 `HangMonitor` watches three signals on the **wrapper** PID:
  stdout silence, `/proc/<pid>/io` idleness, and GPU idleness. For a
  wrapper-delegated workload the parent goes quiet and does almost no
  I/O of its own while the descendants do the work, tripping
  `stdout_silent` + `io_idle` (two of three) and latching `hang_detected`
  on a healthy run.
- The latched flag was passed straight into `classify_trial(...)` with
  **no reconciliation** against the actual exit.

This change gates the latched flag in `SubprocessWorkload.run()`: a
process that voluntarily exited `0` within its timeout was not hung, so
the advisory flag is dropped. Genuine hangs (`timed_out=True`) and
crashes (`exit_code != 0`) keep the flag and still classify as `fail`.

The larger follow-ups from #224 (descendant-tree-aware hang predicate,
distinct matrix outcome classes) are intentionally out of scope here to
keep this a small, surgical correctness fix.

## Changes

- `src/aorta/workloads/_subprocess.py`: reconcile `hang_detected`
  against `exit_code == 0 and not timed_out` after `proc.wait(...)`;
  add a module logger and an `info`-level breadcrumb when the flag is
  discarded.
- `tests/probe/test_subprocess_workload.py`: 3 regression tests.

## Test plan

- [x] `test_clean_exit_discards_latched_hang_flag` — exit 0 within timeout, latched flag → dropped, verdict `pass`, no `tier2:hang`.
- [x] `test_nonzero_exit_keeps_latched_hang_flag` — exit != 0, latched flag preserved, `tier2:hang` fires.
- [x] `test_timed_out_keeps_latched_hang_flag` — real timeout, latched flag preserved, `tier2:hang` fires.
- [x] Full probe suite green (`pytest tests/probe`, 311 passed).

### DIFF
--- a/src/aorta/workloads/_subprocess.py
+++ b/src/aorta/workloads/_subprocess.py
@@ -41,6 +41,7 @@ Phase-1 minimum shape keeps working.
 from __future__ import annotations
 
 import json
+import logging
 import os
 import re
 import stat
@@ -65,6 +66,8 @@ from aorta.probe.classifier.tier3_kernel import (
 )
 from aorta.probe.classifier.verdict import Verdict
 from aorta.workloads._base import Workload, WorkloadResult
+
+log = logging.getLogger(__name__)
 
 # Process-wide Tier 3 state. Shared across every SubprocessWorkload
 # instance the dispatcher constructs over one ``aorta probe`` invocation
@@ -428,6 +431,36 @@ class SubprocessWorkload(Workload):
         # are bugs, not trial outcomes).
         log_text = _read_log_text(stdout_path, stderr_path)
         hang_detected = bool(hang_monitor and hang_monitor.hang_detected)
+        # Reconcile the latched hang flag against the actual exit.
+        #
+        # The in-flight HangMonitor watches three signals on the
+        # *wrapper* PID: stdout silence on this trial's stdout.log, I/O
+        # idleness on /proc/<wrapper_pid>/io, and GPU idleness. For a
+        # workload that delegates its real work to a child process tree
+        # the parent can't see (``sudo`` -> ``bash run.sh`` ->
+        # ``docker run`` -> container -> ``python3``), the wrapper goes
+        # quiet and does almost no I/O of its own while the descendants
+        # do all the work. That trips ``stdout_silent`` + ``io_idle``
+        # (two of three) and latches ``hang_detected`` even though the
+        # workload is alive and busy. Those two legs are structurally
+        # blind to delegated work, so a mid-run latch is advisory only.
+        #
+        # A process that voluntarily exited 0 within its timeout was, by
+        # definition, not hung -- a real hang would have either kept the
+        # process alive until ``proc.wait(timeout=...)`` fired
+        # (``timed_out=True``) or surfaced as a non-zero exit. Drop the
+        # latched flag in that case so a clean run is never classified
+        # ``tier2:hang``. Genuine hangs (``timed_out=True``) and crashes
+        # (``exit_code != 0``) keep the flag and still classify as fail.
+        if hang_detected and exit_code == 0 and not timed_out:
+            log.info(
+                "tier2:hang latched mid-run for argv[0]=%r but the process "
+                "exited 0 within the timeout; discarding the advisory flag "
+                "(the monitor's stdout/io legs are blind to work delegated "
+                "to a child process tree).",
+                argv[0] if argv else "<unknown>",
+            )
+            hang_detected = False
 
         # Best-effort ``peak_vram_mib`` from the Tier-3 amd-smi
         # snapshots. We only have two samples (pre + post Popen) so
--- a/src/aorta/workloads/_subprocess.py
+++ b/src/aorta/workloads/_subprocess.py
@@ -452,6 +452,7 @@ class SubprocessWorkload(Workload):
         # latched flag in that case so a clean run is never classified
         # ``tier2:hang``. Genuine hangs (``timed_out=True``) and crashes
         # (``exit_code != 0``) keep the flag and still classify as fail.
+        hang_reconciled_away = False
         if hang_detected and exit_code == 0 and not timed_out:
             log.info(
                 "tier2:hang latched mid-run for argv[0]=%r but the process "
@@ -461,6 +462,7 @@ class SubprocessWorkload(Workload):
                 argv[0] if argv else "<unknown>",
             )
             hang_detected = False
+            hang_reconciled_away = True
 
         # Best-effort ``peak_vram_mib`` from the Tier-3 amd-smi
         # snapshots. We only have two samples (pre + post Popen) so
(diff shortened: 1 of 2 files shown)


src/aorta/workloads/_subprocess.py
```diff
@@ -428,6 +431,36 @@ def run(self) -> WorkloadResult:
...
+        # (``exit_code != 0``) keep the flag and still classify as fail.
+        if hang_detected and exit_code == 0 and not timed_out:
+            log.info(
+                "tier2:hang latched mid-run for argv[0]=%r but the process "
+                "exited 0 within the timeout; discarding the advisory flag "
+                "(the monitor's stdout/io legs are blind to work delegated "
+                "to a child process tree).",
+                argv[0] if argv else "<unknown>",
+            )
+            hang_detected = False
```
oyazdanb, line 464: When we discard the latched `tier2:hang` flag in `_subprocess.py:455-463`, the only trace is the `log.info` line — nothing lands in `result.json`. Since the monitor's false-positive behavior is exactly what the #224 follow-ups  need to study, could we set `hang_reconciled_away = True` in the gate and then add `result_doc["capture"]["tier2_hang_latched_but_reconciled"] = True` after the result dict is built (around line 573)? That leaves a durable per-trial breadcrumb while keeping the verdict a clean pass.

amd-vivekag: Done in 8ea7bd8 — exactly as you suggested. The gate now sets `hang_reconciled_away = True`, and after `result_doc` is built I add `result_doc["capture"]["tier2_hang_latched_but_reconciled"] = True` (before the `result.json` write, so it's persisted) only on the clean-exit reconcile path. Verdict stays a clean pass.

Extended the three reconciliation tests: the breadcrumb is asserted present on clean-exit discard, and absent when the flag is preserved (non-zero exit, timeout). Swept the file — this is the only flag-discard gate, so it's the only place that needed the trace.

src/aorta/workloads/_subprocess.py
```diff
@@ -538,6 +573,13 @@ def run(self) -> WorkloadResult:
             "timed_out": timed_out,
             "env": dict(cell_env_snapshot),
         }
+        # Durable per-trial breadcrumb when a latched tier2:hang was
+        # reconciled away on a clean exit. The verdict stays a clean pass,
+        # but the #224 follow-ups (descendant-tree-aware hang predicate)
+        # need to study exactly these wrapper-delegated false positives, so
+        # leave a trace in result.json rather than only the log.info above.
```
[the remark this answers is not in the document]
amd-vivekag: You're right — fixed in 0d2c61b. Reworded the comment to say only the tier2:hang signal is dropped and that the final verdict still reflects the other tiers (a clean exit can still fail Tier-3/4/5). The breadcrumb itself is unchanged.

oyazdanb (approved):

PR #303, opened 2026-07-29
ci: nightly evaluation harness, dashboard, alerting, and automated bumps

vivekkhandelwal1:
Implements the CI/CD plan in `docs/ci-plan.md` (proposed in #300) end-to-end, as one task. Nothing changes the existing PR gate; this adds the nightly evaluation system.

## What this adds

**Phase 1 — Nightly evaluation harness**
- Installs the **released nightly wheel** (`amd-aorta` rc, chained off the `Nightly wheels` workflow) inside the pinned ROCm container and evaluates the workloads.
- `config/ci/nightly_eval_matrix.yaml` (workload x config), `scripts/ci/eval_lib.py` (pure harvester + comparator), `scripts/ci/nightly_eval.py` (runner → results JSON with build/ROCm metadata).
- `config/ci/regression_baselines.yaml` (expected-outcome baselines; **empty ⇒ record-only** so nightly isn't red before baselines are blessed).

**Phase 2 — Dashboard + history**
- `scripts/ci/gen_dashboard.py`: self-contained HTML dashboard (status table + inline-SVG step-time trends + pass-rate trend), published to the `gh-pages` branch (`results/<date>.json` history) via GitHub Pages.

**Phase 3 — Alerting + broader matrix**
- `scripts/ci/alert_issue.py`: opens/updates one `nightly-regression` issue on failure, closes it when green.
- 8-GPU variants of the distributed workloads added (dtype axis noted as a recipe follow-up).

**Phase 4 — Automated bumps + reproducible deps**
- `.github/dependabot.yml` (pip / docker ROCm digest / actions).
- `scripts/ci/lock_requirements.sh` + `lock-requirements.yml`: exact-pin `requirements.lock` generated **inside the ROCm container**, opened as a PR; nightly-eval installs with it as a constraint when present.

**Phase 5 — Perf gating (opt-in)**
- Baselines are correctness-only by default; `refresh_baselines.py --perf-gate` adds step-time/throughput bounds the comparator enforces.

**Baseline lifecycle:** `refresh-baselines.yml` (dispatch) regenerates baselines on the runner and opens a bless PR — never auto-commits.

## Correctness vs performance
- Correctness is gated (a cell that should pass but errors/fails ⇒ job fails).
- Performance is trend-only by default (captured + charted, not blocking).

## Tests
- 20 CI-gate-safe unit/integration tests (`tests/ci/`) for the harvester, comparator, harness, dashboard render, and alert decision — all green. Dashboard render verified from synthetic results.

## Operating notes (post-merge)
1. Enable **GitHub Pages** (source: `gh-pages` branch) to serve the dashboard.
2. `nightly-eval.yml`'s `workflow_run` trigger activates once on the default branch (dispatch works meanwhile).
3. First nightly is **record-only**; then run **Refresh baselines** on the runner to bless, and optionally **Lock requirements**.

Docs: `docs/ci-nightly-eval.md` (operation guide). Design: `docs/ci-plan.md` (#300).

## Test plan
- [ ] Merge, enable Pages, dispatch **Nightly Evaluation** on the runner (record-only run + dashboard publish).
- [ ] Dispatch **Refresh baselines**; review/merge the bless PR.
- [ ] Confirm a seeded failure files the `nightly-regression` issue.


scripts/ci/refresh_baselines.py
```diff
@@ -0,0 +1,106 @@
...
+                st = metrics.get("mean_step_time_ms")
+                if st:
+                    spec["step_time_ms"] = {"max": round(st * (1.0 + step_time_margin), 4)}
+                tp = metrics.get("throughput") or {}
+                if tp:
+                    spec["throughput"] = {
+                        k: {"min": round(v * (1.0 - throughput_margin), 4)}
+                        for k, v in tp.items()
+                        if v
+                    }
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in 1f74cc2: `build_baselines` now uses `is not None` checks for step time and throughput, so a legitimate `0.0` metric still emits bounds under `--perf-gate`.

scripts/ci/gen_dashboard.py
```diff
@@ -0,0 +1,180 @@
...
+Usage:
+    python scripts/ci/gen_dashboard.py --results-dir site/results --out-dir site
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+from pathlib import Path
+from typing import Any
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in 1f74cc2: `gen_dashboard` now imports `html.escape` and escapes all untrusted result fields (reasons, workload::cell keys, build/version strings, generated_at) before interpolating into HTML. Added a test asserting a `<script>` reason is escaped.

scripts/ci/alert_issue.py
```diff
@@ -0,0 +1,113 @@
...
+        f"- Build: aorta `{build.get('amd_aorta_version','?')}`, "
+        f"torch `{build.get('torch','?')}`, ROCm `{build.get('rocm','?')}`",
+        f"- Generated: {results.get('generated_at','')}",
+    ]
+    if run_url:
+        lines.append(f"- Run: {run_url}")
+    lines += ["", "| workload::cell | reasons |", "| --- | --- |"]
+    for e in fails:
+        reasons = "; ".join(e.get("reasons", []) or []) or (e.get("error") or "")
+        lines.append(f"| `{e.get('entry')}::{e.get('cell')}` | {reasons} |")
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in 1f74cc2: added `_md_cell()` which escapes `|` and flattens newlines; applied to both the workload::cell and reasons columns in `render_issue`. Test added.

scripts/ci/nightly_eval.py
```diff
@@ -0,0 +1,172 @@
...
+    argv: list[str] = []
+    if nproc:
+        argv += ["torchrun", "--standalone", f"--nproc_per_node={int(nproc)}"]
+    argv += [aorta, "sweep", "run", "--recipe", recipe, "--output-dir", str(run_out), "--strict"]
+
+    print(f"RUN {entry['name']}: {' '.join(argv)}", flush=True)
+    proc = subprocess.run(argv, cwd=REPO_ROOT)
+
+    matrices = sorted(run_out.rglob("matrix.json"), key=lambda p: p.stat().st_mtime)
+    return proc.returncode, (matrices[-1] if matrices else None)
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in 1f74cc2: `run_entry` now `shutil.rmtree`s the per-entry output dir before running, so a stale `matrix.json` from a prior run can't be picked up by `rglob` when a run fails before producing a fresh one.

scripts/ci/nightly_eval.py
```diff
@@ -0,0 +1,176 @@
...
+        raise SystemExit(f"{path} must be a YAML mapping, got {type(data).__name__}")
+    return data
+
+
+def gpu_count() -> int:
+    try:
+        import torch
+
+        return torch.cuda.device_count() if torch.cuda.is_available() else 0
+    except Exception:
```
amd-vivekag: **[Medium]** `gpu_count()` catches *all* exceptions and returns 0, so on the GPU runner a present-but-broken torch (bad libs, CUDA init error) makes every entry `skip` → `fail=0` → exit 0. The nightly goes green while actually running nothing, and alerting will close any open regression issue.

Fix: only treat a genuine `ImportError` as 0 GPUs; let other failures surface (or fail fast on a GPU runner). e.g.
```python
def gpu_count() -> int:
    try:
        import torch
    except ImportError:
        return 0
    return torch.cuda.device_count() if torch.cuda.is_available() else 0
```

amd-vivekag (commented):
## Review summary

**Verdict: ship-with-nits** (one Medium worth a quick fix; the rest are Low).

Solid, well-structured, well-tested addition. I verified the harness's `matrix.json` assumptions against aorta's real schema (`write_matrix_json` / `CellStats`) and the `aorta sweep run --recipe/--output-dir/--strict` flags against the CLI — all match. All 10 matrix recipes exist, the `Nightly wheels` workflow and `rocm-ci-setup` action referenced exist, and the 23 `tests/ci/` tests pass locally. The record-only-until-blessed design keeps nightly from going red before baselines land, and the workflows are careful (digest-pinned busybox, least-ish privilege, no untrusted PR code executed).

**Re-review note — all 4 prior Copilot comments are fixed correctly:**
- refresh_baselines zero-value → now `is not None` (verified).
- gen_dashboard XSS → `html.escape` applied to all untrusted fields + test (verified).
- alert_issue markdown table → `_md_cell()` escapes `|`/newlines + test (verified).
- nightly_eval stale `matrix.json` → `shutil.rmtree` before run (verified).

**Findings:** 0 High · 1 Medium · 4 Low

- **[Medium]** `gpu_count()` swallows *all* exceptions and returns 0 → a present-but-broken torch/CUDA on the GPU runner yields an all-`skip`, exit-0 (green) nightly that silently stops catching regressions.

Note: `ruff` is configured (`select` includes `F`) but not run in pre-commit or CI, so the F541 nit below won't fail the gate — it'll only surface on a local `ruff check`.

scripts/ci/alert_issue.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def _req(method: str, url: str, token: str, payload: dict | None = None) -> Any:
+    data = json.dumps(payload).encode() if payload is not None else None
+    req = urllib.request.Request(url, data=data, method=method)
+    req.add_header("Authorization", f"Bearer {token}")
+    req.add_header("Accept", "application/vnd.github+json")
+    if data:
+        req.add_header("Content-Type", "application/json")
+    with urllib.request.urlopen(req) as resp:
```
amd-vivekag: **[Low]** `_req()` has no network/HTTP error handling, and the "Alert on regression" step runs under `if: always()`. A transient GitHub API/network blip here fails the whole nightly job even when the eval itself passed (and no issue gets filed).

Fix: wrap the API calls in `main()` and treat alerting as best-effort (log a warning, `return 0`) on `URLError`/`HTTPError`. (Copilot flagged this too.)

scripts/ci/alert_issue.py
```diff
@@ -0,0 +1,119 @@
...
+        body = resp.read().decode()
+        return json.loads(body) if body else {}
+
+
+def _find_open_issue(repo: str, token: str) -> dict | None:
+    url = f"{API}/repos/{repo}/issues?state=open&labels={LABEL}&per_page=1"
```
amd-vivekag: **[Low]** `/repos/{repo}/issues` also returns pull requests, so if a PR ever carries the `nightly-regression` label, `_find_open_issue()` could return it and the script would PATCH/close/comment on a PR. Mitigated by the dedicated label, but cheap to harden.

Fix: skip PRs, e.g. `items = [i for i in _req(...) if "pull_request" not in i]`. (Copilot flagged this too.)

scripts/ci/eval_lib.py
```diff
@@ -0,0 +1,137 @@
...
+
+    reasons: list[str] = []
+    deltas: dict[str, Any] = {}
+    metrics = harvested.get("metrics", {})
+
+    # Correctness: require the cell to have passed if the baseline expects it.
+    if baseline.get("passed", True) and not harvested.get("passed", False):
+        err = harvested.get("error")
+        reasons.append(
+            f"expected passing cell but it did not pass"
```
amd-vivekag: **[Low]** f-string with no placeholders (`ruff` F541).

Fix:
```suggestion
            "expected passing cell but it did not pass"
```

scripts/ci/gen_dashboard.py
```diff
@@ -0,0 +1,183 @@
...
+    "fail": "#d1242f",
+    "record": "#9a6700",
+    "skip": "#57606a",
+}
+
+
+def load_results(results_dir: Path) -> list[dict[str, Any]]:
+    """Load and sort all result docs by generated_at (oldest first)."""
+    docs: list[dict[str, Any]] = []
+    for p in sorted(results_dir.glob("*.json")):
```
amd-vivekag, line 58: **[Low]** `results/*.json` on `gh-pages` grows unbounded (one per day, kept forever) and the dashboard re-reads the entire history on every run. Not a near-term problem (files are tiny), but there's no retention/pruning.

Fix: cap the history (e.g. keep last N days, or prune in the publish step), or note the retention decision in `docs/ci-nightly-eval.md`.

scripts/ci/nightly_eval.py
vivekkhandelwal1: Fixed in a35f959: `gpu_count()` now only treats `ImportError` as 0 GPUs; a present-but-broken torch (HIP/CUDA init error) propagates and fails the nightly loudly instead of skipping everything to a false green.

scripts/ci/alert_issue.py
vivekkhandelwal1: Fixed in a35f959: alerting is now best-effort — `main()` wraps the API calls and catches `URLError`/`HTTPError`, logging a `::warning::` and returning 0, so a transient GitHub API blip under `if: always()` can't fail a passing nightly.

scripts/ci/alert_issue.py
vivekkhandelwal1: Fixed in a35f959: `_find_open_issue()` filters out any item carrying `pull_request`, so a PR labeled `nightly-regression` can never be PATCHed/closed/commented on.

scripts/ci/eval_lib.py
vivekkhandelwal1: Fixed in a35f959: dropped the stray `f` prefix (F541).

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in a35f959: the publish step now prunes `results/*.json` to the most recent 180 on gh-pages, and the dashboard caps rendering with `--max-builds 180`. Retention is documented in docs/ci-nightly-eval.md.

scripts/ci/refresh_baselines.py
```diff
@@ -0,0 +1,140 @@
...
+    ngpu = nightly_eval.gpu_count()
+    baselines: dict[str, Any] = {}
+    incomplete: list[str] = []
+
+    for entry in matrix_doc.get("entries") or []:
+        name = entry["name"]
+        min_gpus = int(entry.get("min_gpus", entry.get("nproc", 1)))
+        if ngpu < min_gpus:
+            incomplete.append(f"{name} (skipped: needs {min_gpus} GPU(s), have {ngpu})")
+            continue
```
amd-vivekag, line 62: **[Medium]** `refresh_baselines` can never succeed on a runner with fewer GPUs than the largest `min_gpus` in the matrix.

A GPU-skipped entry is appended to `incomplete`, and `if incomplete:` (line 101) refuses the whole refresh before writing anything. The matrix ships 8-GPU entries (`min_gpus: 8`); on any runner with <8 GPUs those always skip, so a full baseline can never be produced — perf gating (Phase 5) and even the default correctness-metric baselines become unblessing-able. This contradicts the matrix design, which explicitly expects those entries to skip when <8 GPUs are visible.

Fix: treat insufficient-GPU skips as legitimate (don't add them to `incomplete`), and preserve any existing baseline entry for a skipped key rather than dropping it on the full-file rewrite. Refuse only on genuine failures (timeout / no-matrix / cell-didn't-pass).

scripts/ci/gen_dashboard.py
```diff
@@ -0,0 +1,254 @@
...
+            f"<td>{_esc(val_txt)}</td>"
+            f"<td>{_svg_sparkline(mhist[(k, m)])}</td></tr>"
+        )
+
+    s = latest.get("summary", {})
+    meta = _esc(
+        f"aorta {build.get('amd_aorta_version', '?')} · "
+        f"torch {build.get('torch', '?')} · ROCm {build.get('rocm', '?')} · "
+        f"HIP {build.get('hip', '?')}"
+    )
```
amd-vivekag: **[Low]** `build_metadata()` sets these keys to `None` when undetected, so `build.get('rocm', '?')` returns `None` (the key exists) and the dashboard renders the literal `ROCm None`, not `ROCm ?`. Same `.get(k, '?')`-with-None pattern in `alert_issue.render_issue`. (Copilot flagged this.)

Fix: use `or '?'`, e.g. `build.get('rocm') or '?'`.

scripts/ci/gen_dashboard.py
```diff
@@ -0,0 +1,254 @@
...
+  </table>
+  <p class="muted">Verdicts: pass/fail = vs blessed baseline · record = no baseline yet (metrics captured) · skip = insufficient GPUs.</p>
+
+  <h2 style="font-size:1.1rem;margin-top:2rem;">Performance / metric trends</h2>
+  <table>
+    <thead><tr><th>workload::cell</th><th>metric</th><th>policy</th>
+      <th>latest</th><th>trend</th></tr></thead>
+    <tbody>
+      {{metric_table}}
+    </tbody>
```
amd-vivekag, line 229: **[Low]** The page is built as `f"...{{metric_table}}...".replace("{metric_table}", ...)` (see the `.replace` at line 229). `html.escape` doesn't escape braces, so any escaped reason/key/metric that contains the literal text `{metric_table}` would get the whole metric-table HTML spliced into that cell. Cosmetic, but avoidable.

Fix: compute `metric_html` first and interpolate it directly in the f-string instead of the placeholder + `.replace`.

scripts/ci/nightly_eval.py
```diff
@@ -0,0 +1,271 @@
...
+            results.append(
+                {"entry": name, "recipe": entry["recipe"], "cell": None,
+                 "verdict": "fail", "reasons": [f"no matrix.json produced (exit {rc})"],
+                 "metrics": {}, "deltas": {}, "duration_sec": dur,
+                 "error": f"exit {rc}", "matrix_path": None}
+            )
+            continue
+
+        harvested_cells = eval_lib.harvest_matrix_json(matrix_path)
+        if not harvested_cells:
```
amd-vivekag, line 211: **[Low]** `harvest_matrix_json` can raise `JSONDecodeError` on a truncated/corrupt `matrix.json` (e.g. a crash mid-write). That aborts `evaluate()` before the results JSON is written, so the job goes red but the alert + dashboard silently no-op (no results artifact) — exactly the regression they exist to surface. (Copilot flagged this.)

Fix: wrap the harvest in try/except and record a per-entry `fail` with the decode error so a results JSON is still produced for alerting/dashboard.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship** — one Medium in `refresh_baselines.py` worth landing before anyone tries to bless baselines; the rest are Low. The nightly harness itself is solid and safe to merge for the record-only path.

Re-review of the current diff. This covers both the previously-flagged code and the substantial new code added since my last pass (reusable eval workflow, bump-validate, per-metric policies, per-entry timeout + process-group kill, partial-refresh refusal, ci-constraints, Pages via `deploy-pages`).

**Prior review — all resolved (verified in code, not just replies):**
- `gpu_count()` (my Medium) → now only swallows `ImportError`; a present-but-broken torch surfaces. ✅
- `alert_issue` best-effort → `main()` catches `URLError`/`HTTPError` + `timeout=30`. ✅
- `_find_open_issue` PR filtering ✅ · F541 ✅ · history retention (180 + `--max-builds`) ✅
- Copilot's 4 original inline comments (refresh zero-value, dashboard XSS, alert md-table, stale matrix.json) all fixed + tested. ✅

**Findings:** 0 High · 1 Medium · 3 Low

**Medium (should land before baseline blessing):**
- `refresh_baselines.py` refuses the whole refresh when any entry is GPU-skipped. With `min_gpus: 8` entries in the matrix, baselines can never be blessed on a runner with <8 GPUs.

Nice touches: fail-closed semantics (zero-work / empty-matrix / timeout all fail), timeout kills the whole process group, digest-pinned busybox, no fork code on the self-hosted runner, and 36 CPU-gate-safe tests. Dismissed a few Copilot low-confidence notes (missing UA — urllib sends a default; `head -n -180` — publish runs on ubuntu-latest with GNU head).

scripts/ci/refresh_baselines.py
vivekkhandelwal1: Fixed in b48346b: insufficient-GPU skips are no longer added to `incomplete`. Such an entry now carries over its existing baselines unchanged (never reverted to record-only); an entry with no prior baseline is warned, not fatal. Only genuine failures (timeout / no-matrix / empty-cells / did-not-pass) still refuse atomically. Tests added for carry-over, non-fatal-without-prior, and refuse-on-ran-but-failed.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in b48346b: build metadata now renders via `build.get(k) or '?'` so undetected fields show `?` instead of the literal `None`. Same fix applied in `alert_issue.render_issue`.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in b48346b: `metric_table` is computed first and interpolated directly in the f-string; the `{metric_table}` placeholder + `.replace` is gone, so an escaped field containing that literal text can no longer splice in the table HTML.

scripts/ci/nightly_eval.py
vivekkhandelwal1: Fixed in b48346b: `harvest_matrix_json` is wrapped in try/except (JSONDecodeError is a ValueError; also OSError). A corrupt/truncated matrix.json now records a per-entry `fail` (`error=corrupt_matrix`) so the results JSON is still written and alerting/dashboard surface the regression. Test added.

scripts/ci/refresh_baselines.py
```diff
@@ -0,0 +1,175 @@
...
+        if timed_out:
+            incomplete.append(f"{name} (timed out)")
+            continue
+        if matrix_path is None:
+            incomplete.append(f"{name} (no matrix.json; exit {rc})")
+            continue
+
+        harvested_cells = eval_lib.harvest_matrix_json(matrix_path)
+        if not harvested_cells:
+            incomplete.append(f"{name} (matrix.json had no cells)")
```
amd-vivekag: **[Low]** Consistency: `harvest_matrix_json` here is unguarded, unlike the `nightly_eval` path you just wrapped. A corrupt/truncated `matrix.json` aborts the refresh with a raw traceback instead of the clean atomic "refusing partial refresh" message. It's still fail-closed (no bad baseline PR), so this is minor. (Copilot flagged this too.)

Fix: wrap it and route decode/IO errors into `incomplete`, e.g.
```python
try:
    harvested_cells = eval_lib.harvest_matrix_json(matrix_path)
except (ValueError, OSError) as exc:
    incomplete.append(f"{name} (unreadable matrix.json: {exc})")
    continue
```

amd-vivekag (commented):
## Re-review summary

**Verdict: ship** — all four of my previous findings are fixed correctly; only one Low consistency nit remains.

Verified each fix in `b48346b` against the code (not just the replies):

- **[Medium] `refresh_baselines` unblessing-able on <8-GPU runners** → Fixed well. GPU-skipped entries now carry over their existing baselines (matched by `entry` prefix) instead of being dropped or refusing the whole refresh; entries with no prior baseline warn (non-fatal); `incomplete` (atomic refuse) is now only ran-but-failed cases; `main()` loads the existing baselines and passes them in. Carry-over / non-fatal / refuse-on-failed tests added. ✅
- **[Low] `None` rendered as literal `None`** → `build.get(k) or '?'` in both `gen_dashboard` and `alert_issue`. ✅
- **[Low] `{metric_table}` splice** → `metric_table` computed first and interpolated directly; placeholder + `.replace` gone. ✅
- **[Low] corrupt `matrix.json` in `nightly_eval`** → wrapped in `try/except (ValueError, OSError)`, records a per-entry `corrupt_matrix` fail so a results JSON is still written for alerting/dashboard. ✅

**Findings this round:** 0 High · 0 Medium · 1 Low

One small consistency gap remains (also a Copilot suppressed note): `refresh_baselines.build_baselines` still calls `harvest_matrix_json` unguarded, unlike the `nightly_eval` path you just hardened. It's fail-closed (a traceback still aborts the refresh, so no bad baseline PR is opened), hence Low — worth mirroring the guard for a clean message. I dismissed Copilot's other note (`grep -E '\s'`): GNU grep supports `\s` as an extension and this only runs in the GNU-grep ROCm container, so the pin count is correct there.

Not approving only because the skill treats a newly-surfaced item as not-clean; functionally this is ship-ready.

scripts/ci/refresh_baselines.py
vivekkhandelwal1: Fixed in 680532f: `harvest_matrix_json` in `refresh_baselines.build_baselines` is now wrapped in try/except `(ValueError, OSError)` and routes decode/IO errors into `incomplete`, so a corrupt/truncated matrix.json fails with the clean atomic "refusing partial refresh" message instead of a raw traceback — consistent with the `nightly_eval` guard. Test added.

amd-vivekag (approved):
## Approving — clean re-review

**Verdict: ship.** All findings from my prior reviews are now fixed correctly (verified in code, with tests):

- **[Medium]** `refresh_baselines` unblessing-able on <8-GPU runners → carry-over of existing baselines for GPU-skipped entries; atomic refuse only on ran-but-failed. ✅
- **[Low]** `None` rendered literally → `build.get(k) or '?'`. ✅
- **[Low]** `{metric_table}` splice → interpolated directly. ✅
- **[Low]** corrupt `matrix.json` in `nightly_eval` → `try/except` records a `corrupt_matrix` fail. ✅
- **[Low]** corrupt `matrix.json` in `refresh_baselines` (last round) → now guarded, routed into `incomplete` for the clean atomic refuse (680532f). ✅

Gap sweep on the current diff (incl. the merge from `main`) found no new code issues.

Non-blocking FYI (no change requested): the PR **description** still says the dashboard publishes to `gh-pages`, but the implementation (and `docs/ci-nightly-eval.md`) correctly use the `ci-results` data branch + `actions/deploy-pages`. Worth updating the description so operators aren't confused. I also dismissed Copilot's `grep -E '\s'` note — GNU grep supports `\s` and this only runs in the GNU-grep ROCm container.

PR #322, opened 2026-08-04
ci(dashboard): serve the dashboard at the site root and rebuild its layout

vivekkhandelwal1:
## Why

Two problems, both about the nightly results being unusable in practice.

**Nobody could find the dashboard.** It was served at `/ci/`, and the landing page at the site root never linked to it — the word "dashboard" did not appear anywhere in the README. Anyone visiting https://rocm.github.io/aorta/ saw an unchanged project README and reasonably concluded nothing had shipped.

**The page itself was unreadable for a newcomer.** It was two long tables. On the live data, three of the five columns in the first table carried no information at all: `status` said `record` 16 times, `trend` said `n/a` 16 times, and `notes` repeated "no baseline (record-only)" 16 times — and those were the widest columns. Step times spanning 39 ms to 19,125 ms sat in a plain column with no sense of scale, and `llm_determinism`'s eight cells were flat rows with no grouping, so the sweep matrix was invisible.

### Site layout

The dashboard now owns the site root; the Jekyll-rendered README moves to `/docs/`, which had no index of its own (it 404s today, so nothing is overwritten). The README gains a link to the dashboard.

The rendered README mixes root-absolute links with relative ones, and moving it one directory deeper breaks only the relative ones, so those are re-pointed with a `../` prefix. On the real page this rewrites exactly one link, `recipes/` → `../recipes/`, which resolves to a live 200; fragment, root-absolute and absolute links are untouched and the result is still well-formed.

The no-history path also changed: rather than deploying docs only, the root publishes the dashboard's own empty state, so the site root can never 404.

### The dashboard

- **Grouped by workload**, with a per-group verdict tally, so the matrix structure is visible.
- **Uninformative columns are dropped** instead of rendered as a wall of `n/a`; the record-only page went from 117 `n/a` strings to 3. A note shared by *every* cell collapses into a single banner — but any variation keeps the column, because a failing cell's reason must never be dropped.
- **The metric-trends table is folded into collapsible per-cell rows**, which also carry the recipe, duration and trial count.
- **Step time gets a bar scaled within its workload group**, where the ratio is meaningful; single-cell groups get no bar, since that would only compare a value to itself.
- **Rebuilt header**: the toolchain identity is a labelled grid rather than one cramped dot-separated line, with links to the commit, the workflow run and the docs, plus a plain-English description of what the page shows.
- **Honest status**: a run with no blessed baselines reports `recording`, not `passing`, and a banner explains why there are no verdicts.

## Verification

Rendered four history shapes — the real published history, plus synthetic empty, multi-build-with-a-failure, and mixed-verdict-with-skips:

- All four are well-formed HTML (zero unclosed tags, zero nesting errors).
- The multi-build render draws sparklines and surfaces the failing cell's reason.
- The link rewrite was checked against the live landing page: 1 of 25 links changed, and it resolves to a 200.

Test suite goes from 11 to 16 tests. New coverage for: a failure's reason always rendering, a shared note collapsing to a banner, record-only not reporting as passing, workload grouping, and the trend column appearing only once there is history. Two existing tests were updated for the restructured markup (`workload::cell` is now a group row plus a cell row, and the "Performance / metric trends" heading is gone).

## Note for the reviewer

While building this I introduced and then fixed a bug worth flagging, since the column-collapsing logic is the subtle part: the first version dropped the notes column whenever there was only one *distinct* reason string, which meant a graded failure rendered a bare `fail` badge with its explanation nowhere on the page. `test_dashboard_always_explains_a_failure` now guards it.

### DIFF
--- a/scripts/ci/gen_dashboard.py
+++ b/scripts/ci/gen_dashboard.py
@@ -82,6 +82,17 @@ def _isnum(v: Any) -> bool:
         return False
 
 
+def _count(v: Any) -> int:
+    """A summary count as an int; anything unusable counts as 0.
+
+    Counts drive branching and division, so a string in a malformed summary
+    would crash generation outright ("4" + 0 raises) instead of rendering badly.
+    Display still goes through ``_fmt_num``, which shows "—" rather than a 0 it
+    cannot vouch for.
+    """
+    return int(v) if _isnum(v) else 0
+
+
 def load_results(results_dir: Path) -> list[dict[str, Any]]:
     """Load and sort all result docs by generated_at (oldest first)."""
     docs: list[dict[str, Any]] = []
@@ -145,13 +156,13 @@ def _latest_status(results: list[dict[str, Any]]) -> tuple[str, str]:
     if not results:
         return "unknown", "#57606a"
     s = results[-1].get("summary", {}) or {}
-    if s.get("fail", 0):
+    if _count(s.get("fail")):
         return "failing", _VERDICT_COLOR["fail"]
-    if (s.get("total", 0) or 0) == 0:
+    if _count(s.get("total")) == 0:
         return "empty", "#57606a"
-    if s.get("pass", 0) or 0:
+    if _count(s.get("pass")):
         return "passing", _VERDICT_COLOR["pass"]
-    if s.get("record", 0) or 0:
+    if _count(s.get("record")):
         return "recording", _VERDICT_COLOR["record"]
     return "skipping", _VERDICT_COLOR["skip"]
 
@@ -276,13 +287,14 @@ def build_dashboard_html(results: list[dict[str, Any]]) -> str:
     passrate = []
     for doc in results:
         ds = doc.get("summary", {}) or {}
-        graded = (ds.get("pass", 0) + ds.get("fail", 0)) or 0
-        passrate.append((ds.get("pass", 0) / graded) if graded else None)
+        passed, failed = _count(ds.get("pass")), _count(ds.get("fail"))
+        graded = passed + failed
+        passrate.append((passed / graded) if graded else None)
 
-    total = s.get("total", 0) or 0
-    graded_now = (s.get("pass", 0) or 0) + (s.get("fail", 0) or 0)
-    record_now = s.get("record", 0) or 0
-    skip_now = s.get("skip", 0) or 0
+    total = _count(s.get("total"))
+    graded_now = _count(s.get("pass")) + _count(s.get("fail"))
+    record_now = _count(s.get("record"))
+    skip_now = _count(s.get("skip"))
     # "Nothing was graded" splits three ways and they must not be described
     # identically: every cell recorded a baseline, some recorded while others
     # were skipped, or nothing ran at all. The last of those is unreachable for
@@ -474,7 +486,7 @@ def build_dashboard_html(results: list[dict[str, Any]]) -> str:
     # Counts go through _fmt_num so a malformed summary shows "—" rather than
     # printing whatever str() makes of it ("nan", "True").
     cards = [
-        ("results", _fmt_num(total), ""),
+        ("results", _fmt_num(s.get("total", 0)), ""),
         ("pass", _fmt_num(s.get("pass", 0)), "#3fb950"),
         ("fail", _fmt_num(s.get("fail", 0)), "#f85149"),
         ("record", _fmt_num(s.get("record", 0)), "#d29922"),
--- a/scripts/ci/gen_dashboard.py
+++ b/scripts/ci/gen_dashboard.py
@@ -112,8 +112,9 @@ def _has_trend(series: list[list[float | None]]) -> bool:
 def _latest_status(results: list[dict[str, Any]]) -> tuple[str, str]:
     """Headline verdict for the newest build.
 
-    ``recording`` is distinct from ``passing``: with no blessed baselines
-    nothing was actually graded, and calling that "passing" overstates it.
+    Only a graded pass earns ``passing``. A build with no blessed baselines is
+    ``recording``, and one where every cell was skipped is ``skipping``:
+    reporting either as "passing" would claim a result nothing established.
     """
     if not results:
         return "unknown", "#57606a"
@@ -122,9 +123,11 @@ def _latest_status(results: list[dict[str, Any]]) -> tuple[str, str]:
         return "failing", _VERDICT_COLOR["fail"]
     if (s.get("total", 0) or 0) == 0:
         return "empty", "#57606a"
-    if not (s.get("pass", 0) or 0) and (s.get("record", 0) or 0):
+    if s.get("pass", 0) or 0:
+        return "passing", _VERDICT_COLOR["pass"]
+    if s.get("record", 0) or 0:
         return "recording", _VERDICT_COLOR["record"]
-    return "passing", _VERDICT_COLOR["pass"]
+    return "skipping", _VERDICT_COLOR["skip"]
 
 
 def _fmt_num(v: Any) -> str:
@@ -250,7 +253,12 @@ def build_dashboard_html(results: list[dict[str, Any]]) -> str:
 
     total = s.get("total", 0) or 0
     graded_now = (s.get("pass", 0) or 0) + (s.get("fail", 0) or 0)
-    record_only = bool(total) and not graded_now
+    record_now = s.get("record", 0) or 0
+    skip_now = s.get("skip", 0) or 0
+    # "Nothing was graded" splits three ways and they must not be described
+    # identically: every cell recorded a baseline, some recorded while others
+    # were skipped, or nothing ran at all.
+    nothing_graded = bool(total) and not graded_now
 
     # Columns that would be uniformly empty are dropped rather than rendered as
     # a wall of "n/a" -- with one night of history that was three of five.
@@ -385,12 +393,22 @@ def build_dashboard_html(results: list[dict[str, Any]]) -> str:
             "<div class='notice'>No nightly results have been published yet. "
             "This page fills in after the first Nightly Evaluation run.</div>"
         )
-    elif record_only:
+    elif nothing_graded and record_now:
+        scope = (
+            f"all {total} cells" if record_now == total
+            else f"{record_now} of {total} cells ({skip_now} skipped)"
+        )
         extra = f" Every cell reports: {_esc(shared_reason)}." if shared_reason else ""
         notices.append(
             f"<div class='notice'>No baselines are blessed yet, so nothing was graded "
-            f"pass or fail — this run <strong>recorded</strong> metrics for all {total} "
-            f"cells to become the reference.{extra}</div>"
+            f"pass or fail — this run <strong>recorded</strong> metrics for {scope} "
+            f"to become the reference.{extra}</div>"
+        )
+    elif nothing_graded:
+        notices.append(
+            f"<div class='notice'>Nothing ran: all {total} cells were "
+            f"<strong>skipped</strong>, so this build establishes nothing. Check that "
+            f"the runner exposes as many GPUs as the matrix asks for.</div>"
         )
     elif shared_reason:
         notices.append(
--- a/scripts/ci/gen_dashboard.py
+++ b/scripts/ci/gen_dashboard.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import math
 import sys
 from html import escape as _esc
 from pathlib import Path
@@ -64,13 +65,21 @@ _VERDICT_ORDER = ("fail", "record", "skip", "pass")
 
 
 def _isnum(v: Any) -> bool:
-    """Numeric for display purposes.
+    """Numeric, finite, and representable as a float — i.e. safe to render.
 
-    ``bool`` is a subclass of ``int``, so a stray ``true`` in the (untrusted)
-    results JSON would otherwise format as a step time, scale a bar, or count
-    towards a trend.
+    Three things in the (untrusted) results JSON pass a bare isinstance check
+    and should not: ``bool``, which is a subclass of ``int`` and would format as
+    a step time or count towards a trend; ``NaN``/``Infinity``, which
+    ``json.loads`` accepts and which plot as ``nan`` coordinates and a bar that
+    silently reads full-width; and an integer too large to convert to a float,
+    which raises ``OverflowError`` in every formatter below.
     """
-    return isinstance(v, (int, float)) and not isinstance(v, bool)
+    if not isinstance(v, (int, float)) or isinstance(v, bool):
+        return False
+    try:
+        return math.isfinite(v)
+    except OverflowError:  # int beyond the float range
+        return False
 
 
 def load_results(results_dir: Path) -> list[dict[str, Any]]:
@@ -85,8 +94,10 @@ def load_results(results_dir: Path) -> list[dict[str, Any]]:
     return docs
 
 
-def _svg_sparkline(values: list[float], width: int = 160, height: int = 32) -> str:
-    """Tiny inline SVG line chart for a metric's history (ignores None)."""
+def _svg_sparkline(
+    values: list[float | None], width: int = 160, height: int = 32
+) -> str:
+    """Tiny inline SVG line chart for a metric's history (ignores non-numbers)."""
     pts = [v for v in values if _isnum(v)]
     if len(pts) < 2:
         return '<span class="muted">n/a</span>'
@@ -444,16 +455,18 @@ def build_dashboard_html(results: list[dict[str, Any]]) -> str:
             "runs; they appear automatically once more history accumulates.</div>"
         )
 
+    # Counts go through _fmt_num so a malformed summary shows "—" rather than
+    # printing whatever str() makes of it ("nan", "True").
     cards = [
-        ("cells", total, ""),
-        ("pass", s.get("pass", 0), "#3fb950"),
-        ("fail", s.get("fail", 0), "#f85149"),
-        ("record", s.get("record", 0), "#d29922"),
-        ("skip", s.get("skip", 0), "#8b949e"),
+        ("cells", _fmt_num(total), ""),
+        ("pass", _fmt_num(s.get("pass", 0)), "#3fb950"),
+        ("fail", _fmt_num(s.get("fail", 0)), "#f85149"),
+        ("record", _fmt_num(s.get("record", 0)), "#d29922"),
+        ("skip", _fmt_num(s.get("skip", 0)), "#8b949e"),
     ]
     # Only worth a card when something was actually graded; otherwise it would
     # sit next to the trend card reading "n/a" twice.
-    if passrate and isinstance(passrate[-1], float):
+    if passrate and _isnum(passrate[-1]):
         cards.append(("pass rate", f"{passrate[-1] * 100:.0f}%", ""))
     cards_html = "".join(
         f'<div class="card"><div class="k">{_esc(k)}</div>'
--- a/scripts/ci/gen_dashboard.py
+++ b/scripts/ci/gen_dashboard.py
@@ -332,12 +332,23 @@ def build_dashboard_html(results: list[dict[str, Any]]) -> str:
         tally_txt = " · ".join(
             f"{tally[v]} {v}" for v in _VERDICT_ORDER if tally.get(v)
         )
-        cell_word = "cell" if len(cell_keys) == 1 else "cells"
+        n = len(cell_keys)
+        head = f"{n} result{'' if n == 1 else 's'} · {tally_txt}"
+        # duration_sec is measured once around the whole recipe invocation and
+        # copied onto every record it produced, so it belongs to the workload and
+        # is stated here once -- repeated in each cell it reads as per-cell time.
+        durs = {
+            d for d in (latest_by_key[k].get("duration_sec") for k in cell_keys)
+            if _isnum(d)
+        }
+        # A skipped workload reports 0.0 -- it never ran, so it has no duration.
+        if len(durs) == 1 and max(durs) > 0:
+            head += f" · workload run {max(durs):,.0f}s"
 
         rows = [
             f"<tr class='grp'><th colspan='{ncols}' scope='rowgroup'>"
             f"<span class='wl'>{_esc(entry_name)}</span>"
-            f"<span class='muted'> {len(cell_keys)} {cell_word} · {_esc(tally_txt)}</span>"
+            f"<span class='muted'> {_esc(head)}</span>"
             f"</th></tr>"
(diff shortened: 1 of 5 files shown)


scripts/ci/gen_dashboard.py
```diff
@@ -91,24 +102,114 @@ def _svg_sparkline(values: list[float], width: int = 160, height: int = 32) -> s
...
-    s = results[-1].get("summary", {})
+    s = results[-1].get("summary", {}) or {}
     if s.get("fail", 0):
         return "failing", _VERDICT_COLOR["fail"]
-    if s.get("total", 0) == 0:
+    if (s.get("total", 0) or 0) == 0:
         return "empty", "#57606a"
+    if not (s.get("pass", 0) or 0) and (s.get("record", 0) or 0):
+        return "recording", _VERDICT_COLOR["record"]
     return "passing", _VERDICT_COLOR["pass"]
```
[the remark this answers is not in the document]
vivekkhandelwal1: Good catch, and it applied in two places. `_latest_status` now only returns `passing` for a graded pass: a build with records but no grades is `recording`, and one where every cell was skipped is `skipping`, using the existing skip colour. The same conflation reached the banner — `record_only` was "total > 0 and nothing graded", which is equally true of a skip-only or record+skip build, so the page could claim it had "recorded metrics for all N cells" when some or all never ran. Those three cases are now worded apart. Covered by `test_dashboard_skip_only_run_is_neither_passing_nor_recording` and `test_dashboard_partial_record_does_not_claim_every_cell_recorded`.

scripts/ci/gen_dashboard.py
```diff
@@ -126,152 +230,333 @@ def build_dashboard_html(results: list[dict[str, Any]]) -> str:
...
+                    f"<td class='muted'>{_esc('; '.join(e.get('reasons') or []))}</td>"
+                )
+            rows.append(f"<tr>{''.join(cells)}</tr>")
 
-    s = latest.get("summary", {})
-    meta = _esc(
-        f"aorta {build.get('amd_aorta_version') or '?'} · "
-        f"torch {build.get('torch') or '?'} · ROCm {build.get('rocm') or '?'} · "
-        f"HIP {build.get('hip') or '?'}"
+            mrows = _metric_rows(k, e, mhist, show_trend)
```
amd-vivekag: **[Medium]** Per-cell metric trends are gated on step-time history, so a metric-only workload loses real trend signal and gets a misleading notice.

`show_trend` is derived only from `mean_step_time_ms` history, but `eval_lib.harvest` sets `mean_step_time_ms` to `cell.get(...)`, which can be `None` while `metrics.summary` has values. I reproduced it: two runs of a cell with only `decode_latency_ms` draw no metric sparkline and still render "Trend charts need at least two nightly runs". Not triggered by today's workloads (they emit step time), so it's latent — Copilot's second-review suppressed comments flagged the same root.

Fix: compute a separate `show_metric_trend` from `mhist` and use it for the `_metric_rows(...)` call here and the inner `<th>trend</th>` header (line 347), and gate the notice at line 417 on step-time OR per-metric history.

amd-vivekag (commented):
## Review summary

**Verdict: ship** — one non-blocking Medium follow-up.

Re-review after the skip-only fix. The redesign is well thought through: columns collapse only when truly redundant (with an explicit test that a failure's reason is never dropped), the status/banner wording now distinguishes graded-pass / recording / skip-only honestly, group colspans line up so the markup stays well-formed, the Pages workflow fails closed on a transport error and the site root can never 404. I re-rendered empty / record-only / multi-build-fail / skip-only histories (all well-formed HTML), confirmed the `sed` link-rewrite only touches double-quoted relative links (root-absolute, fragment, `https://`, `mailto:` left alone), verified the result-schema fields the header reads (`head_sha`, `recipe`, `duration_sec`, `trials`, …) are actually populated by `nightly_eval.py`, and ran the suite — 18 passed.

**Prior review comments:** the skip-only/`record_only` conflation Copilot flagged is **resolved correctly** (`_latest_status` only returns `passing` for a graded pass; the banner splits record-all / partial+skip / skip-only apart, each covered by a new test).

- **[Medium]** Per-cell metric sparklines and the "need two runs" notice are gated on step-time history only, so a metric-only workload (no `mean_step_time_ms`) loses real trend signal and shows a misleading notice. Latent for today's workloads (they emit step time). Copilot's 2nd-review suppressed comments flagged the same thing. Details inline.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed, and thanks for the repro — that was a real latent bug. One `show_trend` was gating three different things, so a cell reporting `metrics.summary` with `mean_step_time_ms` of `None` lost metric trends that genuinely existed while the page simultaneously claimed there was not enough history to chart anything.

There is now a separate `show_metric_trend` computed from `mhist`, used for the `_metric_rows(...)` call and the inner `<th>trend</th>` header. The step-time column keeps its own gate, since it has nothing to chart in that case, and the notice is now gated on `show_trend or show_metric_trend` so it only appears when neither kind of history can be drawn. This also covers both suppressed comments from Copilot second review (lines 350 and 421), which were the same root cause.

`test_dashboard_metric_trends_survive_a_missing_step_time` reproduces your scenario directly: two runs of a cell with only `decode_latency_ms` now draw the sparkline, the notice is gone, and the step-time column stays collapsed.

scripts/ci/gen_dashboard.py
```diff
@@ -91,24 +102,117 @@ def _svg_sparkline(values: list[float], width: int = 160, height: int = 32) -> s
...
+    """Compact number for display; keeps large integer counts readable."""
+    if not isinstance(v, (int, float)) or isinstance(v, bool):
+        return "—"
+    if float(v).is_integer() and abs(v) < 1e12:
+        return f"{int(v):,}"
+    return f"{v:.4g}"
+
+
+def _fmt_ms(v: Any) -> str:
+    return f"{v:,.1f} ms" if isinstance(v, (int, float)) else "—"
```
amd-vivekag: **[Low]** `bool` passes as a number here (and in `_bar` L181, `_has_trend` L109, the `step_times` max L296-300), so a stray `true` in the untrusted results JSON renders "1.0 ms", draws a bar, or fakes a trend. `_fmt_num` (L135) already excludes `bool` — these siblings should match.

Fix: add one shared guard and use it at all four sites, e.g.
```python
def _isnum(v: Any) -> bool:
    return isinstance(v, (int, float)) and not isinstance(v, bool)
```
then `_isnum(v)` in place of the bare `isinstance(v, (int, float))` checks.

amd-vivekag (commented):
## Re-review (commit fef1f62)

**Verdict: ship** — one optional Low nit; not blocking.

**Prior Medium — resolved correctly.** `show_metric_trend` is now computed from `mhist` and drives the metric rows, the inner `trend` header, and (with `show_trend`) the "needs two runs" notice; the step-time column keeps its own gate. I re-ran my repro: a two-run metric-only cell (`mean_step_time_ms=None`) now draws its sparkline, the notice is gone, and the step-time trend column stays collapsed. `test_dashboard_metric_trends_survive_a_missing_step_time` covers it, and the full suite is green (19 passed).

**One new gap (why I'm not auto-approving):** you asked me to approve only if I found nothing else, and I did find a Low item — so I'm surfacing it and leaving the call to you.

- **[Low]** `bool` is treated as a number in `_fmt_ms`, `_bar`, `_has_trend`, and the `step_times` max, while `_fmt_num` already guards against it. On the untrusted results JSON a stray `true` would render "1.0 ms", draw a bar, or fake a trend. Cosmetic only; details inline. Copilot flagged the same across those four spots.

Say the word if you consider the Low non-blocking and I'll approve; otherwise a one-line shared guard closes it.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed with the shared guard you suggested. There is now one `_isnum()` helper, and it replaces every bare numeric check in the file rather than just the four you cited, since they all read the same untrusted JSON:

- `_fmt_ms` (the anchor here) and `_fmt_num`, which now delegates its existing `bool` exclusion to the helper instead of open-coding it
- `_bar` and the per-workload `step_times` max, so a boolean can neither draw a bar nor set `group_max` for its neighbours
- `_has_trend` **and** `_svg_sparkline`, which had the same hole in two places. These two had to move together: had I guarded only `_has_trend`, a `[true, true]` series would have dropped the trend column while the sparkline still happily drew the line, so the "is there a trend" decision and what actually renders would have disagreed.
- while there, `duration_sec` and `trials` in the metric-details provenance line had it too (`isinstance(trials, int)` is `True` for `True`). `trials` is now `_isnum` + `int()`, which also stops a float count rendering as "1.0 trial".

The only remaining `isinstance` numeric check is the pass-rate one at L447, and that is deliberate: `passrate` holds the result of an internal true division, never a value off the wire, and `bool` is not a subclass of `float`.

Verified it changes nothing for real input: rendering the live `data.json` through the module before and after the change produces byte-identical HTML, and the diff only shows up on a bool-poisoned document (before: `1.0 ms` plus a bar; after: `—` and no bar). `test_dashboard_does_not_treat_a_boolean_as_a_measurement` pins both the helpers and the rendered page. Suite green — 20 passed in `tests/ci/test_dashboard_and_alert.py`, 51 across `tests/ci`.

This also covers the four suppressed comments in Copilot's third review (lines 109/143/181/300), which were the same root cause.

scripts/ci/gen_dashboard.py
```diff
@@ -91,24 +112,115 @@ def _svg_sparkline(values: list[float], width: int = 160, height: int = 32) -> s
...
         return "failing", _VERDICT_COLOR["fail"]
-    if s.get("total", 0) == 0:
+    if (s.get("total", 0) or 0) == 0:
         return "empty", "#57606a"
-    return "passing", _VERDICT_COLOR["pass"]
+    if s.get("pass", 0) or 0:
+        return "passing", _VERDICT_COLOR["pass"]
+    if s.get("record", 0) or 0:
+        return "recording", _VERDICT_COLOR["record"]
+    return "skipping", _VERDICT_COLOR["skip"]
```
amd-vivekag, line 167: **[Medium]** This `skipping` state (and the `elif nothing_graded:` "Nothing ran" banner) can't happen for real data. `nightly_eval.py` L244-250 appends a synthetic `_nightly_eval` `fail` entry whenever every cell is skipped or nothing ran, so a real zero-work build has `fail>=1` and `_latest_status` returns `failing` above. I verified: a realistic all-skip doc (skip cell + synthetic fail) renders `failing`, no `skipping` pill, no banner. So `test_dashboard_skip_only_run_is_neither_passing_nor_recording` asserts a shape the pipeline never produces — and the earlier `skip-only -> passing` bug it "fixed" was unreachable for the same reason.

Fix: either drop the unreachable `skipping`/"Nothing ran" branches, or if you keep them as defensive rendering, change that test to the real zero-work shape (include the `_nightly_eval` fail entry) so it asserts the actual production behaviour — `failing`, with the "no matrix entry ran" reason surfaced.

scripts/ci/gen_dashboard.py
```diff
@@ -91,24 +112,115 @@ def _svg_sparkline(values: list[float], width: int = 160, height: int = 32) -> s
...
+    cell_key: str,
+    entry: dict[str, Any],
+    mhist: dict[tuple[str, str], list[float | None]],
+    show_trend: bool,
+) -> str:
+    summary = ((entry.get("metrics") or {}).get("summary") or {})
+    out = []
+    for m in sorted(summary):
+        unit = _METRIC_UNITS.get(m, "")
+        val = f"{_fmt_num(summary.get(m))} {unit}".strip()
```
amd-vivekag: **[Low]** Unit is appended even when the value is missing, so a non-numeric mean renders `— ms`. `_fmt_num` returns `—`, then ` {unit}` is glued on. Only bites on malformed data (summary means are normally numeric), so minor.

Fix: only attach the unit when numeric, e.g.
```suggestion
        raw = summary.get(m)
        val = f"{_fmt_num(raw)} {unit}".strip() if _isnum(raw) else _fmt_num(raw)
```

amd-vivekag (commented):
## Re-review (commit 1fab5d0)

**Verdict: fix-before-ship** — I did not approve. My prior Low is fixed well, but the fresh Copilot pass surfaced a real one I'm confirming.

**My prior Low (bool-as-number): fixed correctly.** One `_isnum()` helper now replaces every bare numeric check (including `trials`, which used `isinstance(..., int)` and had the same bug), with a test covering the helpers and an end-to-end case. Nice — cleaner than the four-site patch I suggested.

**New — and it walks back some of my earlier praise.** I'd called the skip-only handling from the earlier commit "honest", but Copilot's latest suppressed comment is right and I verified it: `nightly_eval.py` (lines 244-250) appends a synthetic `verdict: "fail"` entry whenever every entry is skipped or nothing ran, so a real zero-work build carries `fail >= 1` and `_latest_status` returns **`failing`** before it can reach `skipping`. I rendered a realistic all-skip doc (skip cell + the synthetic `_nightly_eval` fail): headline was `failing`, and neither the `skipping` pill nor the "Nothing ran" banner appeared. So that state and `test_dashboard_skip_only_run_is_neither_passing_nor_recording` describe a document this repo never emits. Details inline.

**Findings:** 0 High · 1 Medium · 1 Low

- **[Medium]** `skipping` status + "Nothing ran" banner are unreachable for real data (synthetic `fail` wins); the test asserts an impossible shape.
- **[Low]** `_metric_rows` renders `— ms` (unit on a missing value); minor.

Neither is a production correctness bug — an all-skip nightly correctly shows `failing`. But the tests give false confidence, so I'm flagging before the green check.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Confirmed, and thanks — the false confidence was the real problem here. I reproduced your finding by building the doc the way `nightly_eval` does (skip cell, then the synthetic `_nightly_eval` entry from L244-250, summarised through `eval_lib.summarize`): summary comes out `{total: 2, pass: 0, fail: 1, record: 0, skip: 1}`, `_latest_status` returns `failing`, and neither the `skipping` pill nor the "Nothing ran" banner appears. I also checked the invariant's age — the zero-work guard landed in `89176cd`, the same commit that introduced the harness, so no published document predates it and there is no historical `data.json` that could reach the branch either.

I took your second option rather than deleting the branches, because deleting isn't quite free: `_latest_status` has to return *something* for a summary with `total > 0` and every cell skipped, and every remaining label overclaims — falling through to `passing` is exactly the bug Copilot opened this thread's ancestor about. So it isn't an unreachable statement so much as the terminal branch of an exhaustive classification, and the honest fix is to stop pretending it describes production:

- `test_dashboard_skip_only_run_is_neither_passing_nor_recording` is gone, replaced by `test_dashboard_zero_work_run_reports_the_failure_the_pipeline_records`, which builds the real shape (skip cell **plus** the synthetic fail entry) and asserts `failing`, no `skipping` pill, and — the part I care most about — that the `no matrix entry ran (empty matrix or all skipped -- check GPUs)` reason actually reaches the page rather than sitting behind a bare `fail` badge.
- The all-skip classification keeps one test, `test_dashboard_all_skip_summary_is_labelled_skipping_not_passing`, whose comment states outright that it is defensive and unreachable for `nightly_eval` output, and points at the test above for production behaviour. It's there so that if someone later relaxes the zero-work guard, the "skip-only reads as passing" bug can't silently come back.
- Both the `_latest_status` docstring and the `nothing_graded` comment now name the invariant and say the branch only guards a summary from some other producer, so the next reader doesn't take `skipping` for live behaviour. The docstring previously asserted the opposite, which is how this got past me twice.

Suite green: 22 in `tests/ci/test_dashboard_and_alert.py`, 53 across `tests/ci`.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Applied, essentially as suggested:

```python
raw = summary.get(m)
unit = _METRIC_UNITS.get(m, "")
# A unit on an unknown value ("— ms") reads as a measurement of nothing.
val = f"{_fmt_num(raw)} {unit}".strip() if _isnum(raw) else _fmt_num(raw)
```

This was the only site with the problem — the step-time column goes through `_fmt_ms`, which returns a bare `—` and never appends a unit, and `_METRIC_UNITS` has no other consumer.

`test_dashboard_omits_the_unit_when_the_value_is_unknown` covers both directions in one doc: a `None` metric renders with no unit, while a real value alongside it keeps its `ms`. Rendering the live `data.json` is byte-identical before and after, and `— ms` never appeared there, which matches your read that it only bites on malformed data.

amd-vivekag (commented):
## Re-review (commit 8716b0a)

**Verdict: ship** — two tiny Lows, both from Copilot's latest pass; I'm holding the approve only because you asked me to gate it on finding nothing else.

**Both of my prior findings are fixed correctly.**
- *Medium (unreachable `skipping`):* ideal resolution. The misleading test is replaced by `test_dashboard_zero_work_run_reports_the_failure_the_pipeline_records`, which builds the real shape (skip cell + synthetic `_nightly_eval` fail) and asserts `failing` plus the "no matrix entry ran" reason on the page; the defensive branch stays but the docstring/comment now say plainly it's unreachable for this pipeline. False confidence gone.
- *Low (`— ms`):* applied as suggested, with a two-direction test. `tests/ci` is green (53).

**Two new Lows from Copilot's 15:01 pass** — both real, both cosmetic:

- **[Low]** `_isnum` accepts non-finite floats. `json.loads` parses `NaN`/`Infinity`, and a metric mean can legitimately be `NaN`, so it would reach a sparkline as `nan` coords or a bar as `width:nan%`. Add `math.isfinite`.
- **[Low]** `_svg_sparkline`'s `values: list[float]` hint is now stale — callers pass `None`/`bool`. Widen to `list[float | None]`.

This is the minor tail — the PR is in great shape. Say the word and I'll approve now treating these two as non-blocking, or I'll approve once they're in.

scripts/ci/gen_dashboard.py
```diff
@@ -51,6 +59,19 @@ def _metric_policy(_name: str):  # type: ignore
...
+
+
+def _isnum(v: Any) -> bool:
+    """Numeric for display purposes.
+
+    ``bool`` is a subclass of ``int``, so a stray ``true`` in the (untrusted)
+    results JSON would otherwise format as a step time, scale a bar, or count
+    towards a trend.
+    """
+    return isinstance(v, (int, float)) and not isinstance(v, bool)
```
amd-vivekag: **[Low]** `_isnum` accepts non-finite floats. `json.loads` parses `NaN`/`Infinity` (and a metric mean can genuinely come out `NaN`), so it flows into `_svg_sparkline` as `nan` coordinates and `_bar` as `width:nan%`. Cosmetic, but easy to close and consistent with the untrusted-JSON stance elsewhere.

Fix (needs `import math`):
```suggestion
    return isinstance(v, (int, float)) and not isinstance(v, bool) and math.isfinite(v)
```

scripts/ci/gen_dashboard.py
```diff
@@ -66,15 +87,15 @@ def load_results(results_dir: Path) -> list[dict[str, Any]]:
 
 def _svg_sparkline(values: list[float], width: int = 160, height: int = 32) -> str:
```
amd-vivekag: **[Low]** Stale type hint. `_svg_sparkline` is called with `history[k]` and `passrate`, which hold `None` (and are `_isnum`-filtered for `bool`), so `values: list[float]` misleads callers.

Fix:
```suggestion
def _svg_sparkline(values: list[float | None], width: int = 160, height: int = 32) -> str:
```

amd-vivekag (approved):
Approving. Both of my earlier findings (the unreachable `skipping` state / misleading test, and the `— ms` unit) are fixed correctly, and `tests/ci` is green (53). The two remaining items are cosmetic Lows (non-finite `_isnum`, stale `_svg_sparkline` type hint) — worth a quick follow-up but not blocking.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Good catch — and it turned out to be worse than cosmetic in two ways, so I widened the fix slightly rather than taking the one-liner.

First, the visible damage isn't `width:nan%`. `_bar` does `max(2.0, min(100.0, nan))`, and `min(100.0, nan)` returns `100.0` because the `nan < 100.0` comparison is False — so a `NaN` step time silently drew a **full-width** bar, reading as the slowest cell in its group rather than as obviously broken.

Second, the suggested `math.isfinite(v)` alone would have turned a rendering glitch into a crash. `math.isfinite` raises `OverflowError` on an int too large to convert to a float, and `json.loads('{"n": 999...}')` (400 digits) yields exactly that. `_fmt_num` and `_fmt_ms` already crashed on such a value before this change — I confirmed `build_dashboard_html` raised `OverflowError: int too large to convert to float` on HEAD and renders `—` after. So the guard catches both:

```python
if not isinstance(v, (int, float)) or isinstance(v, bool):
    return False
try:
    return math.isfinite(v)
except OverflowError:  # int beyond the float range
    return False
```

Chasing it through the callers turned up two more sites that don't go through `_isnum` at all:

- the pass-rate card divided summary counts directly, so a non-finite count printed `nan%`. It now uses the shared guard — which also retires the `isinstance(passrate[-1], float)` exception I defended in the earlier thread. That reasoning was only ever about `bool`, and `NaN` *is* a `float`, so the exception no longer holds.
- the count cards rendered `_esc(str(v))`, which prints `nan` (or `True`) straight onto the page. They now go through `_fmt_num`, so a malformed summary shows `—`. My own new test caught this one after I'd fixed the first two.

`test_dashboard_rejects_values_it_cannot_render` covers `NaN`, `±Infinity`, the oversized int, `True`, `None` and a string across `_isnum`/`_fmt_num`/`_fmt_ms`/`_bar`, then asserts end-to-end that a document carrying a `NaN` step time and an oversized metric renders without raising, with no `nan` anywhere and no bar. Rendering the live `data.json` is byte-identical before and after. `tests/ci` green at 54.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed — the hint now matches what every caller has always passed:

```python
def _svg_sparkline(
    values: list[float | None], width: int = 160, height: int = 32
) -> str:
    """Tiny inline SVG line chart for a metric's history (ignores non-numbers)."""
```

Wrapped across lines to stay inside the line length. I also adjusted the docstring from "ignores None" to "ignores non-numbers", since after the `_isnum` change on the other thread it now skips `bool`, `NaN`/`Infinity`, and unfloatable ints as well, not just `None`.

`_has_trend` was already annotated `list[list[float | None]]`, so the two helpers that read the same series now agree.

PR #333, opened 2026-08-06
ci(sanitizers): fix nightly checkout perms + Ninja-less build

amd-vivekag:
## Summary
Two independent failures were blocking the **Sanitizers Nightly** gfx950 job on the shared self-hosted MI350 runner:

1. **`Check out code` — `EACCES: permission denied, rmdir '.../results/gpu_smoke'`.** Sibling GPU jobs (`gpu-tests.yml` et al.) run `aorta` inside a **root** ROCm CI container that leaves root-owned artifacts in the mounted checkout; `actions/checkout`'s `git clean` runs as the runner user and can't delete them. Added the same **Reclaim workspace ownership** step already used by `gpu-tests.yml` / `eval-reusable.yml` / `refresh-baselines.yml` (throwaway root `busybox` container `chown`, with `sudo docker` / host `sudo` fallbacks) before checkout. `sanitizers-nightly.yml` was the only GPU workflow missing it.
2. **`Build rocjitsu sanitizer binaries` — `CMake was unable to find a build program corresponding to "Ninja"`.** The runner has no Ninja on PATH. Now prefers Ninja only when present, otherwise falls back to CMake's default generator, and builds via `cmake --build --target ...`.

See failing [run 31102697357](https://github.com/ROCm/aorta/actions/runs/31102697357).

## Test plan
- [ ] Re-run **Sanitizers Nightly** via `workflow_dispatch` and confirm checkout succeeds and the rocjitsu build + downstream ConSan/waitcheck steps run.

### DIFF
--- a/.github/workflows/sanitizers-nightly.yml
+++ b/.github/workflows/sanitizers-nightly.yml
@@ -52,8 +52,16 @@ jobs:
           git clone --filter=blob:none https://github.com/ROCm/rocm-systems.git "${ROCJITSU_SRC}"
           git -C "${ROCJITSU_SRC}" checkout "${ROCJITSU_REF}"
           echo "ROCJITSU checkout: $(git -C "${ROCJITSU_SRC}" rev-parse HEAD)"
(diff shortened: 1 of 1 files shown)


vivekkhandelwal1 (approved):
LGTM

PR #357, opened 2026-08-10
ROCm stack health dashboard UX

vivekkhandelwal1:
## Summary

- **Clickable workloads** — category tiles, nightly health grid columns/cells, and workload titles link to reproduction steps (active workloads only)
- **Reproduce locally** — step-by-step panels per workload (GPU requirements, install guide, exact command, recipe link)
- **Nightly release health** — customer-facing overall health per release row
- **Category health tiles** — Platform · Inference · Training · Correctness
- **Training scaling table** — weak-scaling efficiency (`p50_2 / p50_8`) for 2→8 GPU recipes
- **Change summary** — correctness vs performance grouping
- **status.json** — structured latest-night summary (`schema_version: 2`); **`data.json` unchanged** (top-level history array for existing pollers)
- **Failure next steps** — env probe + ROCm issue guidance

Presentation-only in `gen_dashboard.py` / `dashboard_metadata.py`.

## Test plan

- [x] `pytest tests/ci/test_dashboard_and_alert.py` (77 tests)
- [ ] Verify `/data.json` and `/ci/data.json` remain identical arrays after deploy
- [ ] Verify `/status.json` for embed consumers

### DIFF
--- a/scripts/ci/gen_dashboard.py
+++ b/scripts/ci/gen_dashboard.py
@@ -34,7 +34,7 @@ except Exception:  # pragma: no cover - dashboard must render even if import fai
         return None
 
     def _is_correctness_metric(_name: str) -> bool:  # type: ignore
-        return _name.endswith("_checksum") or _name.endswith("checksum")
+        return _name.endswith("checksum")
 
     def _is_performance_metric(name: str) -> bool:  # type: ignore
         return name in _METRIC_UNITS and name not in (
@@ -167,25 +167,56 @@ def _engineer_only_metrics() -> frozenset[str]:
     return _DEFAULT_ENGINEER_METRICS | frozenset(str(m) for m in extra)
 
 
-def _category_for_workload(entry_name: str) -> str:
-    for cat_id, cat in (load_dashboard_metadata().get("categories") or {}).items():
-        if entry_name in (cat.get("workloads") or []):
-            return str(cat_id)
-    return "other"
+# Correctness counters: any night-over-night change is significant.
+_CORRECTNESS_COUNTERS = frozenset({
+    "ranks_with_divergence",
+    "layer_checksum_mismatches",
+})
 
 
-def _format_headline_metric(name: str, raw: Any) -> str:
+def _is_correctness_change_metric(name: str) -> bool:
+    """True when any value change in this metric should surface in Correctness."""
+    return _is_correctness_metric(name) or name in _CORRECTNESS_COUNTERS
+
+
+def _checksum_headline(name: str, raw: Any, entry: dict[str, Any]) -> str:
+    """Headline for checksum metrics — neutral unless comparison confirms match."""
+    label = name.replace("_", " ")
+    if not _isnum(raw):
+        return f"{label}: —"
+    deltas = entry.get("deltas") or {}
+    metric_delta = (deltas.get("metrics") or {}).get(name)
+    if metric_delta:
+        policy = metric_delta.get("policy")
+        expected = metric_delta.get("value")
+        observed = metric_delta.get("observed", raw)
+        if policy == "equal" and expected is not None:
+            if observed == expected:
+                return f"{label}: match ✓"
+            return f"{label}: mismatch ✗"
+    verdict = str(entry.get("verdict", ""))
+    if verdict == "fail":
+        needle = name.replace("_", " ")
+        for reason in entry.get("reasons") or []:
+            text = str(reason).lower()
+            if name in text or needle in text or "checksum" in text:
+                return f"{label}: mismatch ✗"
+    return f"{label}: captured ({_fmt_num(raw)})"
+
+
+def _format_headline_metric(
+    name: str, raw: Any, entry: dict[str, Any] | None = None,
+) -> str:
     """One headline metric for customer view."""
-    if name in ("ranks_with_divergence", "layer_checksum_mismatches"):
+    entry = entry or {}
+    if name in _CORRECTNESS_COUNTERS:
         if _isnum(raw) and float(raw) == 0:
             return f"{name.replace('_', ' ')}: 0 ✓"
         if _isnum(raw):
             return f"{name.replace('_', ' ')}: {_fmt_num(raw)} ✗"
         return f"{name.replace('_', ' ')}: —"
     if name in ("logits_checksum", "output_checksum", "checksum"):
-        if _isnum(raw):
-            return f"{name.replace('_', ' ')}: match ✓"
-        return f"{name.replace('_', ' ')}: —"
+        return _checksum_headline(name, raw, entry)
     unit = _METRIC_UNITS.get(name, "")
     if name == "mean_step_time_ms":
         return f"step time: {_fmt_ms(raw)}"
@@ -209,7 +240,7 @@ def _headline_block(entry: dict[str, Any], entry_name: str) -> str:
             raw = metrics.get(name)
         else:
             raw = summary.get(name)
-        parts.append(_format_headline_metric(str(name), raw))
+        parts.append(_format_headline_metric(str(name), raw, entry))
     if not parts:
         return ""
     return (
@@ -258,7 +289,8 @@ def build_category_summary(
         color = _VERDICT_COLOR.get(worst, _UNKNOWN_COLOR)
         glyph = _VERDICT_GLYPH.get(worst, _UNKNOWN_GLYPH)
         tally_txt = _tally_text(counts)
-        anchor = f"wl-{_esc(str(workloads[0]))}" if workloads else ""
+        present = [str(wl) for wl in workloads if by_entry.get(str(wl))]
+        anchor = f"wl-{_esc(present[0])}" if present else ""
         tiles.append(
             f"<a class='cat-tile' href='#{anchor}' style='border-left-color:{color}'>"
             f"<div class='cat-k'>{_esc(label)}</div>"
@@ -712,21 +744,34 @@ def build_change_summary(results: list[dict[str, Any]]) -> str:
         after_m = _measurements(latest_by[k])
         for metric in sorted(set(before_m) & set(after_m)):
             b, a = before_m[metric], after_m[metric]
+            if b == a:
+                continue
+            if _is_correctness_change_metric(metric):
+                movers_corr.append((0, 0, k, metric, b, a, latest_by[k]))
+                continue
             if not b:  # a move from zero has no meaningful percentage
                 continue
             pct = (a - b) / abs(b) * 100.0
             if abs(pct) > _MOVE_PCT:
                 row = (abs(pct), pct, k, metric, b, a, latest_by[k])
-                if _is_correctness_metric(metric):
-                    movers_corr.append(row)
-                elif _is_performance_metric(metric):
+                if _is_performance_metric(metric):
                     movers_perf.append(row)
                 else:
                     movers_other.append(row)
-    movers_corr.sort(reverse=True)
+    movers_corr.sort(key=lambda r: (r[5], r[4]), reverse=True)
     movers_perf.sort(reverse=True)
     movers_other.sort(reverse=True)
     movers = movers_corr + movers_perf + movers_other
+
+    def _corr_mover_li(row: tuple) -> str:
+        _, _, _k, metric, b, a, entry = row
+        unit = _METRIC_UNITS.get(metric, "")
+        return (
+            f"<li class='corr'><span class='mono'>{_esc(_cell_label(entry))}</span> "
+            f"<span class='mono'>{_esc(metric)}</span> "
+            f"<span class='muted'>{_esc(_fmt_num(b))} → {_esc(_fmt_num(a))}"
+            f"{(' ' + _esc(unit)) if unit else ''}</span></li>"
+        )
 
     def _mover_li(row: tuple, css: str) -> str:
         _, pct, _k, metric, b, a, entry = row
@@ -742,7 +787,7 @@ def build_change_summary(results: list[dict[str, Any]]) -> str:
     if movers_corr:
         items.append("<li class='change-hdr corr'>Correctness</li>")
         for row in movers_corr[:4]:
-            items.append(_mover_li(row, "corr"))
+            items.append(_corr_mover_li(row))
     if movers_perf:
         items.append("<li class='change-hdr perf'>Performance</li>")
         for row in movers_perf[:4]:
--- a/scripts/ci/gen_dashboard.py
+++ b/scripts/ci/gen_dashboard.py
@@ -128,6 +128,7 @@ _CUSTOMER_STATUS = {
     "unrecognised": "Unrecognised verdict",
     "failing": "Regression detected",
     "passing": "Healthy",
+    "partial": "Partial run",
     "recording": "Baseline setup",
     "skipping": "Incomplete run",
 }
@@ -408,6 +409,10 @@ def _latest_status(results: list[dict[str, Any]]) -> tuple[str, str]:
     if total > sum(_count(s.get(v)) for v in _VERDICT_ORDER):
         return "unrecognised", _UNKNOWN_COLOR
     if _count(s.get("pass")):
+        # A mix of pass and skip is not a fully healthy run — some workloads
+        # never executed (e.g. not enough GPUs for the 8-GPU matrix entries).
+        if _count(s.get("skip")):
+            return "partial", _VERDICT_COLOR["record"]
         return "passing", _VERDICT_COLOR["pass"]
     if _count(s.get("record")):
         return "recording", _VERDICT_COLOR["record"]
@@ -1072,6 +1077,14 @@ def build_dashboard_html(results: list[dict[str, Any]]) -> str:
             f"<div class='notice'>Nothing ran: all {total} results were "
             f"<strong>skipped</strong>, so this build establishes nothing. Check that "
             f"the runner exposes as many GPUs as the matrix asks for.</div>"
+        )
+    elif status == "partial":
+        pass_now = _count(s.get("pass"))
+        notices.append(
+            f"<div class='notice'>This run is <strong>partial</strong>: "
+            f"{pass_now} of {total} results passed but {skip_now} workload(s) were "
+            f"<strong>skipped</strong> — not every configured matrix entry ran. "
+            f"Check that the runner exposes as many GPUs as the matrix asks for.</div>"
         )
     elif shared_reason:
         notices.append(
--- a/scripts/ci/gen_dashboard.py
+++ b/scripts/ci/gen_dashboard.py
@@ -27,9 +27,19 @@ sys.path.insert(0, str(Path(__file__).resolve().parent))
 try:
     import eval_lib  # for metric_policy display
     _metric_policy = eval_lib.metric_policy
+    _is_correctness_metric = eval_lib.is_correctness_metric
+    _is_performance_metric = eval_lib.is_performance_metric
 except Exception:  # pragma: no cover - dashboard must render even if import fails
     def _metric_policy(_name: str):  # type: ignore
         return None
+
+    def _is_correctness_metric(_name: str) -> bool:  # type: ignore
+        return _name.endswith("_checksum") or _name.endswith("checksum")
+
+    def _is_performance_metric(name: str) -> bool:  # type: ignore
+        return name in _METRIC_UNITS and name not in (
+            "logits_checksum", "output_checksum", "checksum"
+        )
 
 # Links back to the source of truth. This dashboard only ever describes this
 # repo's nightly run, so the slug is fixed rather than plumbed through.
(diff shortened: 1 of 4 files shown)


scripts/ci/gen_dashboard.py
```diff
@@ -104,6 +114,167 @@ def _metric_policy(_name: str):  # type: ignore
...
+    if name in ("ranks_with_divergence", "layer_checksum_mismatches"):
+        if _isnum(raw) and float(raw) == 0:
+            return f"{name.replace('_', ' ')}: 0 ✓"
+        if _isnum(raw):
+            return f"{name.replace('_', ' ')}: {_fmt_num(raw)} ✗"
+        return f"{name.replace('_', ' ')}: —"
+    if name in ("logits_checksum", "output_checksum", "checksum"):
+        if _isnum(raw):
+            return f"{name.replace('_', ' ')}: match ✓"
+        return f"{name.replace('_', ' ')}: —"
```
amd-vivekag: **[High]** `match ✓` is shown for any numeric checksum, but a captured number only proves a checksum exists — not that it matched a baseline. On record-only runs (the "Baseline setup" state this PR adds) or when a blessed baseline omits the metric, this renders a green correctness pass with no comparison behind it. That's exactly the false-positive a customer health view must not emit. (Copilot flagged this too.)

Fix: pass the comparison outcome down and only assert a match when equality is confirmed; otherwise render a neutral captured state. e.g. thread the entry verdict/policy result into `_format_headline_metric` and return `checksum: captured` (not `✓`) unless the checksum was actually compared and equal.

scripts/ci/gen_dashboard.py
```diff
@@ -543,26 +716,50 @@ def build_change_summary(results: list[dict[str, Any]]) -> str:
...
-                movers.append((abs(pct), pct, k, metric, b, a, latest_by[k]))
-    movers.sort(reverse=True)
-    for _, pct, _k, metric, b, a, entry in movers[:6]:
+                row = (abs(pct), pct, k, metric, b, a, latest_by[k])
+                if _is_correctness_metric(metric):
+                    movers_corr.append(row)
+                elif _is_performance_metric(metric):
+                    movers_perf.append(row)
+                else:
+                    movers_other.append(row)
```
amd-vivekag, line 936: **[Medium]** Correctness (equal-policy) metrics are bucketed into the new "Correctness" heading only after passing the performance `_MOVE_PCT` (10%) gate above. For an exact-equality checksum, *any* inequality is a regression — a 2% checksum drift is silently dropped, so the Correctness section under-reports the very thing it's named for. (Copilot flagged this too.)

Fix: classify correctness metrics before the `_MOVE_PCT` threshold and emit every unequal value as a Correctness change; keep `_MOVE_PCT` for performance/other movers only.

scripts/ci/gen_dashboard.py
```diff
@@ -27,10 +27,20 @@
...
     import eval_lib  # for metric_policy display
     _metric_policy = eval_lib.metric_policy
+    _is_correctness_metric = eval_lib.is_correctness_metric
+    _is_performance_metric = eval_lib.is_performance_metric
 except Exception:  # pragma: no cover - dashboard must render even if import fails
     def _metric_policy(_name: str):  # type: ignore
         return None
 
+    def _is_correctness_metric(_name: str) -> bool:  # type: ignore
+        return _name.endswith("_checksum") or _name.endswith("checksum")
```
amd-vivekag: **[Low]** `_name.endswith("_checksum")` is a subset of `_name.endswith("checksum")`, so the first clause is dead. This fallback also diverges from the real `eval_lib.is_correctness_metric` (policy == "equal", i.e. the `logits_checksum`/`output_checksum`/`checksum` allowlist), so on import failure metric grouping shifts.

Fix:
```suggestion
        return _name.endswith("checksum")
```

scripts/ci/gen_dashboard.py
```diff
@@ -104,6 +114,167 @@ def _metric_policy(_name: str):  # type: ignore
...
+        for wl in workloads:
+            group.extend(by_entry.get(str(wl), []))
+        if not group:
+            continue
+        counts = _tally(group)
+        worst = _worst_verdict(counts)
+        color = _VERDICT_COLOR.get(worst, _UNKNOWN_COLOR)
+        glyph = _VERDICT_GLYPH.get(worst, _UNKNOWN_GLYPH)
+        tally_txt = _tally_text(counts)
+        anchor = f"wl-{_esc(str(workloads[0]))}" if workloads else ""
```
amd-vivekag: **[Low]** The tile anchors to `workloads[0]`, but the group renders from whichever workloads actually have results. If only the 8gpu variant ran (e.g. `training_ddp_8gpu` present, `training_ddp` absent), the link targets `wl-training_ddp`, which has no rendered row, so the tile jumps nowhere.

Fix: anchor to the first workload actually present in `group` rather than `workloads[0]`.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

Solid, well-contained presentation change: metadata is cleanly separated, the import-fallback keeps the dashboard rendering, and test coverage for the new tiles/panels is good. Two issues are worth fixing before merge — both center on the customer-facing view asserting *correctness* signals it hasn't actually verified, which is the one thing this dashboard shouldn't get wrong. I independently confirmed the two Copilot findings; they're real.

**Findings:** 1 High · 1 Medium · 2 Low

- **[High]** `scripts/ci/gen_dashboard.py` (~L185): checksum headline renders `match ✓` for any numeric checksum, with no baseline comparison — shows green on record-only / "Baseline setup" runs where nothing was compared.

No change to `data.json` schema or the CI harness, so the blast radius is contained to the rendered page. Verdict badges/category tiles still reflect true pass/record/fail; the two issues are about the *supplementary* correctness chips and the "What changed" grouping.

scripts/ci/gen_dashboard.py
```diff
@@ -543,26 +716,50 @@ def build_change_summary(results: list[dict[str, Any]]) -> str:
...
             pct = (a - b) / abs(b) * 100.0
             if abs(pct) > _MOVE_PCT:
-                movers.append((abs(pct), pct, k, metric, b, a, latest_by[k]))
-    movers.sort(reverse=True)
-    for _, pct, _k, metric, b, a, entry in movers[:6]:
+                row = (abs(pct), pct, k, metric, b, a, latest_by[k])
+                if _is_correctness_metric(metric):
+                    movers_corr.append(row)
+                elif _is_performance_metric(metric):
+                    movers_perf.append(row)
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in 950a2d6: correctness metrics are checked before `_MOVE_PCT`; any inequality surfaces under Correctness with absolute values (incl. 0→1). Tests added.

scripts/ci/gen_dashboard.py
```diff
@@ -104,6 +114,167 @@ def _metric_policy(_name: str):  # type: ignore
...
+    """One headline metric for customer view."""
+    if name in ("ranks_with_divergence", "layer_checksum_mismatches"):
+        if _isnum(raw) and float(raw) == 0:
+            return f"{name.replace('_', ' ')}: 0 ✓"
+        if _isnum(raw):
+            return f"{name.replace('_', ' ')}: {_fmt_num(raw)} ✗"
+        return f"{name.replace('_', ' ')}: —"
+    if name in ("logits_checksum", "output_checksum", "checksum"):
+        if _isnum(raw):
+            return f"{name.replace('_', ' ')}: match ✓"
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in 950a2d6: checksum headlines use `_checksum_headline()` — record-only runs show `captured (value)`; `match ✓` only when `deltas.metrics[name]` confirms equality; `mismatch ✗` on fail. Test: `test_headline_checksum_record_only_shows_captured_not_match`.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in 950a2d6 (same as Copilot thread): `_checksum_headline()` only asserts match when comparison deltas confirm equality; otherwise neutral captured/mismatch text. Tests: `test_headline_checksum_record_only_shows_captured_not_match`, `test_headline_checksum_shows_match_when_comparison_confirms`.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in 950a2d6: `_is_correctness_change_metric()` runs before the percentage gate; every unequal correctness metric is listed under Correctness with absolute deltas. Tests: `test_change_summary_reports_correctness_counter_without_pct_gate`, `test_change_summary_reports_checksum_inequality_below_pct_gate`.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in 950a2d6: import-fallback now uses `_name.endswith("checksum")` only, matching `eval_lib.is_correctness_metric` (equal-policy allowlist).

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in 950a2d6: category tile anchor uses the first workload in the category that actually has results (`present = [wl for wl in workloads if by_entry.get(wl)]`), not `workloads[0]`. Test: `test_category_tile_anchors_to_workload_that_actually_ran`.

scripts/ci/gen_dashboard.py
```diff
@@ -533,36 +736,75 @@ def build_change_summary(results: list[dict[str, Any]]) -> str:
...
+                continue
             if not b:  # a move from zero has no meaningful percentage
                 continue
             pct = (a - b) / abs(b) * 100.0
             if abs(pct) > _MOVE_PCT:
-                movers.append((abs(pct), pct, k, metric, b, a, latest_by[k]))
-    movers.sort(reverse=True)
-    for _, pct, _k, metric, b, a, entry in movers[:6]:
+                row = (abs(pct), pct, k, metric, b, a, latest_by[k])
+                if _is_performance_metric(metric):
```
amd-vivekag: **[Medium]** Performance grouping misses the dashboard's own headline metrics.

`eval_lib.is_performance_metric` only returns true for the min/max gating allowlist (`gflops`, `gbps`, `tokens_per_sec`, `samples_per_sec`, `throughput`, `prefill_latency_ms`, `decode_latency_ms`, `latency_ms`). So the training headline metrics `step_time_p50`/`step_time_p99` and the harness timings `mean_step_time_ms`/`mean_wall_clock_sec` (the gpu_smoke headline) fall into the unlabelled `movers_other` bucket — capped at 2 and shown without the Performance heading. A training step-time regression therefore isn't grouped under Performance, which is the headline feature of this PR. Note the import-fallback `_is_performance_metric` *would* classify them correctly, so the two code paths disagree. (Copilot flagged this too, suppressed.)

Fix: classify performance for display off the dashboard's own set, not the eval gate — e.g. treat anything in `_METRIC_UNITS` that isn't a checksum/correctness metric as performance.

scripts/ci/gen_dashboard.py
```diff
@@ -533,36 +736,75 @@ def build_change_summary(results: list[dict[str, Any]]) -> str:
...
     since = _fmt_timestamp(str(prev.get("generated_at") or "")).split(" ")[0]
     if not items:
         body = (
             "<p class='steady'>Nothing moved: same workloads, same verdicts, and no "
             f"metric changed by more than {_MOVE_PCT:.0f}%.</p>"
         )
     else:
-        extra = len(movers) - 6
         more = (
             f"<p class='muted'>and {extra} more metric"
```
amd-vivekag: **[Low]** The "past 10%" footer over-claims for hidden correctness movers.

`extra` is `len(movers) - shown`, and `movers` includes `movers_corr`, which were added *without* passing the `_MOVE_PCT` gate. With >4 correctness changes, this footer can say "and 1 more metric past 10%" for a `0 → 1` counter or a sub-10% checksum drift. (Copilot flagged this too, suppressed.)

Fix: count `extra` from the perf/other buckets only, or use threshold-neutral wording (e.g. "and N more changes").

scripts/ci/gen_dashboard.py
```diff
@@ -104,6 +114,199 @@ def _metric_policy(_name: str):  # type: ignore
...
+
+def _customer_status_label(status: str) -> str:
+    return _CUSTOMER_STATUS.get(status, status.replace("_", " ").title())
+
+
+def _workload_meta(entry_name: str) -> dict[str, Any]:
+    return (load_dashboard_metadata().get("workloads") or {}).get(entry_name) or {}
+
+
+def _engineer_only_metrics() -> frozenset[str]:
```
amd-vivekag: **[Low]** Dead code: `_engineer_only_metrics()` (and `_DEFAULT_ENGINEER_METRICS` / the `engineer_only_metrics` metadata list) is defined but never called. The headline-vs-engineer split is driven solely by the `headline_metrics` allowlist, so this list has no effect.

Fix: either wire it in (filter these out of the headline row / engineer table) or drop it and the metadata field.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

Re-review of `950a2d6`. All four findings from the last round are genuinely fixed and verified: checksum headlines now render `captured (…)` unless a comparison confirms equality (`_checksum_headline`), correctness metrics are classified before the `_MOVE_PCT` gate, the import-fallback is now `endswith("checksum")`, and category tiles anchor to a workload that actually ran — each backed by a new test. Nice, well-covered work.

The gap sweep on the new code surfaces one Medium worth fixing before merge: the new **What changed → Performance** grouping doesn't actually catch the dashboard's own headline timing metrics, because it gates on `eval_lib.is_performance_metric` (the eval allowlist), not on what the dashboard displays. Two Lows alongside it. These confirm the two suppressed Copilot comments on this commit.

**Findings:** 0 High · 1 Medium · 2 Low

Prior review: 4/4 resolved (verified). No previously-flagged High items remain open.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in 0468dfd: added `_is_dashboard_performance_metric()` using `_METRIC_UNITS` + harness timings (not the eval-lib gating allowlist), so `step_time_p50`/`p99` and `mean_step_time_ms` appear under Performance. Tests updated/added.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in 0468dfd: footer now splits hidden correctness changes (`and N more correctness changes`) from threshold-gated perf/other (`past 10%`). Test: `test_change_summary_footer_splits_correctness_from_thresholded`.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in 0468dfd: removed unused `_engineer_only_metrics()`, `_DEFAULT_ENGINEER_METRICS`, and the `engineer_only_metrics` metadata list — headline metrics are driven solely by each workload's `headline_metrics` allowlist.

scripts/ci/gen_dashboard.py
```diff
@@ -104,6 +114,195 @@ def _metric_policy(_name: str):  # type: ignore
...
+        "categories": {}, "workloads": {},
+    }
+
+# Plain-language headline for each internal _latest_status() label.
+_CUSTOMER_STATUS = {
+    "unknown": "Unknown",
+    "empty": "No data",
+    "unrecognised": "Unrecognised verdict",
+    "failing": "Regression detected",
+    "passing": "Healthy",
```
amd-vivekag, line 130: **[High]** A mixed pass-and-skip run is still labeled `Healthy`.

`_latest_status()` returns `passing` as soon as any pass exists, and this new customer mapping hides that other workloads never ran; no incomplete-run notice is shown either.

Fix: Check for skips before returning `passing` (or add an explicit partial state), so `Healthy` requires zero skipped workloads, and add a pass+skip regression test.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

Third-pass re-review of `0468dfd`. All seven prior findings are fixed correctly: performance grouping now uses dashboard metrics, footer counts are split accurately, dead engineer metadata is gone, and the earlier checksum/correctness/anchor fixes remain intact. The latest suppressed Copilot import concern does not hold because this module prepends `scripts/ci` to `sys.path` before importing its sibling metadata module.

The gap sweep found one customer-facing status issue: a partial run with both passes and skips is still promoted to the top-level **Healthy** label.

**Findings:** 1 High · 0 Medium · 0 Low

- **[High]** Reserve `Healthy` for runs with no skipped workloads.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — thanks.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — thanks.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — thanks.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in c901a25: `_latest_status()` returns `partial` (maps to **Partial run**, not Healthy) when pass and skip coexist; added a banner explaining skipped workloads. Test: `test_dashboard_pass_and_skip_is_not_labelled_healthy`.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

Fourth-pass re-review of `c901a25`. The partial-run fix is correct and well tested: pass+skip now maps to **Partial run**, includes an explanatory banner, and no longer renders **Healthy**. All eight prior human findings are resolved.

I independently confirmed the new Copilot finding on this head: checksum values are converted from signed int64 to float in `triage/matrix.py` before the dashboard receives them. Distinct values above 2^53 can therefore compare equal, so the customer view can still show `match ✓` for a real mismatch.

**Findings:** 1 High · 0 Medium · 0 Low

- **[High]** Preserve checksum values exactly through aggregation/comparison before claiming a match. Until then, keep the headline neutral (`captured`) and do not treat float equality as correctness confirmation.

Not approving while this false-positive correctness path remains.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — thanks.

scripts/ci/gen_dashboard.py
```diff
@@ -104,6 +114,196 @@ def _metric_policy(_name: str):  # type: ignore
...
+    deltas = entry.get("deltas") or {}
+    metric_delta = (deltas.get("metrics") or {}).get(name)
+    if metric_delta:
+        policy = metric_delta.get("policy")
+        expected = metric_delta.get("value")
+        observed = metric_delta.get("observed", raw)
+        if policy == "equal" and expected is not None:
+            if observed == expected:
+                return f"{label}: match ✓"
+            return f"{label}: mismatch ✗"
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in 9c96eae. `_checksum_headline()` no longer uses `observed == expected` on float deltas.

- **match ✓** only when `verdict == "pass"` and the baseline delta has `policy == "equal"` (harness already validated)
- **mismatch ✗** when `verdict == "fail"` and the failure reasons mention the checksum metric
- otherwise **captured (value)** — including record-only runs

Added a regression test where colliding float equality in deltas would previously show match ✓ but the harness failed.

The change-summary `b == a` skip at line ~747 is unchanged: nightly JSON only carries float summaries, so cross-night trend detection cannot recover exact int64 equality without pipeline changes. The user-facing false-match bug was in the headline path.

scripts/ci/gen_dashboard.py
```diff
@@ -104,6 +114,352 @@ def _metric_policy(_name: str):  # type: ignore
...
+    """
+    label = name.replace("_", " ")
+    if not _isnum(raw):
+        return f"{label}: —"
+    verdict = str(entry.get("verdict", ""))
+    deltas = entry.get("deltas") or {}
+    metric_delta = (deltas.get("metrics") or {}).get(name)
+    if verdict == "fail" and _checksum_metric_failed(entry, name):
+        return f"{label}: mismatch ✗"
+    if verdict == "pass" and metric_delta and metric_delta.get("policy") == "equal":
```
amd-vivekag: **[High]** A `pass` verdict still cannot confirm an exact checksum match.

`compare_to_baseline()` receives the same float-rounded `metrics_summary` values, so two distinct int64 checksums can collide and produce `verdict == pass`; this branch still renders a false `match ✓`.

Fix: Preserve exact checksum values through matrix aggregation/baseline comparison, or keep this headline neutral until an exact-value channel exists.

scripts/ci/gen_dashboard.py
```diff
@@ -104,6 +114,352 @@ def _metric_policy(_name: str):  # type: ignore
...
+    )
+    rows = []
+    for two_gpu, eight_gpu, label in pairs:
+        p50_2 = _step_p50(two_gpu)
+        p50_8 = _step_p50(eight_gpu)
+        if p50_2 is None or p50_8 is None:
+            continue
+        ratio = p50_8 / p50_2 if p50_2 else None
+        # Ideal 8-GPU step time with perfect 4× scaling from 2 GPUs.
+        ideal_8 = p50_2 / 4.0
```
amd-vivekag: **[High]** This applies strong-scaling math to a weak-scaling workload.

Both matrix entries use the same recipe with `batch_size: 2` per rank, so global work grows 4× from 2 to 8 GPUs; ideal step time is flat, yet equal times render as 25% efficiency.

Fix: Compute weak-scaling efficiency as `p50_2 / p50_8 * 100` and update the copy/test, or use separate recipes that hold global batch fixed before claiming ideal 4× speedup.

scripts/ci/gen_dashboard.py
```diff
@@ -393,7 +753,10 @@ def build_history_grid(results: list[dict[str, Any]], max_runs: int = _GRID_RUNS
         return ""
 
     head = "".join(
-        f"<th scope='col' class='wlcol' title='{_esc(n)}'>{_esc(n)}</th>" for n in ordered
+        f"<th scope='col' class='wlcol'>"
+        f"<a href='#wl-{_esc(n)}' title='{_esc(n)} — jump to workload'>"
```
amd-vivekag: **[Medium]** Retired workload columns link to fragments that do not exist.

`ordered` intentionally retains workloads absent from the newest run, but `#wl-*` sections are rendered only from newest entries, so those history links are dead.

Fix: Render retired headings/cells as plain text, or emit a historical stub section with the matching anchor.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

The partial-run status remains fixed, and the expanded dashboard is well tested. However, the newest commit adds breaking and misleading behavior, and the checksum fix is only partial: the harness `pass` verdict is derived from the same float-rounded checksum comparison, so trusting that verdict still cannot prove an exact match.

**Findings:** 3 High · 1 Medium · 0 Low

- **[High] Still open from last review:** preserve checksums exactly before rendering `match ✓`; a float-derived `pass` verdict is not independent confirmation.
- **[High]** Keep the existing array-root `/data.json` contract or publish the v2 object at a new endpoint. Copilot flagged this too; `.github/workflows/pages.yml` explicitly promises `/ci/data.json` remains unchanged.
- **[High]** Correct the training scaling formula: current recipes keep batch size per rank, so this is weak scaling and ideal step time is flat, not 4× faster.
- **[Medium]** Do not link retired history workloads to `#wl-*` fragments that are absent from the current workload table.

Not approving while these gaps remain.

scripts/ci/gen_dashboard.py
```diff
@@ -1035,7 +1560,8 @@ def main() -> int:
         results = results[-args.max_builds:]
     args.out_dir.mkdir(parents=True, exist_ok=True)
     (args.out_dir / "index.html").write_text(build_dashboard_html(results), encoding="utf-8")
-    (args.out_dir / "data.json").write_text(json.dumps(results, indent=2), encoding="utf-8")
+    (args.out_dir / "data.json").write_text(
+        json.dumps(build_data_json(results), indent=2), encoding="utf-8")
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in 5f879c7. `data.json` is back to the original top-level history array for `/ci/data.json` pollers. The structured v2 summary now lives in a separate `status.json` (linked from the header nav). Pages verify step updated to require `status.json`.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in 5f879c7. Removed pass→match branch; checksum headlines only show mismatch ✗ from failure reasons, otherwise captured (value).

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in 5f879c7. Weak-scaling efficiency p50_2/p50_8 with updated copy and test.

scripts/ci/gen_dashboard.py
vivekkhandelwal1: Fixed in 5f879c7. Retired history columns are plain text; only active workloads link to #wl-*.

amd-vivekag (approved):

scripts/ci/gen_dashboard.py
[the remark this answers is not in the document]
amd-vivekag: Resolved — thanks.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — thanks.

scripts/ci/gen_dashboard.py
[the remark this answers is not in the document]
amd-vivekag: Resolved — thanks.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — thanks.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — thanks.

PR #362, opened 2026-08-11
feat(sanitizers): render caller-supplied ConSan code objects in the nightly dashboard (informational, non-gating)

amd-vivekag:
## Summary

Follow-up to #347. Renders ConSan runs over **caller-supplied code objects / commands**
(`source.consan_command`, added in #347) in the **sanitizer nightly dashboard**, as a
**non-gating "informational" section** alongside the three baseline-gated daily controls.
This lets us share the current guardrail state (drive ConSan over real/extracted GEMM
code objects and small synthetic kernels) without touching the daily gate.

The objects are **generic, public hipBLASLt Tensile content** (extracted in CI from the
synthetic `gemm_shapes_unique.csv`) and small synthetic kernels — **no customer data**,
**no binaries committed** (all built/extracted by the nightly).

## What's in it

- **`gen_sanitizer_dashboard.py`** — new optional `--informational-results-dir`; enumerates
  `DIR/<case>/sanitizer_report.json` and renders an "Informational · caller-supplied code
  objects (non-gating)" section (verdict + the fail-closed **reason**, which the gated
  tables omit). It **does not feed the gate**; a missing dir degrades to no section.
- **`prepare_gemm_isa.py`** — fixes the copy-one-blob bug: extracts the real **per-transpose
  f32 `Type_SS`** Tensile object (16 MB / 490 kernels) instead of a bf8 `SB` blob copied to
  every `sol_*.hsaco`, and emits one heavy object for the ConSan case. The daily
  `waitcheck-gemm` baseline (`warn`) is **unchanged** (validated on gfx950).
- **New generic fixtures** (`recipes/sanitizers/fixtures/kernels/*.hip`): `lds_reduce` +
  `lds_dispatch` (dispatched), `tiny_vecadd`, and a generic load-only `consan_load` driver.
- **New recipes**: `daily-consan-{gemm,lds-dispatch,tiny}.yaml` (informational, non-gating).
- **Nightly wiring**: builds the fixtures, runs the 3 informational recipes into
  `out/informational/<case>/`, and passes them to the dashboard. The gate comparator keys
  off a fixed case list, so informational cases can never fail the nightly.

## Current state it renders

On the current rocjitsu build these fail closed (never a false pass) — documenting the
open upstream issues, and they flip to `pass` once fixed:

| Case | Verdict | Reason | Upstream |
|---|---|---|---|
| consan-gemm (16 MB f32 GEMM) | error | `combined_hook_timeout` | ROCm/rocm-systems#9964 / #9970 |
| consan-lds-dispatch | error | `combined_hook_exit_86` (0 records) | ROCm/rocm-systems#9972 |
| consan-tiny | error | `combined_hook_exit_86` (no sites) | by design |

## Test plan

Validated locally on a gfx950 host (rocjitsu build):
- `pytest tests/sanitizers/test_dashboard.py` → **36 passed** (incl. new informational tests). `ruff` clean.
- Extractor produces the real 16 MB / 490-kernel f32 objects; `daily-waitcheck-gemm` still **warn** (64 findings, baseline holds), ~129 s.
- All 3 informational recipes run end-to-end (fail-closed reasons above); the rendered dashboard shows **HEALTHY — 3/3** (daily gate unchanged) plus the informational section.
- Workflow YAML parses; the 3 recipes load under the schema.

Note: the full nightly path (CI fixture build + publish) is validated by design (graceful degradation keeps the daily gate safe) but needs a live `sanitizers-nightly` run to confirm end-to-end.

Made with Cursor


.github/workflows/sanitizers-nightly.yml
```diff
@@ -149,7 +149,25 @@ jobs:
...
+              --csv "${FIXTURES}/gemm_shapes_unique.csv" --out "${FIXTURES}/isa" --top-n 3 \
+              --consan-object "${FIXTURES}/isa/consan_gemm_f32.hsaco"
+
+            # Informational (non-gating) ConSan-over-code-object fixtures (#347):
+            # the small synthetic kernels + their load-only / dispatch drivers, and
+            # the heavy f32 SS object extracted above. All generic; no customer data.
+            KROOT="${FIXTURES}/kernels"
+            hipcc --genco --offload-arch=gfx950 "${KROOT}/lds_reduce.hip" -o /tmp/_lds.o
+            clang-offload-bundler --type=o --unbundle --input=/tmp/_lds.o \
+              --targets=hipv4-amdgcn-amd-amdhsa--gfx950 --output="${FIXTURES}/isa/lds.hsaco"
```
vivekkhandelwal1: `hipcc --genco` already emits a standalone loadable code object — the HRX module path writes it directly (`src/aorta/workloads/hrx.py` uses `-o …/kernel.hsaco` with no bundler step). Feeding genco output back through `clang-offload-bundler --unbundle` assumes a fat bundle and can fail on a plain code object.

Suggest writing straight to the final path, e.g.:
```bash
hipcc --genco --offload-arch=gfx950 "${KROOT}/lds_reduce.hip" -o "${FIXTURES}/isa/lds.hsaco"
hipcc --genco --offload-arch=gfx950 "${KROOT}/tiny_vecadd.hip" -o "${FIXTURES}/isa/tiny.hsaco"
```

This block runs under `set -euo pipefail` **before** the gated recipes, so a bundler failure currently aborts the whole nightly, not just the informational section.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -570,11 +570,94 @@ def _kernel_detail_html(rows: dict[str, dict[str, Any]]) -> str:
...
+            continue
+        summary = summarize_case(report, report.get("overall_verdict"))
+        primary = ("\u2014", "\u2014", "\u2014")
+        preflight = "\u2014"
+        for check in report.get("checks", []):
+            san = check.get("sanitizer")
+            if san == "waitcheck_preflight":
+                preflight = str(check.get("verdict"))
+            else:
+                primary = (str(san), str(check.get("verdict")), str(check.get("reason")))
```
vivekkhandelwal1: When these cases eventually pass (after rocm-systems fixes), successful checks serialize `reason` as JSON `null`. `str(check.get("reason"))` becomes the literal string `None` in the dashboard table.

Normalize absent values to the existing em-dash placeholder, e.g.:
```python
reason = check.get("reason")
primary = (..., reason if reason is not None else "\u2014")
```
Same applies to `preflight` if a passing preflight omits a reason.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -570,11 +570,94 @@ def _kernel_detail_html(rows: dict[str, dict[str, Any]]) -> str:
...
+        f"<td>{_observed_html(c['verdict'])}</td>"
+        f"<td class=mono>{_esc(_clean_msg(c['reason'], 200))}</td>"
+        f"<td>{_observed_html(c['preflight'])}</td></tr>"
+        for c in cases
+    )
+    return (
+        "<h2>Informational &middot; caller-supplied code objects (non-gating)</h2>"
+        "<p class=secondary>Experimental ConSan runs over caller-supplied kernels/objects "
+        "(<span class=mono>source.consan_command</span>, #347). These do <b>not</b> affect the "
+        "gate above. On the current rocjitsu build they fail closed (never a false pass); the "
```
vivekkhandelwal1: This copy will become incorrect as soon as upstream fixes land and a case shows `pass`. The table already reports the live verdict/reason — the section intro should stay outcome-neutral, e.g. "These do **not** affect the gate above; observed verdict and fail-closed reason (when present) are shown below."

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -570,11 +570,94 @@ def _kernel_detail_html(rows: dict[str, dict[str, Any]]) -> str:
...
+
+def build_informational_md(cases: list[dict[str, Any]]) -> str:
+    if not cases:
+        return ""
+    lines = [
+        "## Informational \u00b7 caller-supplied code objects (non-gating)",
+        "",
+        "Experimental ConSan runs over caller-supplied kernels/objects "
+        "(`source.consan_command`, #347). These do **not** affect the gate. On the current "
+        "rocjitsu build they fail closed (never a false pass).",
```
vivekkhandelwal1: Same stale wording issue as the HTML paragraph above — drop "On the current rocjitsu build they fail closed" so the markdown section stays accurate when cases start passing.

recipes/sanitizers/fixtures/kernels/consan_load.hip
```diff
@@ -0,0 +1,24 @@
+// Generic, self-contained load-only ConSan target: hipModuleLoad one code object
+// (path baked in at build time via -DOBJECT="...") and resolve its first kernel, so
```
vivekkhandelwal1: The header says this driver "resolve[s] its first kernel", but `main` only calls `hipModuleLoad` / `hipModuleUnload` — no `hipModuleGetFunction` or launch. Please adjust the comment to match the load-only behaviour (instrumentation at module load), so readers don't expect a dispatch step.

vivekkhandelwal1 (changes requested):
## Summary

Good follow-up to #347: the informational dashboard section is cleanly separated from the baseline gate (`CASES` / `compare_verdict_baselines.py` unchanged), the `prepare_gemm_isa.py` f32-per-transpose fix addresses a real fixture bug, and the new tests verify that informational rows never affect the HEALTHY gate banner.

I am **not approving** yet — a few items below should be fixed before merge, mainly around CI fixture build reliability and dashboard text that will go stale once upstream rocjitsu fixes land.

## What looks good

- **Gate isolation** is sound: informational results live under `out/informational/<case>/`, the comparator still keys off the fixed three-case map, and `test_informational_section_renders_reason_and_is_non_gating` guards the HEALTHY banner.
- **`prepare_gemm_isa.py` refactor** (real f32 `Type_SS` per transpose instead of copying one bf8 blob) is the right fix for meaningful waitcheck/ConSan fixtures.
- **Synthetic HIP fixtures** (`lds_dispatch` using `hipModuleLaunchKernel`, load-only `consan_load`) are well motivated and documented in code.
- **Dashboard reason column** for informational cases fills a real gap vs the gated tables.

## Required before merge

1. **Workflow `.hsaco` build** — `hipcc --genco` output should be written directly to the final path (see inline on `sanitizers-nightly.yml`). The extra `clang-offload-bundler --unbundle` pass matches the Tensile bundle path, not standalone genco output (`hrx.py` writes genco straight to `.hsaco`). Under `set -euo pipefail` a bundler failure here aborts the entire GPU job **before** the daily gate runs.
2. **Null `reason` rendering** — when upstream fixes make these cases pass, `str(None)` will show the literal `None` in the Reason column (inline on `gen_sanitizer_dashboard.py`).
3. **Outcome-neutral section copy** — the HTML/MD boilerplate currently asserts the current build always fails closed; that becomes wrong the moment a case passes (inline comments).

## Suggested (non-blocking but worth doing)

- **Isolate informational fixture build from the gate**: even after fixing genco, consider running the informational fixture compile + three `aorta sweep` steps *after* `compare_verdict_baselines.py`, or guard them so a failure cannot take down the gated path. The PR goal is observability without risking the 3/3 gate.
- **`prepare_gemm_isa.py` tests**: the extraction logic changed materially but has no unit coverage (layout map, missing bundle error). Even a small test for `_LAYOUT` / `_SS_HEAVY` formatting would help.
- **Pass-case dashboard test**: extend `test_informational_from_dir_*` with a report whose checks have `"reason": null` once the null-handling fix lands.
- **`consan_load.hip` header** — minor doc accuracy (inline).

## Test plan gap

CPU CI covers dashboard rendering; the full fixture-build path (`hipcc --genco`, Tensile unbundle, three informational sweeps) only executes on the live `sanitizers-nightly` GPU job. Fixing item (1) and ideally reordering/isolating the informational build would reduce risk to the daily gate on the first nightly run post-merge.

.github/workflows/sanitizers-nightly.yml
```diff
@@ -149,7 +149,25 @@ jobs:
...
+            hipcc --genco --offload-arch=gfx950 "${KROOT}/tiny_vecadd.hip" -o /tmp/_tiny.o
+            clang-offload-bundler --type=o --unbundle --input=/tmp/_tiny.o \
+              --targets=hipv4-amdgcn-amd-amdhsa--gfx950 --output="${FIXTURES}/isa/tiny.hsaco"
```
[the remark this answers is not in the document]
amd-vivekag: Good catch — fixed in 8d6fc69. On this ROCm build `hipcc --genco` actually emits a clang-offload **bundle** (so writing it straight to `.hsaco` would hand the loader/waitcheck a bundle, not a raw code object), but on other builds it emits a raw ELF where `--unbundle` fails. So the fixture build is now conditional: unbundle only when the `--genco` output starts with `__CLANG_OFFLOAD_BUNDLE__`, otherwise copy it directly — robust to both toolchains.

scripts/sanitizers/gen_sanitizer_dashboard.py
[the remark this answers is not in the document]
amd-vivekag: Good catch — fixed in 8d6fc69. `informational_from_dir` now normalizes absent check fields (a passing check serializes `reason=null`) to the em-dash placeholder instead of the literal `None`, and I added a regression test asserting a passing case renders `—` and that `None` never appears in the section.

recipes/sanitizers/fixtures/kernels/consan_load.hip
```diff
@@ -0,0 +1,24 @@
...
+// the combined rocjitsu ConSan hook instruments exactly that object at load time.
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in 8d6fc69 — the driver is load-only (`hipModuleLoad` only, no `hipModuleGetFunction`), so I dropped the 'resolve its first kernel' claim from the comment.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -570,11 +570,94 @@ def _kernel_detail_html(rows: dict[str, dict[str, Any]]) -> str:
...
+        f"<td class=mono>{_esc(_clean_msg(c['reason'], 200))}</td>"
+        f"<td>{_observed_html(c['preflight'])}</td></tr>"
+        for c in cases
+    )
+    return (
+        "<h2>Informational &middot; caller-supplied code objects (non-gating)</h2>"
+        "<p class=secondary>Experimental ConSan runs over caller-supplied kernels/objects "
+        "(<span class=mono>source.consan_command</span>, #347). These do <b>not</b> affect the "
+        "gate above. On the current rocjitsu build they fail closed (never a false pass); the "
+        "exact reason is shown.</p>"
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in 8d6fc69 — made the informational section's prose verdict-neutral in both the HTML and the Markdown twin (line 639/641, matching the suppressed comment): it now states the table reports each case's observed verdict and reason for this run, rather than asserting they fail closed. Stays accurate once an upstream rocjitsu fix flips a case to pass.

.github/workflows/sanitizers-nightly.yml
amd-vivekag: Thanks — fixed in 8d6fc69. On this ROCm build `--genco` actually emits a clang-offload *bundle*, but on others (like the HRX path you cite) it's a raw ELF where `--unbundle` would fail. So the build now unbundles only when the output starts with `__CLANG_OFFLOAD_BUNDLE__`, else copies it directly — robust to both.

scripts/sanitizers/gen_sanitizer_dashboard.py
amd-vivekag: Thanks — fixed in 8d6fc69. Absent check fields (a passing check's `reason=null`) now normalize to the em-dash placeholder instead of literal `None`, with a regression test covering the pass case.

scripts/sanitizers/gen_sanitizer_dashboard.py
amd-vivekag: Thanks — fixed in 8d6fc69. The HTML prose is now verdict-neutral (reports each case's observed verdict + reason for the run) so it stays correct once a case flips to pass.

scripts/sanitizers/gen_sanitizer_dashboard.py
amd-vivekag: Thanks — fixed in 8d6fc69, same change applied to the Markdown twin (dropped the 'fail closed' assertion).

recipes/sanitizers/fixtures/kernels/consan_load.hip
amd-vivekag: Thanks — fixed in 8d6fc69. It's load-only (`hipModuleLoad` only), so I dropped the 'resolve its first kernel' claim from the header.

.github/workflows/sanitizers-nightly.yml
amd-vivekag: Follow-up in 9910d75 for your 'aborts the gate before compare_verdict_baselines' concern: the **entire** informational path (fixture build + the three sweeps) now runs in a guarded `( set +e … ) || true` subshell placed **before** the gate, so no genco/bundler/compile failure can reach the gated step. The gated waitcheck fixtures build independently; the guarded block extracts only the ConSan object (`--top-n 0`). On genco itself: on this ROCm build it emits a *bundle*, so I unbundle conditionally (only when the output starts with `__CLANG_OFFLOAD_BUNDLE__`) rather than writing directly — robust to both the bundle and raw-ELF toolchains.

amd-vivekag:
Thanks for the thorough review @vivekkhandelwal1 — addressed everything across `8d6fc69` + `9910d75`.

**Required**
1. **Workflow build aborting the gate** — `9910d75`: the whole informational path (fixture build + the 3 sweeps) now runs in a guarded `( set +e … ) || true` subshell **before** the gate, so a genco/bundler/compile failure can't reach `compare_verdict_baselines.py`. On genco itself: on this ROCm build `--genco` emits a clang-offload *bundle* (writing it straight to `.hsaco` would hand the loader a bundle, and waitcheck a non-raw object), so I unbundle **conditionally** (only when the output is a bundle) rather than always — robust to both the bundle and raw-ELF (`hrx.py`) toolchains.
2. **Null `reason` rendering** — `8d6fc69`: absent check fields normalize to the em-dash placeholder instead of literal `None`; regression test added.
3. **Outcome-neutral copy** — `8d6fc69`: HTML + MD prose is verdict-neutral (reports each case's observed verdict/reason for the run).

**Suggested**
- **Isolate informational build from the gate** — done via the guarded subshell above; `--top-n 0` extracts only the ConSan object so the gated waitcheck fixtures are untouched.
- **`prepare_gemm_isa.py` tests** — `9910d75`: added `tests/sanitizers/test_prepare_gemm_isa.py` (layout map, SS-heavy bundle-name formatting, CSV/top-n selection incl. `0`, fail-closed missing-bundle path).
- **Pass-case dashboard test** — landed in `8d6fc69`: `test_informational_from_dir_*` now includes a `reason: null` pass case asserting `—` and that `None` never appears.
- **`consan_load.hip` header** — `8d6fc69`.

Local validation: 50 sanitizer tests pass, ruff clean, YAML parses. Re-requesting your review.

vivekkhandelwal1 (approved):
## Re-review — approve

The two follow-up commits (`8d6fc69`, `9910d75`) address all items from the earlier review:

- **`genco_object()`** — detects clang-offload bundles vs raw code objects before unbundling; robust across ROCm builds.
- **Null `reason` handling** — `_cell()` normalizes absent fields to the em-dash placeholder; covered by the new passing-case test.
- **Verdict-neutral section copy** — HTML/MD no longer assert the current build always fails closed.
- **`consan_load.hip`** — header matches load-only behaviour.
- **Gate isolation** — the entire informational path (fixture build + sweeps) runs in a guarded `( set +e … ) || true` subshell after the gated sweeps and before `compare_verdict_baselines.py`; gated waitcheck fixtures stay on `--top-n 3` independently.
- **`test_prepare_gemm_isa.py`** — layout map, bundle naming, CSV selection (`top-n 0`), and missing-bundle fail-closed path.

CI is green (CPU py3.10–3.12, GPU MI350, pre-commit). Gate isolation design remains sound and the informational section is a good observability layer for #347 without touching the 3/3 nightly gate.

PR #363, opened 2026-08-11
Dashboard layout v2: card layout and exhaustive repro guides

vivekkhandelwal1:
## Summary
- Replace the wide nightly release health table with vertical release cards and wrapping workload chips (no horizontal scroll for history).
- Replace the single mega-table of workloads with category-grouped card grids (Platform, Inference, Training, Correctness).
- Add exhaustive per-workload reproduction guides in `dashboard_metadata.py`: prerequisites, install/setup, dry-run, run command, verify steps, success criteria, and compare notes.

## Test plan
- [x] `pytest tests/ci/test_dashboard_and_alert.py` (80 tests)
- [ ] Spot-check generated HTML locally with two+ nightly runs
- [ ] Confirm release history chips link to `#wl-*` and open repro panels via hash
- [ ] Verify distributed workloads show torchrun/NCCL/race-specific setup in repro panels


scripts/ci/dashboard_metadata.py
```diff
@@ -7,6 +7,94 @@
 
 from typing import Any
 
+_READ_MATRIX = [
+    "ls -lt triage_results/*/*/$(ls -t triage_results/*/* | head -1 | xargs basename) 2>/dev/null || true",
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in a186b62: replaced the brittle ls glob with a workload-scoped find command via `_read_matrix_cmds()`, and verification paths now use the recipe artifact workload name (not the dashboard entry id).

scripts/ci/gen_dashboard.py
```diff
@@ -813,74 +882,39 @@ def _wl_col_header(name: str) -> str:
...
-        rows.append(
-            f"<tr><th scope='row' class='runcell'>{when}{flag}"
-            f"<span class='runmeta'>{_esc(str(build.get('amd_aorta_version') or ''))}</span></th>"
-            f"<td class='center'><span class='badge sm' style='background:{color}' "
-            f"title='{_esc(status)}'>{_esc(customer_status)}</span></td>"
-            f"{''.join(cells)}</tr>"
+        chips = "".join(
+            _wl_chip(name, by_entry.get(name) or [], retired=name not in active)
+            for name in ordered
+        )
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in a186b62: release history chips now use native `<ul class="wl-chips"><li>...</li></ul>` markup instead of `<div role="list">`, so each chip is a proper list item for assistive tech.

vivekkhandelwal1:
Also addressed the additional Copilot suggestions from the review summary (commit a186b62):

- **Artifact paths**: verification commands use recipe `workload:` names (`inference`, `training`, …) via explicit `artifact_workload`, not dashboard entry IDs.
- **`pip install --upgrade`**: nightly wheel install step now upgrades existing packages.
- **Inference prerequisites**: removed HuggingFace/cache requirement; notes built-in `RepeatedBlockModel` instead.
- **DDP prerequisites**: replaced NVLink wording with PCIe/xGMI for AMD/RCCL topology.

The two inline review threads are replied to and marked resolved.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

The card grouping is a useful improvement, and all current CI checks pass. The two prior inline Copilot threads and six first-pass suppressed concerns are resolved correctly, but the current reproduction commands still cannot be copied end-to-end and the card conversion introduces responsive and accessibility regressions.

**Findings:** 0 High · 8 Medium · 0 Low

scripts/ci/dashboard_metadata.py
```diff
@@ -7,6 +7,106 @@
...
+
+
+def _read_matrix_cmds(artifact_workload: str) -> list[str]:
+    """Shell helpers to locate and read the newest matrix for one recipe workload."""
+    wl = artifact_workload
+    return [
+        (
+            "find triage_results -mindepth 3 -maxdepth 3 -type d "
+            f"-path '*/{wl}/*' -printf '%T@ %p\\n' 2>/dev/null "
+            "| sort -rn | head -1 | cut -d' ' -f2-"
```
amd-vivekag: **[Medium]** Verification stops using the run directory it just found.

The first command only prints a path; the later `<TICKET>`/`<timestamp>` tokens are parsed by the shell as redirections, so the documented commands fail when copied.

Fix: assign the result to `RUN_DIR`, then use quoted paths such as `cat "$RUN_DIR/matrix.md"` in every base and workload-specific check.

scripts/ci/dashboard_metadata.py
```diff
@@ -63,6 +182,34 @@
...
+                ),
+                min_gpus=1,
+                distributed=False,
+                dry_run=(
+                    "aorta sweep run --recipe "
+                    "recipes/inference/example-inference-smoke.yaml --dry-run"
+                ),
+                verify_title="Inspect latency, throughput, and checksum artifacts",
+                verify_extra=[
+                    "grep -E 'tokens_per_sec|prefill|decode|checksum' "
```
amd-vivekag: **[Medium]** This grep targets an artifact that does not contain workload metrics.

`matrix.md` explicitly points these values to `perf.md`/`matrix.json`; the same issue affects the training, determinism, and race greps below.

Fix: read `"$RUN_DIR/perf.md"` or parse `matrix.json::cells[*].metrics_summary` (and trial JSON where per-trial detail is required).

scripts/ci/dashboard_metadata.py
```diff
@@ -7,6 +7,106 @@
...
+    setup = _install_setup(min_gpus=min_gpus, distributed=distributed)
+    if setup_extra:
+        setup.extend(setup_extra)
+    return {
+        "prerequisites": prerequisites,
+        "setup": setup,
+        "dry_run": {
+            "title": "Validate the recipe YAML (no GPU execution)",
+            "command": dry_run,
+        },
```
amd-vivekag, line 179: **[Medium]** This does not reproduce the nightly execution path promised by the guide.

Nightly adds `--strict` and runs `nightly_eval.py` to compare `matrix.json` with blessed baselines; a standalone sweep cannot produce the dashboard's pass/record verdicts used by the success sections.

Fix: mirror the harness invocation and comparison, or label this as a standalone smoke and replace baseline-verdict claims with local artifact checks.

scripts/ci/dashboard_metadata.py
```diff
@@ -74,6 +221,34 @@
...
+                    "recipes/training/example-training-ddp-smoke.yaml --dry-run"
+                ),
+                verify_title="Confirm both ranks finished and step times recorded",
+                verify_extra=[
+                    "grep -E 'step_time_p50|step_time_p99' "
+                    "triage_results/<TICKET>/training/<timestamp>/matrix.md",
+                ],
+                success=(
+                    "Both ranks exit cleanly; matrix.md lists step_time_p50/p99 per cell. "
+                    "Nightly grades lower step time as better vs blessed baseline."
```
amd-vivekag: **[Medium]** The guide claims performance grading that is not enabled.

`regression_baselines.yaml` has only `passed: true` for training and only checksum equality for inference, so step times, latency, and throughput do not currently determine a pass.

Fix: describe the active correctness gates, or add explicit performance baseline policies before saying these metrics are graded.

scripts/ci/gen_dashboard.py
```diff
@@ -1035,6 +1069,173 @@ def _mover_li(row: tuple, css: str) -> str:
...
+                    f"{'s' if n_metrics != 1 else ''}</summary>"
+                    f"<p class='prov muted'>Additional summary metrics for this recipe variant. "
+                    f"<strong>Grading rule</strong> is how nightly CI compares the value; "
+                    f"<strong>this run</strong> is tonight; "
+                    f"<strong>history</strong> is recent nights (when available).</p>"
+                    f"<p class='prov'>{' · '.join(prov)}</p>"
+                    f"<table class='inner'><thead><tr><th>metric</th>"
+                    f"<th class='center'>grading rule</th><th class='num'>this run</th>"
+                    f"{'<th>history</th>' if show_metric_trend else ''}</tr></thead>"
+                    f"<tbody>{mrows}</tbody></table>"
```
amd-vivekag: **[Medium]** Detailed metrics are clipped inside the new narrow cards.

The 160px history SVG plus the other columns exceeds a 340px card, while `.wl-card { overflow:hidden; }` removes access to the right-hand columns.

Fix: wrap this nested table in a horizontally scrollable `.tablewrap` and add a narrow-viewport regression check.

scripts/ci/gen_dashboard.py
```diff
@@ -1035,6 +1069,173 @@ def _mover_li(row: tuple, css: str) -> str:
...
+                f"</div>"
+            )
+
+        return (
+            f"<article class='wl-card' id='wl-{_esc(entry_name)}'>"
+            f"<header class='wl-card-hdr'>"
+            f"<div class='wl-head'>"
+            f"<a class='wl-title' href='#repro-{_esc(entry_name)}'>"
+            f"<strong>{_esc(wl_title)}</strong></a>"
+            f"<span class='mono muted wl-id'>{_esc(entry_name)}</span></div>"
```
amd-vivekag: **[Medium]** Workload cards have no heading in the document outline.

Categories are `h3`s, but the workload name is only a link/`strong` and its reproduction sections start at `h4`, so heading navigation loses the workload parent.

Fix: render the workload title as an `h4` and move reproduction subsection headings to `h5`.

scripts/ci/gen_dashboard.py
```diff
@@ -813,74 +882,39 @@ def _wl_col_header(name: str) -> str:
...
-            f"{''.join(cells)}</tr>"
+        chip_items = [
+            f"<li>{_wl_chip(name, by_entry.get(name) or [], retired=name not in active)}</li>"
+            for name in ordered
+        ]
+        cards.append(
+            f"<article class='release-card'>"
+            f"<header class='release-hdr'>"
+            f"<span class='release-date'>{when}{flag}</span>"
+            f"<span class='badge sm' style='background:{color}' "
```
amd-vivekag, line 954: **[Medium]** Each release article lost the heading previously supplied by its row header.

A generic date span does not name the article for assistive-technology heading navigation.

Fix: render the date as an `h3` (with reset margins) or connect the article to an equivalent label via `aria-labelledby`.

scripts/ci/gen_dashboard.py
```diff
@@ -1404,6 +1488,52 @@ def build_dashboard_html(results: list[dict[str, Any]]) -> str:
...
+                  margin-bottom:.55rem; }}
+  .release-date {{ font-weight:600; font-size:.88rem; }}
+  .wl-chips {{ display:flex; flex-wrap:wrap; gap:.4rem; list-style:none;
+                margin:0; padding:0; }}
+  .wl-chips > li {{ margin:0; padding:0; }}
+  .wl-chip {{ display:inline-flex; align-items:center; gap:.35rem; padding:.25rem .55rem;
+              border:1px solid var(--border); border-radius:999px; font-size:.75rem;
+              text-decoration:none; color:inherit; background:#12161b; max-width:100%; }}
+  a.wl-chip:hover {{ border-color:var(--accent); }}
+  .wl-chip.retired, .wl-chip.absent {{ opacity:.72; cursor:default; }}
```
amd-vivekag: **[Medium]** Absent-workload chip labels become effectively unreadable.

The existing `.absent { color:#39414a }` now applies to the whole chip (about 1.75:1 against `#12161b`) and this rule reduces it further with opacity.

Fix: keep label text at an accessible contrast and dim only the status glyph or border.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — each release article now has an h3 date heading.

scripts/ci/dashboard_metadata.py
amd-vivekag: Partially resolved: performance checks moved to perf.md, but the fallback cannot run after grep | head and determinism still checks matrix.md. I am re-raising those remaining cases.

scripts/ci/dashboard_metadata.py
amd-vivekag: Resolved — the guide now states that standalone --strict writes artifacts but does not apply dashboard baseline gates.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — the nested metrics table now has a horizontally scrollable wrapper.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — workload titles are h4 headings and reproduction subsections are h5.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — the low-contrast absent color and whole-chip opacity were removed.

scripts/ci/dashboard_metadata.py
amd-vivekag: The literal-placeholder issue is fixed. Exact-run selection is still workload-scoped, so I am carrying that remaining gap into this re-review.

scripts/ci/dashboard_metadata.py
amd-vivekag: Resolved — inference and training now distinguish active correctness gates from ungated performance metrics.

scripts/ci/dashboard_metadata.py
```diff
@@ -7,6 +7,129 @@
...
+    "expected; regressions vs the blessed baseline are what nightly CI flags."
+)
+
+_TRAINING_SUCCESS = (
+    "All ranks complete without error. step_time_p50/p99 appear in perf.md. "
+    "Nightly currently records training cells (passed/recording only); "
+    "step-time thresholds are not perf-gated in regression_baselines.yaml yet."
+)
+
+_METRICS_IN_PERF = (
```
amd-vivekag, line 22: **[Medium]** The JSON fallback never runs when `grep` finds no metrics.

Without `pipefail`, this pipeline returns `head`'s successful status, so `|| python ...` is skipped.

Fix: avoid the pipeline, for example `grep -m 10 -E ... || python ...`.

scripts/ci/dashboard_metadata.py
```diff
@@ -7,6 +7,129 @@
...
+    'print(json.dumps(d.get(\\\"cells\\\",{{}}), indent=2)[:2000])" "$RUN_DIR/matrix.json"'
+)
+
+
+def _read_matrix_cmds(artifact_workload: str) -> list[str]:
+    """Shell helpers to locate and read the newest matrix for one recipe workload."""
+    wl = artifact_workload
+    return [
+        (
+            f'RUN_DIR="$(find triage_results -mindepth 3 -maxdepth 3 -type d '
```
amd-vivekag: **[Medium]** Run discovery can select another guide's artifacts.

DDP, FSDP, and the 2/8-GPU variants share workload directories, so a failed or concurrent run can leave this pointing at a different recipe's newest matrix.

Fix: give each guide an entry-specific output directory and derive `RUN_DIR` from that invocation's emitted path.

scripts/ci/dashboard_metadata.py
```diff
@@ -121,6 +392,33 @@
...
+                ),
+                min_gpus=2,
+                distributed=True,
+                dry_run=(
+                    "aorta sweep run --recipe "
+                    "recipes/llm-determinism/example-llm-determinism.yaml --dry-run"
+                ),
+                verify_title="Look for rank divergence in matrix and trial JSON",
+                verify_extra=[
+                    'grep -i diverge "$RUN_DIR/matrix.md" 2>/dev/null | head',
```
amd-vivekag: **[Medium]** Determinism verification still cannot confirm the stated success criterion.

`matrix.md` omits `ranks_with_divergence`, and listing trial filenames does not inspect their values; the 8-GPU guide has the same gap.

Fix: read this metric from `perf.md`/`matrix.json::cells[*].metrics_summary` or parse the trial JSON contents.

scripts/ci/dashboard_metadata.py
```diff
@@ -7,6 +7,129 @@
...
+            "matrix.json to config/ci/regression_baselines.yaml (--strict)"
+        ),
+    ]
+
+
+def _install_setup(*, min_gpus: int, distributed: bool = False) -> list[dict[str, Any]]:
+    steps: list[dict[str, Any]] = [
+        {
+            "title": "Clone AORTA and install the nightly wheel channel",
+            "commands": [
```
amd-vivekag, line 106: **[Medium]** Setup does not install the build shown by the dashboard.

Cloning current `main` and installing the latest prerelease can silently use a different commit and wheel than the recorded `head_sha`/AORTA version.

Fix: populate the guide with the displayed commit and exact wheel version/artifact, then check out and install those pins.

scripts/ci/dashboard_metadata.py
```diff
@@ -48,6 +171,25 @@
...
+                recipe="recipes/ci/gpu-smoke.yaml",
+                run_command="aorta sweep run --recipe recipes/ci/gpu-smoke.yaml",
+                min_gpus=1,
+                distributed=False,
+                dry_run="aorta sweep run --recipe recipes/ci/gpu-smoke.yaml --dry-run",
+                verify_title="Read the smoke matrix",
+                success=(
+                    "Rank 0 prints `Wrote matrix to ...`. matrix.md shows every cell "
+                    "completed; mean_step_time_ms is recorded in perf.md. Nightly "
+                    "records this workload until a blessed baseline exists."
```
amd-vivekag: **[Medium]** This success note says no GPU-smoke baseline exists, but one is already blessed.

`regression_baselines.yaml` contains `gpu_smoke::baseline-local`, so nightly gates the cell outcome today.

Fix: state that the pass/fail outcome is gated while `mean_step_time_ms` remains recorded but not performance-thresholded.

scripts/ci/gen_dashboard.py
```diff
@@ -1035,6 +1072,174 @@ def _mover_li(row: tuple, css: str) -> str:
...
+            verdict = e.get("verdict", "skip")
+            color = _VERDICT_COLOR.get(verdict, "#57606a")
+            st = (e.get("metrics") or {}).get("mean_step_time_ms")
+            cell_name = e.get("cell")
+            label = (
+                f"<span class='mono'>{_esc(str(cell_name))}</span>"
+                if cell_name else "<span class='muted'>whole workload</span>"
+            )
+            trend_html = (
+                f"<span class='variant-spark'>{_svg_sparkline(history.get(k, []))}</span>"
```
amd-vivekag: **[Medium]** The card conversion leaves step-time trends unnamed for screen readers.

The old table header supplied `step time history`; this generic span/SVG has no role, accessible name, or textual values.

Fix: expose an image label containing the metric and oldest-to-newest values, and cover it with an accessibility test.

scripts/ci/gen_dashboard.py
```diff
@@ -780,24 +833,43 @@ def build_history_grid(results: list[dict[str, Any]], max_runs: int = _GRID_RUNS
...
+            )
+        counts = _tally(group)
+        worst = _worst_verdict(counts)
+        n = len(group)
+        worst_n = counts.get(worst, 0)
+        count = str(n) if worst_n == n else f"{worst_n}/{n}"
+        bg = _VERDICT_COLOR.get(worst, _UNKNOWN_COLOR)
+        breakdown = f"{name} — {_tally_text(counts)}"
+        dot = (
+            f"<span class='dot sm' role='img' style='background:{bg}' "
```
amd-vivekag, line 904: **[Medium]** Passing history chips fail normal-text contrast.

White on the pass green `#2ea043` is about 3.37:1, below the 4.5:1 requirement at the new `.dot.sm` font size.

Fix: use an AA-compliant foreground/background pair for the pass chip and add a contrast regression check.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

Six of the eight previous findings are fixed correctly; run discovery and metric verification are only partially fixed. All current CI checks pass, but those remaining gaps plus four newly identified reproducibility/accessibility issues still need attention.

**Findings:** 0 High · 7 Medium · 0 Low

scripts/ci/dashboard_metadata.py
amd-vivekag: Resolved — every dashboard entry now writes and discovers artifacts under its own output root.

scripts/ci/dashboard_metadata.py
amd-vivekag: Resolved — `grep -m` now preserves the non-zero status needed to reach the fallback.

scripts/ci/dashboard_metadata.py
amd-vivekag: Partially resolved: the guide now targets the right metric, but the fallback treats `cells` as a dict and `metrics_summary` omits failed trials. I am re-raising this at the new code.

scripts/ci/dashboard_metadata.py
amd-vivekag: Partially resolved: valid dashboard metadata now pins the SHA and version, but the missing-version fallback comments out the dev-wheel index and silently becomes unpinned. I am re-raising that edge case.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — workload sparklines now expose an oldest-to-newest accessible label.

scripts/ci/dashboard_metadata.py
amd-vivekag: Resolved — the guide now states the active gpu_smoke outcome gate and distinguishes ungated step time.

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — history chips now use an AA-compliant pass green.

scripts/ci/dashboard_metadata.py
```diff
@@ -7,6 +7,145 @@
...
+    '|| python -c "import json,sys; d=json.load(open(sys.argv[1])); '
+    'print(json.dumps(d.get(\\\"cells\\\",{{}}), indent=2)[:2000])" "$RUN_DIR/matrix.json"'
+)
+
+_DIVERGENCE_IN_ARTIFACTS = (
+    'grep -m 5 -E "ranks_with_divergence|diverge" "$RUN_DIR/perf.md" 2>/dev/null '
+    '|| python -c "import json,sys; d=json.load(open(sys.argv[1])); '
+    'cells=d.get(\\\"cells\\\",{}); '
+    'print(json.dumps({k:(v.get(\\\"metrics_summary\\\") or {}).get(\\\"ranks_with_divergence\\\") '
+    'for k,v in cells.items()}, indent=2))" "$RUN_DIR/matrix.json"'
```
amd-vivekag: **[Medium]** The determinism fallback uses the wrong JSON shape and the wrong data source.

`matrix.json::cells` is a list, so `.items()` crashes; moreover `metrics_summary` only aggregates passing trials, so a failed trial with actual divergence is omitted.

Fix: iterate each cell's `trial_paths` and read `result.metrics.ranks_with_divergence` from the raw trial JSON files.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

Five of the seven previous findings are fixed correctly. Determinism verification and the missing-build-metadata pin fallback remain incomplete; all current CI checks pass, but both user-facing repro paths can still fail or mislead.

**Findings:** 0 High · 2 Medium · 0 Low

scripts/ci/gen_dashboard.py
```diff
@@ -279,43 +283,119 @@ def _grading_rule_label(entry: dict[str, Any], name: str) -> str:
...
+        text = text.replace(
+            "git checkout {{HEAD_SHA}}",
+            "# git checkout <head_sha shown in the dashboard header>",
+        )
+    if ver:
+        text = text.replace("{{AORTA_VERSION}}", ver)
+    else:
+        text = text.replace(
+            "'amd-aorta[hw-queue]=={{AORTA_VERSION}}'",
+            "'amd-aorta[hw-queue]'  # pin to the AORTA version in the dashboard header",
```
amd-vivekag: **[Medium]** Missing version metadata produces a malformed unpinned install command.

The rendered inline `# pin ...` comments out the trailing `-f dev-wheels` argument, so the fallback searches the wrong source and no longer matches the dashboard build.

Fix: fail closed with a non-executable metadata warning, or emit the warning separately while preserving the full indexed install command.

amd-vivekag:
Addressed the remaining review feedback in b45ed8a:

- Replaced the invalid one-line determinism fallback with an executable here-doc that reads raw failed-trial metrics and handles repository-root-relative trial paths without double-prefixing them.
- Separated standalone `--strict` behavior from `nightly_eval.py` baseline comparison in both customer-facing explanations.
- Proactively standardized all new copyable Python/pip commands on `python3`.

Validation: 85 dashboard/alert tests pass; Ruff and the self-review knowledge-base pass are clean.

scripts/ci/dashboard_metadata.py
```diff
@@ -145,6 +508,38 @@
...
+                run_command=(
+                    "torchrun --standalone --nproc_per_node=2 $(which aorta) "
+                    "sweep run --recipe recipes/race/race_smoke.yaml"
+                ),
+                min_gpus=2,
+                distributed=True,
+                dry_run="aorta sweep run --recipe recipes/race/race_smoke.yaml --dry-run",
+                verify_title="Check layer checksum mismatches in matrix output",
+                verify_extra=[
+                    _METRICS_IN_PERF.format(pattern="layer_checksum_mismatch"),
```
amd-vivekag: **[Medium]** Race failures still hide the checksum-mismatch value this step asks users to verify.

`perf.md` and `metrics_summary` aggregate only passing trials, while this fallback merely dumps matrix metadata; failed trials therefore expose a path but not `layer_checksum_mismatches`. The 8-GPU guide has the same gap.

Fix: read `result.metrics.layer_checksum_mismatches` from each cell's raw `trial_paths`, as the determinism verifier now does.

scripts/ci/gen_dashboard.py
```diff
@@ -1035,6 +1123,176 @@ def _mover_li(row: tuple, css: str) -> str:
...
+        wl_summary = str(wl_meta.get("summary") or "")
+        summary_html = (
+            f"<p class='wl-sum muted'>{_esc(wl_summary)}</p>" if wl_summary else ""
+        )
+
+        variants: list[str] = []
+        for k in cell_keys:
+            e = latest_by_key[k]
+            verdict = e.get("verdict", "skip")
+            color = _VERDICT_COLOR.get(verdict, "#57606a")
```
amd-vivekag: **[Medium]** New passing card badges still fail normal-text contrast.

Variant badges here and release-health badges at line 954 use white text on `#2ea043` (about 3.37:1), below WCAG AA at their small font sizes.

Fix: use the accessible pass palette already defined for history chips in both new badge paths.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

Both findings from the previous review are fixed correctly, and all current CI checks pass. The full current-diff sweep found two remaining Medium issues in new card contrast and failed-race artifact verification.

**Findings:** 0 High · 2 Medium · 0 Low

scripts/ci/gen_dashboard.py
amd-vivekag: Resolved — release and variant pass badges now use the accessible green.

scripts/ci/dashboard_metadata.py
amd-vivekag: Partially resolved: race now shares the raw-trial helper, but a successful perf.md grep still skips failed-trial values. I am re-raising that remaining case on the current helper.

scripts/ci/dashboard_metadata.py
```diff
@@ -7,6 +7,180 @@
...
+
+_METRICS_IN_PERF = (
+    'grep -m 10 -E "{pattern}" "$RUN_DIR/perf.md" 2>/dev/null '
+    '|| python3 -c "import json,sys; d=json.load(open(sys.argv[1])); '
+    'print(json.dumps(d.get(\\\"cells\\\",[]), indent=2)[:2000])" "$RUN_DIR/matrix.json"'
+)
+
+def _trial_metric_in_artifacts(*, perf_grep_pattern: str, metric_key: str) -> str:
+    """Read a trial metric from perf.md when present, else from raw trial JSON."""
+    return (
```
amd-vivekag: **[Medium]** A successful perf grep hides failed-trial correctness values.

`perf.md` aggregates only passing trials, so a mixed run can print a passing zero, return success here, and skip the raw parser containing non-zero divergence/checksum values.

Fix: make the perf grep advisory (`grep ... || true`), always run the raw-trial parser afterward, and test mixed pass/fail trials for both metrics.

scripts/ci/gen_dashboard.py
```diff
@@ -1404,6 +1546,54 @@ def build_dashboard_html(results: list[dict[str, Any]]) -> str:
...
+                  margin-bottom:.55rem; }}
+  .release-date {{ font-weight:600; font-size:.88rem; margin:0; }}
+  .wl-chips {{ display:flex; flex-wrap:wrap; gap:.4rem; list-style:none;
+                margin:0; padding:0; }}
+  .wl-chips > li {{ margin:0; padding:0; }}
+  .wl-chip {{ display:inline-flex; align-items:center; gap:.35rem; padding:.25rem .55rem;
+              border:1px solid var(--border); border-radius:999px; font-size:.75rem;
+              text-decoration:none; color:inherit; background:#12161b; max-width:100%; }}
+  a.wl-chip:hover {{ border-color:var(--accent); }}
+  .wl-chip.retired {{ opacity:.72; cursor:default; }}
```
amd-vivekag: **[Medium]** Whole-chip opacity makes retired status badges fail contrast.

Compositing the badge and text at 72% opacity drops the small pass/fail/record labels below WCAG AA even with the darker pass palette.

Fix: keep the status badge fully opaque and dim only the retired chip's non-status label or border; add a retired-chip contrast regression test.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The previous card-badge contrast finding is fixed correctly; race verification is only partially fixed. All current CI checks pass, but the full current-diff sweep found one related correctness gap and one separate retired-chip accessibility regression.

**Findings:** 0 High · 2 Medium · 0 Low

amd-vivekag (approved):
There are some minor review comments that I think we can ignore for now. Otherwise, overall it looks good to me.

PR #376, opened 2026-08-12
Bump CI stack to ROCm 7.2.4 and PyTorch 2.10.0

vivekkhandelwal1:
## Summary
- Upgrade the pinned GPU CI base image from `rocm7.2` / PyTorch 2.9.1 to **ROCm 7.2.4** / **PyTorch 2.10.0** on the existing Ubuntu 22.04 + Python 3.10 lane (MI350 runner compatible)
- Update `Dockerfile.ci-gpu` digest, general-purpose `Dockerfile.rocm70_9-1`, and `docs/ci-testing-plan.md`

## Stack change
| | Before | After |
|---|---|---|
| ROCm | 7.2.0 | 7.2.4 |
| PyTorch | 2.9.1 | 2.10.0 |
| Base tag | `rocm7.2_ubuntu22.04_py3.10_pytorch_release_2.9.1` | `rocm7.2.4_ubuntu22.04_py3.10_pytorch_release_2.10.0` |

## Test plan
- [ ] **Bump validation** workflow passes on MI350 (full nightly matrix on this PR)
- [ ] Merge to `main`
- [ ] Run **Refresh baselines** workflow (`refresh-baselines.yml`) so perf thresholds reflect the new stack
- [ ] Confirm tonight's **Nightly Evaluation** reports the new `torch` / `rocm` pins on https://rocm.github.io/aorta/

## Notes
- Dashboard generator/layout unchanged; the live site picks up new version strings from the next nightly result JSON after this stack lands on the runner.
- If a larger jump (e.g. ROCm 7.14 / PyTorch 2.12) is required instead, say so and we can retarget the bump.

### DIFF
--- a/docker/.env.ci
+++ b/docker/.env.ci
@@ -4,7 +4,7 @@
 # digest for reproducible GPU CI runs on MI350 / gfx950 hardware.
 #
 # Update the digest when intentionally bumping the CI base image:
-#   docker buildx imagetools inspect rocm/pytorch:rocm7.2_ubuntu22.04_py3.10_pytorch_release_2.9.1
+#   docker buildx imagetools inspect rocm/pytorch:rocm7.2.4_ubuntu22.04_py3.10_pytorch_release_2.10.0
 
 DOCKERFILE=Dockerfile.ci-gpu
 IMAGE_NAME=aorta:ci-gpu
--- a/docker/Dockerfile.ci-gpu
+++ b/docker/Dockerfile.ci-gpu
@@ -13,12 +13,12 @@
 # has therefore been removed from this image. hipcc (from the ROCm base) still
 # builds the HIP repro fixtures and does not need any of those.
 #
-# Base: rocm/pytorch ROCm 7.2, Ubuntu 22.04, Python 3.10, PyTorch 2.9.1
+# Base: rocm/pytorch ROCm 7.2.4, Ubuntu 22.04, Python 3.10, PyTorch 2.10.0
 # Pinned as tag@digest (not digest-only): the tag anchors Dependabot to the
 # intended ROCm 7.2 stream when it proposes a new digest; the digest keeps the
 # build reproducible. Resolved via:
-#   docker buildx imagetools inspect rocm/pytorch:rocm7.2_ubuntu22.04_py3.10_pytorch_release_2.9.1
-FROM rocm/pytorch:rocm7.2_ubuntu22.04_py3.10_pytorch_release_2.9.1@sha256:376bfab5f4f680c8b4b843c6d0c5d1f0a04e5a84ec3e86728db8d11d79a9d1e3
+#   docker buildx imagetools inspect rocm/pytorch:rocm7.2.4_ubuntu22.04_py3.10_pytorch_release_2.10.0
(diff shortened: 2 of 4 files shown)


amd-vivekag:
Answering the question in your Notes — *"If a larger jump (e.g. ROCm 7.14 / PyTorch 2.12) is required instead, say so and we can retarget the bump"*: yes, we'd like the larger jump. We want CI (and therefore the nightly eval) reporting against the latest ROCm **production** release so it reflects the stack customers are actually on.

I've opened #380 for that, tracked by #379. Rather than ask you to redo the retarget, #380 already carries it plus the extra pieces we wanted, so I'd suggest closing this one in its favour — but please review #380, since it makes two calls yours deliberately avoided:

1. **It moves the OS/Python lane too** (Ubuntu 22.04 / py3.10 → 24.04 / py3.12), where you stayed put for MI350 compatibility. `requires-python` is `>=3.10` and 3.12 is already a supported classifier, but the real unknown is whether the runner's `amdgpu` driver supports a ROCm 7.14 userspace container. Bump-validate on #380 is what proves that. **If it fails there, your 7.2.4 bump is the right answer and we should come back to this PR** — so please don't delete the branch.
2. **It adds a new `Dockerfile.rocm-latest` and makes that the compose default, instead of bumping `Dockerfile.rocm70_9-1` in place.** Same end result for anyone on the default, but people who deliberately selected `rocm70_9-1` for an older-stack comparison don't get moved silently.

One thing your PR surfaced that's worth recording either way: PyTorch publishes no ROCm wheel index above [`rocm7.2`](https://download.pytorch.org/whl/rocm7.2) (`rocm7.14` 403s). That doesn't affect container images, which ship their own userspace ROCm, but it does cap anything pip-installing torch — including `aorta-internal`'s native GPU lane, which pins `rocm7.2` as "the floor". So the container and native paths will legitimately diverge after #380, and the internal comment asserting they match needs updating.

Thanks for getting the bump started — the digest/doc groundwork here is what made the bigger jump straightforward.

amd-vivekag:
Update: **you were right to stay on the 7.2.x lane**, and #380 has retargeted to your exact base — `rocm7.2.4_ubuntu22.04_py3.10_pytorch_release_2.10.0`.

I pushed a build-time guard asserting the classic `/opt/rocm` layout, and it rejected 7.14 in 7 seconds on the MI350 runner. The reason is that `rocm/pytorch` 7.14 installs ROCm from a wheel index into a venv (`PATH=/opt/venv/bin`, `npi.*` labels, no `/opt/rocm` at all), while every ROCm read in this repo is bound to `/opt/rocm`. So the larger jump I asked you about isn't available yet — your conservative choice was the correct one.

#380 now carries that same base plus a few things beyond it: a new default dev image (added alongside the existing ones rather than editing `Dockerfile.rocm70_9-1` in place, so nobody who selected it deliberately gets moved), corrections to the per-image version labels (`rocm70_9-1` was advertised as ROCm 7.0.9.1 but is `FROM rocm7.2…`; `7.0_9-1` is the amdgpu installer revision), and the layout guard.

Given that, #380 fully absorbs this one — happy for you to close it in favour of #380, though please keep the branch around. Thanks for starting the bump and for flagging the runner-compatibility constraint; that's what made the right answer obvious once the guard had data.

The 7.14 jump is tracked in #381 (make ROCm discovery layout-agnostic), after which the flip is a one-line change that the guard validates.

vivekkhandelwal1:
The ROCm version update is done in #380. Hence, closing this PR.

PR #386, opened 2026-08-17
feat(sanitizers): publish a reproducible run area per dashboard case

### LINKED ISSUE #384: Sanitizer dashboard (Tab 2): make the raw-report link a self-sufficient run area so the reproduce command is actionable
## Context

This concerns the live sanitizer-nightly dashboard at https://rocm.github.io/aorta/sanitizers/ — specifically **Tab 2 (Workload survey / observed-only)**, as delivered by #374 / #375 and restyled in #378.

Tab 2 already gives every survey case a copy-paste reproduce command, e.g.

```
aorta sweep run --recipe recipes/sanitizers/daily-consan-gemm.yaml
```

and a card-footer link labelled **view raw report** (`_raw_link_html()` in `scripts/sanitizers/gen_sanitizer_dashboard.py`).

## Problem

The footer link resolves to exactly one file:

```
runs/<id>/survey/<case>/sanitizer_report.json
```

and the publish job copies **only that JSON** (`main()` in `gen_sanitizer_dashboard.py` does a single `shutil.copy2` of `sanitizer_report.json`; the guardrail loop in `.github/workflows/sanitizers-nightly.yml` does the same per case).

Everything else the sweep produced is discarded:

- `consan/consan.log` — written by `run_consan()` (`src/aorta/instrumentation/rocjitsu_sanitizers/consan.py`)
- `waitcheck/waitcheck-<n>.log` — written by `run_waitcheck()` (`.../waitcheck.py`)

These live only in the 30-day Actions artifact (`sanitizer-nightly-<run_id>`) and are then deleted. There is also no copy of the recipe that ran, no record of the fixture build steps, no toolchain/rocjitsu provenance, and no landing page for a case directory.

Net effect: **we tell the user to reproduce the run, but give them nothing to reproduce it against.** Sharing the command without the inputs and the observed logs does not serve the stated purpose.

Additionally, `runs/<id>/index.html` (`build_run_index_html()`) lists only the three **guardrail** recipes, so the top-nav "raw reports" link never mentions survey cases at all.

## Agreed direction

Keep Tab 2's cards exactly as they are — **no new detail blocks on the dashboard page**. The only page-level change is that the card-footer link is renamed and points at a **run area directory** instead of a single JSON file. The substance of the work is making that directory self-sufficient.

## Requirements

**A. Publish the whole case directory, not one file.** Copy the sweep's per-case output tree (`sanitizer_report.json` plus `waitcheck/` and `consan/` logs) into the published run area, in both the nightly publish job and `gen_sanitizer_dashboard.py`'s co-publish path for `--informational-results-dir`. `pages.yml` already copies `dashboard/*` recursively into `_site/sanitizers/`, so no Pages change is needed.

**B. Pin the recipe that ran.** Write a `recipe.yaml` copy into each case directory, resolved through the existing `_survey_recipe_for()` map, so the run area stays accurate even after the recipe changes on main.

**C. Record how to rebuild the inputs.** Add a `REPRODUCE.md` carrying: the commit SHA, the `aorta sweep run` command, the fixture build lines the nightly used (`hipcc --genco` / `clang-offload-bundler` / `prepare_gemm_isa.py`), the env the run needed (`ROCJITSU_PREBUILT`; for ConSan `HSA_TOOLS_LIB` / `HSA_TOOLS_DISABLE_REGISTER` / `RJ_CONSAN_MODE` / `RJ_CONSAN_POLICY`), the rocjitsu prebuilt bundle commit + run URL from its `MANIFEST.json`, and the code-object SHA-256 so a locally rebuilt object can be verified.

Note: useful provenance is already in the report but never surfaced anywhere. For ConSan, `checks[].backend` carries `command`, `command_sha256`, `hook`, `hook_sha256`, `selected_kernel`, `selected_identity_sha256`; for waitcheck it carries the `rj_waitcheck` `path` + `sha256`. (`summarize_case()` only looks for `path`/`sha256`, which is why ConSan cards render an em dash for **Backend**.)

**D. Landing page per case.** Write `runs/<id>/survey/<case>/index.html` listing every published file as a download link, plus the reproduce command, commit, observed verdict/reason, and code-object SHA-256. Trailing-slash link so Pages serves it (GitHub Pages does not auto-index directories).

**E. Retarget + relabel the footer link.** In `_survey_case_html`, point the footer at the case directory and rename it (proposed: **run area**, the term already used in the code comments and tests for `runs/<id>/`). The per-kernel-row `Report` column keeps pointing at the JSON. Directory paths must be derived from `report_rel` and re-validated through the existing `_safe_report_rel()`, so an unsafe or absent `report_rel` still emits no link.

**F. Surface survey cases on the whole-run page.** Extend `build_run_index_html()` with a **Workload survey** table so `runs/<id>/` is no longer guardrail-only.

## Acceptance criteria

- Tab 2 footer renders the new label and an `href` ending `/survey/<case>/`; **Tab 1 carries the same run-area link and label** (per decision 7 below, which supersedes this criterion's original "Tab 1 still renders view raw report"); per-row `report` links still resolve to `sanitizer_report.json`.
- A `consan/consan.log` present in the sweep output is reachable at `/sanitizers/runs/<id>/survey/<case>/consan/consan.log.gz` (gzipped, per decision 6 below, which supersedes this criterion's original uncompressed path).
- Each published case directory contains `index.html`, `sanitizer_report.json`, `recipe.yaml`, `REPRODUCE.md`, and whatever logs the case produced; `index.html` links all of them.
- `build_run_index_html()` lists survey cases and links into their run areas.
- An unsafe/absent `report_rel` emits no footer link (no dead or off-origin href).
- No customer/NDA identifiers (CLAUDE.md rule #4) — public GEMM/LDS/vecadd fixtures only.
- Tab 2 stays observed-only / non-gating; the gate is untouched.

## Decisions (were open questions — all resolved)

These are the answers the implementation follows. Where one supersedes an
acceptance criterion above, that criterion is annotated inline.

| # | Question | Decision |
|---|---|---|
| 1 | Footer link label | **`run area`** |
| 2 | Link target | the **case directory**, trailing slash |
| 3 | Fixture code objects | **publish none**; record path + SHA-256 and rebuild steps |
| 4 | One-click download | **individual file links**, no tarball |
| 5 | Log retention | **newest 7 runs** (`--keep-logs 7`); reports stay for the full 30 |
| 6 | Log size cap | **gzip** each log rather than truncate |
| 7 | Guardrail cases (Tab 1) | **same run-area link and label**, not just logs |
| 8 | Environment snapshot | also a machine-readable **`env.json`** |
| 9 | `REPRODUCE.md` placement | downloadable file **and** rendered into `index.html` |
| 10 | Transitive recipe inputs | **copy every referenced** source-level input |

**How 3 and 10 are reconciled.** They conflict, because the survey recipes
reference exactly the artifacts 3 excludes (`code_object: fixtures/isa/…`,
`consan_command: fixtures/bin/…`, both gitignored because CI builds them). 10 is
honoured for every *source-level* input — cheap, since `gemm_shapes_unique.csv`
is under 5KB and the `.hip` sources are hundreds of bytes to a few KB — while 3
wins for the two *built* kinds, which are recorded by path + SHA-256 with rebuild
instructions instead.

## Out of scope

- Serving the sanitizer backends themselves (`rj_waitcheck`, the ConSan hook) from Pages.
- Changing the reproduce command strings or the Tab 1 card layout.
- Adding new detail blocks to the dashboard page itself.

amd-vivekag:
Closes #384.

## Problem

Tab 2 prints a copy-paste reproduce command, but the card footer ("view raw
report") resolved to a single `sanitizer_report.json`, and the publish job copied
only that file. The sanitizer logs the verdict came from were written by the
sweep, uploaded to the 30-day Actions artifact, and then deleted. So the command
had nothing to reproduce against.

## What this does

Renames that footer to **run area** on both tabs, points it at the case
*directory*, and fills that directory with what a reproduction needs:

| File | What it is |
|---|---|
| `index.html` | Landing page: command, run identity, observed verdict, and every file as a download. GitHub Pages does not auto-index a directory, so without this the new link 404s. |
| `sanitizer_report.json` | Unchanged; still the per-kernel-row `Report` target. |
| `consan/consan.log.gz`, `waitcheck/waitcheck-*.log.gz` | The output the verdict was derived from. Gzipped with `mtime=0`, so an unchanged log is byte-identical on re-publish and does not churn the data branch. |
| `recipe.yaml` | The recipe as it ran. |
| `inputs/` | The recipe's source-level fixture inputs (a few KB). |
| `REPRODUCE.md` / `env.json` | Commit, command, required env, rebuild steps, digests — prose and machine-readable (`aorta.sanitizer_run_area/0.1`). |

No new detail blocks were added to the dashboard page itself, per the issue's
"Agreed direction".

The digests were already in every report and rendered nowhere: waitcheck's binary
`path`/`sha256`, ConSan's `command`/`command_sha256`/`hook`/`hook_sha256`, and each
kernel's `code_object_sha256`. `report_digests()` flattens them and
`artifact_digest_index()` indexes them by basename, because the two sanitizers name
the same kind of artifact under different keys.

## Points a reviewer will want flagged

- **Decisions 3 vs 10 conflict** (publish no built artifacts / copy every
  referenced input). Resolved in the issue body: source-level inputs are copied,
  the two CI-built kinds are recorded by path + SHA-256. The single switch is
  `_is_built_fixture()` plus `_BUILT_FIXTURE_DIRS`.
- **Provenance is scoped to the run that produced it.** All 30 retained areas are
  re-rendered nightly, so the container image, rocjitsu bundle and recipe copy are
  sourced only for the run the job just staged; older areas keep what they captured
  while they were current. Otherwise a three-week-old area would claim today's
  image beside an `env.json` naming its own commit. `build_case_env` therefore
  takes `provenance` as a parameter instead of reading `os.environ`, which also
  makes it pure.
- **`summary.md` stays link-free on purpose.** The job-summary Markdown twin still
  describes only the reproduce command: relative Pages links do not resolve inside
  an Actions summary, the same precedent as #368.
- **`summarize_case()` still ignores ConSan's backend keys**, so Tab 2 ConSan cards
  show Backend as an em dash even though the digests exist. Deliberately out of
  scope here — a small, separate fix.
- **`artifacts_not_published[].sha256` can still be `null`** when the report
  genuinely carries no digest (a bare `isa_dir:` directory reference, or an empty
  backend block). It renders as an em dash.
- **The script stays stdlib-only.** The publish job runs it from a bare checkout
  with no `pip install`, so PyYAML is not importable — which is why recipe scanning
  is a regex over `fixtures/...` paths rather than a YAML parse.

## Workflow changes

1. The GPU job writes `${SANITIZER_OUT}/rocjitsu.json` from the prebuilt bundle's
   `MANIFEST.json`, so the publish job (a different runner that never sees the
   bundle) can record which sanitizer build produced the reports.
2. The publish job copies whole case directories rather than one file per case.
3. It exports `AORTA_ROCJITSU_COMMIT`, `AORTA_ROCJITSU_RUN_URL` and
   `AORTA_CI_IMAGE` (the digest-pinned `FROM` of `docker/Dockerfile.ci-gpu`).
4. The generator is invoked with `--keep-logs 7`.

`pages.yml` needs no change: it already copies `dashboard/*` recursively.

## Retention

`--keep-logs N` (default 7) bounds logs, the recipe copy and inputs for guardrail
and survey areas alike, while reports stay for the full `--keep 30`. A pruned area
is re-rendered from its own manifest so its file table stops linking logs that are
gone, and both renderings then describe it as pruned rather than promising files it
no longer has.

## Test plan

- [x] `pytest tests/sanitizers/test_dashboard.py` — 98 pass (10 new)
- [x] `ruff check` clean on both changed Python files
- [x] `.github/workflows/sanitizers-nightly.yml` parses as YAML
- [x] End-to-end publish of a 4-run history with a survey case: 137 relative links
      all resolve, including from pruned areas
- [x] Log retention verified across *sequential* publishes into one tree (the
      data-branch pattern) — a single publish cannot show it
- [ ] **Not yet run in real CI.** The first nightly after merge is the real test,
      in particular the `cp -a` of case dirs and the `rocjitsu.json` hop between
      jobs. `.log.gz` behaviour over the Pages CDN is also unverified — one `curl`
      after the first deploy is worth it.
- [ ] Real ConSan log sizes are unmeasured (only toy fixtures were gzipped
      locally). `--keep-logs 7` bounds growth; if a real log turns out large,
      decision 6's alternative (a size cap with a truncation note) is the fallback.

### DIFF
--- a/scripts/sanitizers/gen_sanitizer_dashboard.py
+++ b/scripts/sanitizers/gen_sanitizer_dashboard.py
@@ -898,7 +898,10 @@ def _load_survey_report(report_path: str, base_dir: Path | None) -> dict[str, An
 
 
 def survey_cases_from_spec(
-    spec: dict[str, Any] | list[Any], *, base_dir: Path | None = None
+    spec: dict[str, Any] | list[Any],
+    *,
+    base_dir: Path | None = None,
+    rel: str | None = None,
 ) -> list[dict[str, Any]]:
     """Build ordered observed-only *survey* case entries from a spec (pure-ish).
 
@@ -991,17 +994,24 @@ def survey_cases_from_spec(
                 # only a name that is one plain path segment can be claimed, since
                 # the area link is built from it and `..` would traverse.
                 #
-                # The name must also *agree* with report_rel. The two renderers use
-                # different sources -- the run page builds survey/<name>/ while the
-                # card footer takes the directory of report_rel -- so a spec naming
-                # "foo" with a report under survey/bar/ advertised two different run
-                # areas, one of which is wrong. Validating them independently was
-                # not enough; they have to describe the same directory.
+                # The two renderers derive this link from different sources -- the
+                # run page builds ``survey/<name>/`` relative to the run it is
+                # rendering, the card footer takes the directory of ``report_rel``
+                # -- so they can only be trusted if the *whole* area is pinned to
+                # one run. Comparing only the case segment was not enough: a report
+                # under ``runs/other/survey/foo/`` matched ``name="foo"`` while the
+                # footer pointed into a different run. Require the complete path
+                # for the run being rendered, which is exactly what the
+                # informational builder constructs, so the two agree by
+                # construction. Without a run context (``rel=None``) nothing can be
+                # verified and no spec entry may claim an area.
                 "staged": (
                     case.get("staged") is True
                     and summary["present"]
                     and _safe_case_segment(name) is not None
-                    and _report_rel_case_segment(case.get("report_rel")) == name
+                    and bool(rel)
+                    and _safe_report_rel(case.get("report_rel"))
+                    == f"{rel}/survey/{name}/sanitizer_report.json"
                 ),
                 "cls": "survey",
                 "summary": summary,
@@ -1040,6 +1050,17 @@ def _split_survey_case(name: str) -> tuple[str, str | None]:
         if name.startswith(prefix):
             return name[len(prefix):], san
     return name, None
+
+
+def _published_command(case_dir: Path) -> str:
+    """The reproduce command a published run area recorded, or "" (no manifest).
+
+    A fresh sweep's results dir has no ``env.json``, so this is empty there and the
+    caller falls back to the recipe map.
+    """
+    env = _load(case_dir / "env.json")
+    command = env.get("command") if isinstance(env, dict) else None
+    return command if isinstance(command, str) and command else ""
 
 
 def survey_cases_from_informational_dir(
@@ -1109,7 +1130,15 @@ def survey_cases_from_informational_dir(
                 "sanitizer": sanitizer,
                 "backend": str(backend_name),
                 "workload": None,
-                "command": (
+                # An already-published area records the command it actually ran, so
+                # prefer it over recomputing from the current recipe map. This
+                # function also reconstructs a published run's survey (see
+                # ``displayed_survey`` in ``main``), and recomputing there let a
+                # recipe rename rewrite a historical run's reproduce command --
+                # the same "prose must not outlive its manifest" rule the run-area
+                # renderers follow.
+                "command": _published_command(case_dir)
+                or (
                     "aorta sweep run --recipe "
                     f"recipes/sanitizers/{_survey_recipe_for(name)}.yaml"
                 ),
@@ -1213,21 +1242,6 @@ def _safe_case_segment(name: Any) -> str | None:
     if not isinstance(name, str) or not _CASE_SEGMENT_RE.fullmatch(name):
         return None
     return name
-
-
-def _report_rel_case_segment(report_rel: Any) -> str | None:
-    """The case-directory name a ``report_rel`` points at, validated (pure).
-
-    Used to cross-check the two sources the renderers use for one area: the run
-    page builds ``survey/<name>/`` from the case name while the card footer takes
-    the directory of ``report_rel``. Validating each on its own still allowed a
-    spec to name ``foo`` while its report sat under ``survey/bar/``, advertising two
-    different directories for one case.
-    """
-    area = _case_dir_rel(_safe_report_rel(report_rel))
-    if not area:
-        return None
-    return _safe_case_segment(area.rstrip("/").rsplit("/", 1)[-1])
 
 
 def _case_dir_rel(report_rel: str | None) -> str | None:
@@ -3826,7 +3840,9 @@ def main() -> int:
     if args.survey is not None:
         spec = _load(args.survey)
         if isinstance(spec, (dict, list)):
-            survey = survey_cases_from_spec(spec, base_dir=args.survey.parent)
+            survey = survey_cases_from_spec(
+                spec, base_dir=args.survey.parent, rel=current_rel
+            )
     # Caller-supplied ConSan cases (#347) now render as observed-only workload-survey
     # (Tab 2) entries rather than a separate informational section. They are appended
     # after any explicit --survey spec cases so an explicit spec (if wired) leads, and
--- a/scripts/sanitizers/gen_sanitizer_dashboard.py
+++ b/scripts/sanitizers/gen_sanitizer_dashboard.py
@@ -1406,6 +1406,12 @@ def _publish_recipe_inputs(
         return [], []
     (case_dir / "recipe.yaml").write_text(recipe_text, encoding="utf-8")
     source_refs, built_refs = _recipe_fixture_refs(recipe_text)
+    # Plus the sources the rebuild commands read. A recipe names only the built
+    # artifact it consumes -- every daily-consan-* recipe lists a .hsaco and a
+    # loader binary and nothing else -- so scanning the recipe text alone left
+    # inputs/ empty for exactly the cases whose published commands read the most
+    # (decision 10 asks for every input a reproduction needs).
+    source_refs = list(dict.fromkeys(source_refs + rebuild_input_sources(built_refs)))
     recipe_dir = recipe_src.parent
     copied: list[str] = []
     inputs_root = (case_dir / "inputs").resolve()
@@ -1642,10 +1648,42 @@ def _runnable(ref: str) -> str:
     return f"{_FIXTURES_FROM_ROOT}{ref[len('fixtures'):]}" if ref.startswith("fixtures") else ref
 
 
+_GEMM_ISA_CSV = "fixtures/gemm_shapes_unique.csv"
 _GEMM_ISA_SCRIPT = (
     "python scripts/sanitizers/prepare_gemm_isa.py "
-    f"--csv {_FIXTURES_FROM_ROOT}/gemm_shapes_unique.csv --out {_FIXTURES_FROM_ROOT}/isa"
+    f"--csv {_runnable(_GEMM_ISA_CSV)} --out {_FIXTURES_FROM_ROOT}/isa"
 )
+
+
+def rebuild_input_sources(built_refs: list[str]) -> list[str]:
+    """Recipe-relative source files the rebuild commands for these artifacts read.
+
+    Pure, and derived from the same tables ``rebuild_plan`` renders from, so the
+    inputs a run area copies cannot disagree with the commands it publishes.
+
+    Needed because a recipe names only the built artifact it *consumes*: every
+    ``daily-consan-*`` recipe lists a ``.hsaco`` and a loader binary and nothing
+    else, so scanning the recipe text alone left ``inputs/`` empty for exactly the
+    cases whose published commands read the most. A binary's loaded code object is
+    folded in too, since rebuilding the binary means rebuilding that object first
+    -- ``consan_gemm_load`` needs ``consan_load.hip`` *and* the GEMM shape CSV its
+    object is extracted with.
+    """
+    refs = list(built_refs) + [
+        _BIN_OBJECT[ref] for ref in built_refs if ref in _BIN_OBJECT
+    ]
+    sources: list[str] = []
+    for ref in dict.fromkeys(refs):
+        genco = _GENCO_ISA_SOURCES.get(ref)
+        if genco:
+            sources.append(genco)
+        elif f"{ref}/".startswith("fixtures/isa/") or ref == "fixtures/isa":
+            # Every isa/ artifact not built by --genco is extracted by
+            # prepare_gemm_isa.py, which reads the shape CSV.
+            sources.append(_GEMM_ISA_CSV)
+        if ref in _BIN_SOURCES:
+            sources.append(_BIN_SOURCES[ref][0])
+    return list(dict.fromkeys(sources))
 
 # Every rebuild starts here, because the tool none of them can run without is not
 # on PATH in the image that built the artifact: the ROCm container exports only
--- a/scripts/sanitizers/gen_sanitizer_dashboard.py
+++ b/scripts/sanitizers/gen_sanitizer_dashboard.py
@@ -971,6 +971,13 @@ def survey_cases_from_spec(
                 "report_rel": (
                     _safe_report_rel(case.get("report_rel")) if summary["present"] else None
                 ),
+                # Whether this case's whole run-area *directory* is published, not
+                # just its report. A spec entry renders from an inline report or an
+                # out-of-tree report_path, so by default nothing stages the
+                # directory and no run-area link may be emitted for it (a present
+                # report is not evidence the directory exists). A caller that does
+                # publish complete areas can say so with `"staged": true`.
+                "staged": case.get("staged") is True and summary["present"],
                 "cls": "survey",
                 "summary": summary,
             }
@@ -1082,6 +1089,13 @@ def survey_cases_from_informational_dir(
                     f"recipes/sanitizers/{_survey_recipe_for(name)}.yaml"
                 ),
                 "report_rel": report_rel,
+                # These are the cases ``main`` co-publishes as complete run areas
+                # (report + logs + manifests + index.html) under
+                # ``<rel>/survey/<name>/``, so unlike a ``--survey`` spec entry the
+                # directory link is safe to emit. Gated on ``rel`` for the same
+                # reason ``report_rel`` is: the results-dir / runs-root modes
+                # publish nothing co-located.
+                "staged": rel is not None and summary["present"],
                 "cls": "survey",
                 "summary": summary,
             }
@@ -1374,25 +1388,116 @@ _REBUILD_NOTE = (
     ".github/workflows/sanitizers-nightly.yml."
 )
 
+# The env a reproduction needs, shared by both renderings so the page and the
+# downloadable file cannot drift (#384 decision 9 renders REPRODUCE.md into the
+# landing page, so its reproduction details have to be on both).
+_REQUIRED_ENV_NOTE = (
+    "The sanitizer backends (`rj_waitcheck`, the ConSan hook) come from the "
+    "prebuilt rocjitsu bundle; point `ROCJITSU_PREBUILT` at an unpacked bundle "
+    "before running. ConSan additionally requires `HSA_TOOLS_LIB`, "
+    "`HSA_TOOLS_DISABLE_REGISTER=1`, `RJ_CONSAN_MODE` and `RJ_CONSAN_POLICY`, "
+    "which `aorta` sets for you from the recipe's policy block."
+)
+
+
+def _md_code_to_html(text: str) -> str:
+    """Render a shared Markdown prose constant's ``backticks`` as ``<code>``.
+
+    Lets the page and REPRODUCE.md render the same constant natively rather than
+    keeping two hand-synced copies of the same sentence. Escaping happens first,
+    so the split can never introduce markup from the text itself.
+    """
+    parts = _esc(text).split("`")
+    return "".join(
+        part if index % 2 == 0 else f"<code>{part}</code>"
+        for index, part in enumerate(parts)
+    )
+
+# Which source each unpublished fixture is built from, and how. Paths are
+# recipe-relative, matching how a recipe names them.
+#
+# These are per-artifact rather than one hint per directory because the commands
+# are NOT interchangeable, and a generic "hipcc it" hint produces a file whose
+# digest does not match the one recorded beside it:
+#
+# * an ``--genco`` object is a raw code object on some ROCm builds and a
+#   clang-offload bundle on others, so it must be unbundled conditionally -- the
+#   recorded digest is of the unbundled object the loader opens;
+# * the GEMM objects are *extracted* from the shipped Tensile libraries by
+#   prepare_gemm_isa.py, never compiled from a .hip source;
+# * every consan_load / lds_dispatch binary is one source compiled with a define
+#   naming the code object it loads, so omitting the define builds the wrong
+#   binary rather than failing loudly.
+_GENCO_ISA_SOURCES = {
+    "fixtures/isa/lds.hsaco": "fixtures/kernels/lds_reduce.hip",
+    "fixtures/isa/tiny.hsaco": "fixtures/kernels/tiny_vecadd.hip",
+}
+# ref -> (source, define name or None). A define binds the binary to its object.
+_BIN_SOURCES: dict[str, tuple[str, str | None]] = {
+    "fixtures/bin/consan_lds_race": ("fixtures/repro/consan_lds_race.hip", None),
+    "fixtures/bin/consan_lds_race_2wave": (
+        "fixtures/repro/consan_lds_race_2wave.hip",
+        None,
+    ),
+    "fixtures/bin/consan_tiny_load": ("fixtures/kernels/consan_load.hip", "OBJECT"),
+    "fixtures/bin/consan_gemm_load": ("fixtures/kernels/consan_load.hip", "OBJECT"),
+    "fixtures/bin/lds_dispatch": ("fixtures/kernels/lds_dispatch.hip", "LDS_HSACO"),
+}
+# The code object each defined binary loads, so the define can name it.
+_BIN_OBJECT = {
+    "fixtures/bin/consan_tiny_load": "fixtures/isa/tiny.hsaco",
+    "fixtures/bin/consan_gemm_load": "fixtures/isa/consan_gemm_f32.hsaco",
+    "fixtures/bin/lds_dispatch": "fixtures/isa/lds.hsaco",
+}
+_GEMM_ISA_HINT = (
+    "extracted from the shipped Tensile libraries (not compiled from a .hip "
+    "source) by `python scripts/sanitizers/prepare_gemm_isa.py --csv "
+    "fixtures/gemm_shapes_unique.csv --out fixtures/isa"
+)
+
 
 def _rebuild_hints(built_refs: list[str], *, target: str) -> list[str]:
     """One rebuild hint per artifact the run area deliberately does not ship (pure).
 
     Single source of truth so the REPRODUCE.md and landing-page renderings of
-    the rebuild section stay in parity.
+    the rebuild section stay in parity. Each hint is the command the nightly
+    actually ran (see the tables above), because a rebuild that differs from it
+    will not reproduce the SHA-256 recorded next to the artifact.
     """
     hints: list[str] = []
     for ref in built_refs:
-        if f"{ref}/".startswith("fixtures/isa/"):
+        source = _GENCO_ISA_SOURCES.get(ref)
+        if source:
             hints.append(
-                f"{ref} -- a gfx code object: `hipcc --genco --offload-arch={target}` "
-                "from the matching fixtures/kernels/*.hip source, or "
-                "scripts/sanitizers/prepare_gemm_isa.py for the GEMM objects."
+                f"{ref} -- a gfx code object: `hipcc --genco --offload-arch={target} "
+                f"{source} -o tmp.o`, then unbundle only if the result is a bundle "
+                "(`head -c 24 tmp.o | grep -qF __CLANG_OFFLOAD_BUNDLE__`) with "
+                "`clang-offload-bundler --type=o --unbundle --input=tmp.o "
+                f"--targets=hipv4-amdgcn-amd-amdhsa--{target} --output={ref}`, "
+                "otherwise copy tmp.o to it. The recorded digest is of the "
+                "unbundled object, so skipping that check will not match."
             )
-        elif f"{ref}/".startswith("fixtures/bin/"):
+        elif ref == "fixtures/isa/consan_gemm_f32.hsaco":
+            hints.append(f"{ref} -- {_GEMM_ISA_HINT} --top-n 0 --consan-object {ref}`.")
+        elif f"{ref}/".startswith("fixtures/isa/") or ref == "fixtures/isa":
+            # sol_<n>.hsaco, or a bare isa_dir reference covering them.
             hints.append(
-                f"{ref} -- a host repro binary: "
-                f"`hipcc --offload-arch={target} <source>.hip -o {ref}`."
+                f"{ref} -- per-transpose f32 GEMM code objects (sol_<n>.hsaco), "
+                f"{_GEMM_ISA_HINT} --top-n 3`."
+            )
+        elif ref in _BIN_SOURCES:
+            source, define = _BIN_SOURCES[ref]
+            flags = f"--offload-arch={target}"
+            if define:
+                flags += f' -D{define}="\\"$(pwd)/{_BIN_OBJECT[ref]}\\""'
+            hints.append(
+                f"{ref} -- a host repro binary: `hipcc {flags} {source} -o {ref}`."
+                + (
+                    ""
+                    if not define
+                    else f" The -D{define} define is required: it is how the binary "
+                    "learns which code object to load."
+                )
             )
         else:
             hints.append(ref)
@@ -1527,11 +1632,7 @@ def build_reproduce_md(env: dict[str, Any], *, built_refs: list[str]) -> str:
         f"{env.get('command', '')}",
         "```",
         "",
-        "The sanitizer backends (`rj_waitcheck`, the ConSan hook) come from the "
-        "prebuilt rocjitsu bundle; point `ROCJITSU_PREBUILT` at an unpacked bundle "
-        "before running. ConSan additionally requires `HSA_TOOLS_LIB`, "
-        "`HSA_TOOLS_DISABLE_REGISTER=1`, `RJ_CONSAN_MODE` and `RJ_CONSAN_POLICY`, "
-        "which `aorta` sets for you from the recipe's policy block.",
+        _REQUIRED_ENV_NOTE,
         "",
     ]
     hints = _rebuild_hints(built_refs, target=str(env.get("target") or "gfx950"))
@@ -1620,6 +1721,7 @@ def build_case_index_html(
         ("Date", _esc(str(env.get("date", "")))),
         ("Target", _esc(str(env.get("target", "")))),
         ("Recipe", _esc(str(env.get("recipe", "")))),
+        ("Execution", _esc(str(observed.get("execution") or _DASH))),
         ("Findings", _esc(str(observed.get("findings", 0)))),
     ]
     if env.get("container_image"):
@@ -1638,7 +1740,9 @@ def build_case_index_html(
     steps_html = (
         (
             '<p class="cap">Rebuild the fixtures</p><ul class="steps">'
-            + "".join(f"<li>{_esc(hint)}</li>" for hint in hints)
+            # The hints are one shared Markdown constant per artifact, so their
+            # `backticks` become <code> here rather than rendering literally.
+            + "".join(f"<li>{_md_code_to_html(hint)}</li>" for hint in hints)
             + f'</ul><p class="note">{_esc(_REBUILD_NOTE)}</p>'
         )
         if hints
@@ -1656,6 +1760,23 @@ def build_case_index_html(
             '<p class="cap">Artifacts not published</p><div class="table-wrap"><table>'
             "<thead><tr><th>Path</th><th>SHA-256</th></tr></thead>"
             f"<tbody>{np_rows}</tbody></table></div>"
+        )
+    # #384 requires the code-object SHA-256 on this page, and the artifacts table
+    # above cannot carry it on its own: a recipe that names only a bare
+    # ``isa_dir: fixtures/isa`` has one row and no per-object digest, so without
+    # this section those digests would exist only in env.json / REPRODUCE.md.
+    digests_html = ""
+    digests = env.get("digests") or {}
+    if isinstance(digests, dict) and digests:
+        digest_rows = "".join(
+            f"<tr><td class=mono>{_esc(str(key))}</td>"
+            f'<td class="mono wrap-any">{_esc(str(value))}</td></tr>'
+            for key, value in sorted(digests.items())
+        )
+        digests_html = (
+            '<p class="cap">Recorded digests</p><div class="table-wrap"><table>'
+            "<thead><tr><th>Key</th><th>Value</th></tr></thead>"
+            f"<tbody>{digest_rows}</tbody></table></div>"
         )
     run_link = (
         f'<a href="{_esc(str(env["run_url"]))}">workflow run</a>'
@@ -1695,7 +1816,8 @@ def build_case_index_html(
         f"{reason_html}\n"
         '<div class="repro"><span class="lbl">Reproduce</span>'
         f'<code>{_esc(str(env.get("command", "")))}</code></div>\n'
-        f"{steps_html}{np_html}"
+        f'<p class="note">{_md_code_to_html(_REQUIRED_ENV_NOTE)}</p>\n'
+        f"{steps_html}{np_html}{digests_html}"
         "</section>\n"
         '<section class="panel"><h2>Files</h2>\n'
         f'<p class="cap">{_esc(files_caption)}</p>\n'
@@ -2128,7 +2250,7 @@ def _kernel_tables_html(
     )
 
 
-def _raw_link_html(report_rel: str | None) -> str:
+def _raw_link_html(report_rel: str | None, *, staged: bool = True) -> str:
     """The run-area link, as a card footer.
 
     Points at the case's published *directory*, not just its
@@ -2138,10 +2260,17 @@ def _raw_link_html(report_rel: str | None) -> str:
     (#384). The per-kernel-row ``Report`` column still links the JSON directly
     for a schema drill-down.
 
+    ``staged=False`` emits nothing. A directory link needs an ``index.html`` that
+    only the publisher of the area can write, so a case whose *report* was staged
+    by someone else (a ``--survey`` spec supplying an inline report) has a
+    reachable JSON but no reachable directory. Guardrail rows default to ``True``:
+    their ``report_rel`` is generator-computed for present cases, and the same
+    loop publishes the area it names.
+
     Deliberately in the card body rather than the ``<summary>`` row: a link
     inside a summary competes with the disclosure toggle.
     """
-    rel = _case_dir_rel(report_rel)
+    rel = _case_dir_rel(report_rel) if staged else None
     if not rel:
         return ""
     return f'<div class="card-foot"><a href="{_esc(rel)}">run area</a></div>'
@@ -2671,7 +2800,7 @@ def _survey_case_html(entry: dict[str, Any], *, heading: str, kind: str = "") ->
             + _survey_message_html(row)
             + _survey_howto_html(entry)
             + _kernel_tables_html(row, report_rel=entry.get("report_rel"))
-            + _raw_link_html(entry.get("report_rel"))
+            + _raw_link_html(entry.get("report_rel"), staged=bool(entry.get("staged")))
         )
     return (
         f'<details class="kcard {accent}">{summary}'
@@ -2867,10 +2996,17 @@ def _run_index_survey_html(survey: list[dict[str, Any]] | None) -> str:
 
     Without this the page lists only the three gated guardrail recipes, so the
     top-nav "raw reports" link never mentions the survey cases published
-    alongside them under ``survey/<case>/`` (#384). Links are case-local, and
-    only a present report gets one so the page never points at a 404. For a past
-    run the entries come from ``survey_entries_from_published``, whose recorded
-    verdict can be absent -- hence the em-dash fallback rather than a "None".
+    alongside them under ``survey/<case>/`` (#384). For a past run the entries
+    come from ``survey_entries_from_published``, whose recorded verdict can be
+    absent -- hence the em-dash fallback rather than a "None".
+
+    A run-area link is emitted only for an entry marked ``staged`` -- i.e. one
+    whose whole directory was published -- never from the case name plus report
+    presence. A report can be *present* (it parsed) without its directory
+    existing at all: a ``--survey`` spec's entries render from an inline report
+    or an out-of-tree ``report_path``, while only ``--informational-results-dir``
+    entries are staged under ``survey/<case>/``. Keying on presence therefore
+    emitted a 404 on exactly that invocation.
     """
     if not survey:
         return ""
@@ -2880,7 +3016,7 @@ def _run_index_survey_html(survey: list[dict[str, Any]] | None) -> str:
         f"<td class=mono>{_esc(str(entry.get('command') or ''))}</td>"
         + (
             f'<td><a href="survey/{_esc(str(entry.get("name", "")))}/">run area</a></td>'
-            if (entry.get("summary") or {}).get("present")
+            if entry.get("staged")
             else '<td><span class="dash">&mdash;</span></td>'
         )
         + "</tr>"
@@ -2904,6 +3040,9 @@ def survey_entries_from_published(run_dir: Path) -> list[dict[str, Any]]:
     from -- leaving the survey areas still on the data branch under it reachable
     only by typing the URL, on a site that cannot auto-index a directory. Each
     area's own ``env.json`` already records everything that table shows.
+
+    Every entry is ``staged``: these are read off disk, so the directory is known
+    to exist by construction.
     """
     entries: list[dict[str, Any]] = []
     for case_dir in sorted(p for p in (run_dir / "survey").glob("*") if p.is_dir()):
@@ -2915,6 +3054,7 @@ def survey_entries_from_published(run_dir: Path) -> list[dict[str, Any]]:
             "name": case_dir.name,
             "label": str(env.get("case") or case_dir.name),
             "command": str(env.get("command") or ""),
+            "staged": True,
             "summary": {
                 "verdict": observed.get("verdict"),
                 "present": (case_dir / "sanitizer_report.json").is_file(),
@@ -3327,6 +3467,7 @@ def main() -> int:
         # Describes this job's GPU run only, so it is stamped onto the run being
         # published now and never onto the older runs re-rendered beside it.
         provenance = provenance_from_environ()
+        info_dir = args.informational_results_dir
         runs_out = args.out_dir / "runs"
         runs_out_res = runs_out.resolve()
         history_res = args.history_root.resolve()
@@ -3450,7 +3591,6 @@ def main() -> int:
         # that (re)creates out_dir/<rel>, using the same rel + case name as
         # survey_cases_from_informational_dir so the copied path matches the link.
         latest_rel = runs[0].get("rel") if runs else None
-        info_dir = args.informational_results_dir
         if latest_rel and info_dir is not None and info_dir.is_dir():
             survey_by_name = {str(entry.get("name")): entry for entry in survey}
             for case_dir in sorted(p for p in info_dir.iterdir() if p.is_dir()):
--- a/scripts/sanitizers/gen_sanitizer_dashboard.py
+++ b/scripts/sanitizers/gen_sanitizer_dashboard.py
@@ -1187,10 +1187,21 @@ def _recipe_fixture_refs(recipe_text: str) -> tuple[list[str], list[str]]:
     Source inputs (the ``.hip`` repro sources, the GEMM shape CSV) are small --
     hundreds of bytes to a few KB -- and are copied into the run area. Built
     artifacts are recorded by digest only. Order is first-seen and de-duplicated.
+
+    References containing a ``..`` segment are dropped here, at the one boundary
+    both consumers share. ``_FIXTURE_REF_RE`` admits ``.`` so it can match a file
+    extension, which also let ``fixtures/../../../README.md`` through -- and while
+    the copy validated that the *source* stayed inside the repo, it then joined the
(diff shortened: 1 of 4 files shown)


amd-vivekag:
Thanks for the review — all three suppressed comments were correct, and I verified each against the code before acting. Fixed in `f4bbd13`.

**1. Rebuild hints were not the nightly's invocation** (`_rebuild_hints`)

Confirmed, and worse than reported once I swept the sibling branch. The hints are now per-artifact, keyed by the path a recipe names, because the commands turn out not to be interchangeable:

- **`--genco` objects** (`lds.hsaco`, `tiny.hsaco`): the hint now carries the source *and* the conditional unbundle, since `--genco` emits a raw ELF on some ROCm builds and a clang-offload bundle on others, and the recorded digest is of the *unbundled* object the loader opens. Following the old hint could produce the bundle and fail the digest check, exactly as you said.
- **GEMM objects** (`consan_gemm_f32.hsaco`, `sol_<n>.hsaco`): these are *extracted* from the shipped Tensile libraries by `prepare_gemm_isa.py`, never compiled from a `.hip` source, so the old hint pointed at the wrong tool entirely.
- **`fixtures/bin/` binaries** (the sweep finding you didn't quote): the old hint was `hipcc --offload-arch=<target> <source>.hip -o <ref>`, which omitted the `-DOBJECT` / `-DLDS_HSACO` define. That define is how `consan_load.hip` / `lds_dispatch.hip` learn which code object to load, so the old hint would have built a *working but wrong* binary rather than failing loudly.

On "ideally from one source shared with the workflow": I couldn't make that literal, since the workflow side is bash inside a YAML block scalar. Instead `test_rebuild_hint_sources_all_exist_in_the_repo` asserts every source path a hint names still exists, so a rename breaks the test rather than silently shipping a stale command, and `_REBUILD_NOTE` still points at the workflow as the source of truth.

**2. The landing page never rendered `env["digests"]`**

Confirmed, including your specific case: the waitcheck GEMM recipe references only a bare `isa_dir: fixtures/isa`, so the artifacts table has one row with no per-object digest and the `code_object_sha256` values were reachable only in `env.json` / `REPRODUCE.md`. That misses #384 requirement D.

The page now has a **Recorded digests** section. Sweeping the rest of the HTML/MD pair for the same omission also turned up the required-env note and the execution status, so both are on the page now too, and the env note is rendered from the same `_REQUIRED_ENV_NOTE` constant as `REPRODUCE.md` so the two cannot drift (a small `_md_code_to_html` helper renders its Markdown backticks as `<code>`, which also removed the literal backticks the rebuild hints were leaving in the page).

**3. `summary.present` does not prove the directory was published**

Confirmed and reproduced before fixing: with `--history-root … --survey spec.json`, the run index emitted `href="survey/consan-gemm/"` for a directory that does not exist — a real 404, since only `--informational-results-dir` entries get staged.

Rather than validate the link, I took your first suggestion and built the table from verified published areas: survey entries now carry a `staged` flag set only by whoever actually materialises the area (`survey_cases_from_informational_dir`, `survey_entries_from_published`, or an explicit `"staged": true` in a spec, for a future `--survey` caller that publishes complete areas). Nothing is synthesized from the case name any more.

The sweep found the **Tab 2 card footer** had the same defect for the same reason, so it takes the same flag. Its per-kernel `Report` column still links the JSON, which is the caller's to stage and unchanged. I also checked the guardrail equivalent (`_area_cell` in `build_run_index_html`): there `present` does imply the directory exists, because the same loop that reads the report publishes the area it names — so that one is safe by construction, and `test_main_with_a_survey_spec_emits_no_run_area_link_it_cannot_publish` now walks every relative link on a published run page to keep it that way.

Verification: 104 tests pass (6 new, each confirmed to fail without its fix), ruff clean, and the end-to-end link check resolves all 137 relative links across a 4-run history in both publish modes.

amd-vivekag:
Thanks for the second pass — all six were real, and three were more serious than the first round. Each was reproduced before fixing. Fixed in `5325f19`.

**1. A rejected survey report could abort the whole publication** (`main`, co-publish loop)

This was the worst of the six. `survey_cases_from_informational_dir` deliberately skips a report whose nested shape makes the reduction raise, but the copy loop ran `copytree` *before* consulting that decision — so the directory was published with no manifest, and on the **next** nightly the retrofit path re-read that same report and raised. I reproduced it end to end: publish 1 succeeds, publish 2 dies with `TypeError: 'NoneType' object is not iterable`, taking the entire dashboard down a day later for a non-gating input.

Two fixes, because the orphan and the crash are separable:
- The entry is now resolved **before** copying, so nothing the survey loader refused to read persists into history.
- The retrofit isolates `summarize_case` in the same `except (AttributeError, KeyError, TypeError, ValueError)` the loader already uses, so an orphan *already* on the data branch degrades instead of aborting.

**2. `..` in a fixture ref wrote outside the case directory** (`_publish_recipe_inputs`)

Correct, and the mechanism is exactly as you described: the source was resolved and checked against `repo_root`, then the same unvalidated `ref` was joined onto the destination. `fixtures/../../../README.md` landed in the published *run* directory, outside the case dir entirely.

Parent segments are now rejected in `_recipe_fixture_refs` — the single boundary both consumers share, so the copied inputs and the recorded not-published list stay consistent about which refs exist — with a destination containment check as a second layer. (Worth noting the first version of the regression test passed even with both guards removed, because the fixture's `..` count didn't actually reach the planted file; it now asserts the exploit path resolves before exercising it.)

**3. Repro-binary hints omitted `-O1 -g`**

Right, and a fair hit given I'd just rewritten these. The nightly builds the two guardrail repro binaries `-O1 -g` and the loader binaries without, so flags are per-artifact now. More usefully, `test_rebuild_hint_binary_flags_match_the_workflow` joins the workflow's line continuations, finds the single `hipcc` line building each `bin/<name>`, and asserts `-O1 -g` presence matches the map — so this class of drift now fails a test rather than shipping.

**4. `--keep-logs 0` left the survey area unbounded**

Confirmed: the co-publish path always gzipped and `write_case_run_area` defaulted `logs=True`, so at the one setting that asks for nothing to be kept, the survey area was the only thing keeping its logs. The current run's retention decision is now computed once and threaded into the survey publisher too, and the boundary is tested on both classes.

**5. The disjoint-tree copy dropped `src_run/survey`**

Confirmed. The stale-clear empties the output tree and the copy only replaced the guardrail case dirs and `meta.json`, so retained runs lost their survey downloads and their regenerated indexes dropped the survey table. The subtree is copied now.

Chasing the consequence turned up a second defect you didn't quote: a copied area also needs a landing page, and the refresh loop was skipping the whole *current* run rather than just the areas the co-publish loop owns — so a survey area inherited from history got no `index.html`. The skip is now scoped to exactly the co-published names. (That fix also revealed a dedent of mine that had left the retrofit body under a `continue` as dead code; ruff and 109 tests stayed green, and only the new disjoint-tree test caught it.)

**6. `env.json` recorded neither the required env nor the rebuild steps**

Fair, and the PR body did overclaim. `env.json` now carries both as data:

- `required_env`: each variable with `set_by` (`operator` vs `aorta`, since the four ConSan ones are derived from the recipe's policy block) and `purpose`.
- `rebuild`: per-artifact `{path, what, commands, caveat}`, from a new pure `rebuild_plan()`.

Both prose renderings are **generated from** those fields — `_rebuild_hints` renders `rebuild_plan`, and a test asserts every command appears verbatim in the rendered hint — so the manifest, the page and `REPRODUCE.md` cannot disagree about how an artifact was built. The one hand-written sentence (the env note, which reads better as prose) is bound to the data by a test that checks every variable is named in it and that it names no variable the data omits.

Verification: 113 tests pass (9 new; each confirmed to fail with its fix reverted), ruff clean, workflow YAML parses, and the end-to-end link check resolves all 137 relative links across a 4-run history in both publish modes.

amd-vivekag:
Thanks for the third pass — all eight were valid, and the first three were a genuine embarrassment on my part since I'd already rewritten this section twice. Fixed in `dd3212d`.

**1–3. The rebuild commands were not runnable** (`rebuild_plan`, `_GEMM_ISA_SCRIPT`, the binary branch)

You're right on all three counts, and they're one root cause. `REPRODUCE.md` leaves the reader at the repo root after `cd aorta`, but the commands carried recipe-relative paths — and there is no top-level `fixtures/`; the tree is at `recipes/sanitizers/fixtures`, which is exactly what the nightly's own `FIXTURES` variable is set to. So every advertised command failed on its first argument. The `prepare_gemm_isa.py` line was the worst of it, mixing a repo-root-relative script path with recipe-relative `--csv`/`--out` arguments.

Three things changed:
- Commands are rewritten through a `_runnable()` helper so every fixture path is rooted at `recipes/sanitizers/fixtures/...`. The recorded `path` stays recipe-relative, since that's how a recipe names the artifact and how `artifacts_not_published` lists it.
- Each entry now begins with `mkdir -p recipes/sanitizers/fixtures/{isa,bin}`, because those directories are gitignored build outputs.
- The genco conditional is a single executable command — `if head -c 24 tmp.o | grep -qF __CLANG_OFFLOAD_BUNDLE__; then clang-offload-bundler ...; else cp tmp.o ...; fi` — rather than a `caveat` sentence. You were right that prose can't make an automated consumer branch. The `$(pwd)/...` define is rooted too, since `$(pwd)` is the repo root there.

Two new tests lock this down: one asserts no bare `fixtures/...` token survives into any command and that every non-generated path a command names exists in the checkout, the other asserts the branch is one command with a real `else cp`.

**4. `required_env` claimed ConSan's variables for waitcheck cases**

Correct, and it contradicted the prose I'd written right next to it ("ConSan *additionally* requires"). `required_env_for()` now derives the list per case from the report's own check/finding sanitizers, falling back to the case-name convention when there's no usable report, and the internal `consan_only` flag is stripped before publishing. A waitcheck area's manifest is now just `ROCJITSU_PREBUILT`.

**5. The prose was recomputed instead of read from the persisted manifest**

This is the one I'm most glad you caught — it defeats the point of the field I added last round. Every retained area is re-rendered nightly while its `env.json` is preserved, so `_rebuild_hints(built_refs)` and the fixed `_REQUIRED_ENV_NOTE` meant a future command change would rewrite two-week-old instructions and leave each area's prose contradicting the manifest sitting beside it.

Both renderings now go through `plan_from_env(env, built_refs)` and `required_env_note(env["required_env"])`, so the stored fields are authoritative and the prose is a pure view of them; recomputation is the fallback only for an area published before the fields existed. That also made the env note per-case for free. It's the same rule the container image and recipe copy already follow from round 1 — an older area keeps what *it* captured — so thanks for spotting that I'd left these two outside it.

**6. `dirs_exist_ok=True` merged stale files into the destination**

Confirmed. A log that disappeared from the current sweep survived and got listed on the landing page as evidence for the new report. The destination is now replaced, guarded against the source sitting underneath it (the `--history-root` nested-tree lesson from #354). You're also right that the workflow's `rm -rf` masks this — the generator's co-publish path is documented for reusable output trees, so it can't rely on the caller.

**7. In-window survey areas were never gzipped**

Correct, and a plain asymmetry with the guardrail branch two loops up. Now `if keep_logs: _gzip_logs(...) else: _prune_run_area_bulk(...)`, matching it exactly. The test carries a raw `*.log` in through a disjoint history and asserts it lands gzipped.

**8. The run-page link came from the case `name`, not the validated `report_rel`** (inline)

Right, and this was a hole I opened myself last round by honouring `"staged": true` from a spec. `_safe_case_segment()` now accepts a name only as one plain path segment (no `/`, `\`, `:`, `..`, whitespace or control characters), and it gates two places: `survey_cases_from_spec` refuses to mark an unsafe name as staged at all, and `_run_index_survey_html` refuses to build a link from one even if an entry arrives pre-marked. Since `report_rel` is itself built as `<rel>/survey/<name>/sanitizer_report.json` from that same value, validating the segment makes the two renderers agree by construction rather than by coincidence. The name can still appear as escaped label *text* in the Case column — that's display, not a path — and the test asserts no `href` is ever emitted for it.

Verification: 120 tests pass (13 new; each confirmed to fail with its fix reverted), ruff clean, workflow YAML parses, and the end-to-end link check resolves all 137 relative links in both publish modes.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -2471,46 +3476,150 @@ def main() -> int:
...
-                    src_report = src_run / case / "sanitizer_report.json"
-                    if src_report.is_file():
-                        (run_out / case).mkdir(parents=True, exist_ok=True)
-                        shutil.copy2(src_report, run_out / case / "sanitizer_report.json")
+                    # Copy the whole case dir, not just the report: the sanitizer
+                    # logs the verdict came from are what make the run area
+                    # reproducible (#384).
+                    src_case = src_run / case
+                    if src_case.is_dir():
+                        shutil.copytree(src_case, run_out / case, dirs_exist_ok=True)
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in `5325f19`. Confirmed: the stale-clear empties the output tree and the copy only replaced the guardrail case dirs plus `meta.json`, so retained runs lost their survey downloads and their regenerated indexes dropped the survey table. The subtree is copied now.

Chasing the consequence turned up a second defect: a copied area also needs a landing page, and the refresh loop was skipping the whole *current* run rather than just the areas the co-publish loop owns — so an area inherited from history got no `index.html`. That skip is now scoped to the co-published names. (It also surfaced a dedent of mine that had left the retrofit body under a `continue` as dead code; ruff and 109 tests stayed green, and only the new disjoint-tree test caught it.)

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -1069,13 +1089,877 @@ def survey_cases_from_informational_dir(
...
+    """
+    recipe_text = _read_text_or_empty(recipe_src)
+    if not recipe_text:
+        return [], []
+    (case_dir / "recipe.yaml").write_text(recipe_text, encoding="utf-8")
+    source_refs, built_refs = _recipe_fixture_refs(recipe_text)
+    recipe_dir = recipe_src.parent
+    copied: list[str] = []
+    for ref in source_refs:
+        src = recipe_dir / ref
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in `5325f19`. The mechanism is exactly as you described — the source was resolved and checked against `repo_root`, then the same unvalidated `ref` was joined onto the destination, so `fixtures/../../../README.md` landed in the published *run* directory, outside the case dir entirely. I reproduced that before changing anything.

Parent segments are now rejected in `_recipe_fixture_refs`, the one boundary both consumers share, so the copied inputs and the recorded not-published list stay consistent about which refs exist; the destination containment check is the second layer. Worth noting the first version of my regression test passed even with both guards removed, because the fixture's `..` count didn't actually resolve to the planted file — it now asserts the exploit path resolves before exercising it.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -2471,46 +3476,150 @@ def main() -> int:
...
                 dest = args.out_dir / latest_rel / "survey" / case_dir.name
-                dest.mkdir(parents=True, exist_ok=True)
-                shutil.copy2(src_report, dest / "sanitizer_report.json")
+                # The whole case dir, so the ConSan/waitcheck logs behind the
+                # observed verdict ship with the report (#384).
+                shutil.copytree(case_dir, dest, dirs_exist_ok=True)
+                _gzip_logs(dest)
+                entry = survey_by_name.get(case_dir.name)
+                if entry is None:
+                    continue
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in `5325f19`, and this was the most serious of that round. I reproduced it end to end: publish one succeeds, publish two dies with `TypeError: 'NoneType' object is not iterable`, taking the whole dashboard down a day later for a non-gating input.

Two fixes, since the orphan and the crash are separable. The entry is now resolved *before* copying, so nothing the survey loader refused to read persists into history; and the retrofit isolates `summarize_case` in the same `except (AttributeError, KeyError, TypeError, ValueError)` the loader already uses, so an orphan already on the data branch degrades instead of aborting.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -2092,12 +3097,90 @@ def build_html(
...
+    """
+    if not survey:
+        return ""
+    rows = "".join(
+        f"<tr><td class=mono>{_esc(str(entry.get('label') or entry.get('name', '')))}</td>"
+        f"<td>{_verdict_html((entry.get('summary') or {}).get('verdict') or _DASH)}</td>"
+        f"<td class=mono>{_esc(str(entry.get('command') or ''))}</td>"
+        + (
+            f'<td><a href="survey/{_esc(str(entry.get("name", "")))}/">run area</a></td>'
+            if entry.get("staged")
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in `dd3212d`. You're right, and this was a hole I opened myself in the previous round by honouring `"staged": true` from a spec.

`_safe_case_segment()` now accepts a name only as one plain path segment (no `/`, `\`, `:`, `..`, whitespace or control characters), and it gates two places: `survey_cases_from_spec` refuses to mark an unsafe name as staged, and `_run_index_survey_html` refuses to build a link from one even if an entry arrives pre-marked. Since `report_rel` is itself built as `<rel>/survey/<name>/sanitizer_report.json` from that same value, validating the segment makes the two renderers agree by construction.

Your follow-up about them agreeing only *coincidentally* was also right — `dd3212d`'s check was still independent, so `c8abe07` adds the cross-check that `report_rel`'s directory equals the validated name.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -2471,46 +3737,197 @@ def main() -> int:
...
                 # Carry the run manifest too so the published layout
                 # (runs/<id>/meta.json) is complete, not just the reports.
                 src_meta = src_run / "meta.json"
                 if src_meta.is_file():
                     shutil.copy2(src_meta, run_out / "meta.json")
+            # Only the newest run was produced by this job, so only its area may be
+            # written from the live checkout and environment (see
+            # write_case_run_area); older ones are re-rendered from what they
+            # captured themselves.
+            current = position == 0
```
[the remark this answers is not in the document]
amd-vivekag: Fixed in `c8abe07` — good catch, this is the same class as the provenance bug from the first round and I'd used the wrong proxy for it.

Confirmed: the dir id is `<date>-<GITHUB_RUN_ID>` and a re-run reuses its original, lower run id, so it sorts behind a newer same-day run. Keying "the run this job produced" on list position then wrote this job's container image, rocjitsu bundle and recipe copy into the newer run's area, and co-published its survey under a run that hadn't swept it.

The generator now takes `--current-run` and matches the staged id (falling back to the newest run only when the caller doesn't say, for the results-dir/runs-root modes), and the workflow passes "$run_dir_id". `current_rel` also replaces `runs[0]` for the co-publish destination, the `report_rel` threading and the `data.json` survey attachment. The regression test publishes a re-run of an older workflow alongside a newer one and asserts the provenance and survey land on the older area, not the newer.

amd-vivekag:
Round 4 — five more, all valid. Fixed in `c8abe07`. (Replies to the inline threads are on the threads themselves; this covers the four suppressed comments.)

**`position == 0` was the wrong proxy for the staged run** — replied on the inline thread, but summarising here because it's the most consequential of the round: a re-run of an older workflow reuses its lower `GITHUB_RUN_ID`, sorts behind a newer same-day run, and so had this job stamping the *newer* area with its own container image, rocjitsu bundle and recipe copy, and co-publishing its survey under a run that hadn't swept it. The generator now takes `--current-run` and matches the staged id; the workflow passes `"$run_dir_id"`. `current_rel` replaces `runs[0]` for the co-publish destination, the `report_rel` threading, and the `data.json` survey attachment.

**`REPRODUCE.md` ran the sweep before the fixtures it needs**

Correct, and it made the whole document useless on a fresh clone: `fixtures/bin/` and `fixtures/isa/` are gitignored, so the `aorta sweep run` line failed before the reader ever reached "Rebuild the fixtures" printed below it. Reordered to checkout → rebuild → required env → sweep, with the sweep in its own block. `test_reproduce_md_runs_top_to_bottom` asserts the offsets are in that order, so it can't silently regress.

**`staged_now` included `--survey` spec names**

Right, and I'd missed that both options can contribute to the same `survey` list. A spec case whose name already existed under the current run was skipped by the run loop, never rewritten by the co-publish loop (which only handles informational entries), and so kept its raw logs even under `--keep-logs 0`. The informational entries are now tracked separately and `staged_now` is built from that list alone.

**`staged` never checked that `report_rel` points at `survey/<name>/`**

Also right, and it's the residue of the previous round's fix: I validated the name and the report link independently but never that they describe the *same* directory, so `name="foo"` with a report under `survey/bar/` advertised two different run areas — the card footer linking `bar/`, the run index linking `foo/`. `_report_rel_case_segment()` now extracts the validated case segment from `report_rel` and `staged` requires it to equal the name, so the two renderers agree by construction rather than by coincidence.

**Pruning could be undone by raising `--keep-logs`**

Correct — the logs were deleted from the published history and can't be recovered from it, but `refresh_published_case_area` set `logs_published = logs` unconditionally, so a later larger window flipped it back to `true` and the page and `REPRODUCE.md` offered files that were gone. It's now `bool(env.get("logs_published", True)) and logs`, i.e. one-way: only re-staging from the source sweep (which goes through `write_case_run_area`) can restore them.

Verification: 125 tests pass (5 new, each confirmed to fail with its fix reverted), ruff clean, workflow YAML parses, 137 published links resolve. This round I also ran the suite in a clean `git worktree` before pushing, after last round's test passed locally only because gitignored build outputs existed on my machine and then failed all three CPU legs in CI.

Apologies for the delay on these replies, by the way — my `gh` was authenticated as the wrong account for a public repo, and every comment endpoint was refusing with an Enterprise Managed User error. Sorted now.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -2437,15 +3760,17 @@ def main() -> int:
...
         )
-    # Mirror the two-class split into data.json: attach the survey list to the
-    # latest run record (additive; existing per-run keys are untouched).
-    if runs:
-        runs[0]["survey"] = survey
+        survey += informational_survey
+    # Mirror the two-class split into data.json: attach the survey list to the run
+    # this job produced (additive; existing per-run keys are untouched).
+    if current_index is not None:
+        runs[current_index]["survey"] = survey
```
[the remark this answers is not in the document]
amd-vivekag: Confirmed, and reproduced before fixing — this is the sharpest consequence of `--current-run` and I should have caught it in the same round I added it.

With `--current-run` naming the older re-run, the page's Tab 1 came from `runs[0]` while Tab 2 came from this job's survey: `runs/2026-08-05-10/survey/consan-gemm/` cards under a header labelling `2026-08-05-99` as the latest run. And you're right about `data.json` too — nothing reconstructs a past run's survey, so regenerating dropped `runs[0].survey` entirely while attaching this job's list to the older record.

Took your first suggestion. A published `survey/` directory has exactly the `<case>/sanitizer_report.json` layout an informational results dir has, so the newest run's list is rebuilt with `survey_cases_from_informational_dir(history_root/<newest>/survey, rel=runs[0]["rel"])` — full card fidelity (kernel groups, sanitizer split, verdicts, digests), not just the run-index roll-up `survey_entries_from_published` produces. That list is what `build_html` and `build_summary_md` now receive, and it is also stored on `runs[0]`, so each record in `data.json` carries the survey that belongs to it. The incoming list stays attached to `runs[current_index]`, and the co-publish loop still stages under `current_rel`, so the re-run's own area is published and reachable from its own run page.

Scoped to the history mode, where a published area exists to read; the results-dir / runs-root modes publish nothing co-located, so there is nothing to reconstruct from and `current_index` is 0 there anyway.

`test_the_page_renders_the_survey_of_the_run_it_calls_latest` covers both halves — the page and `summary.md` name the newest run's case and neither names the re-run's, and each `data.json` record keeps its own survey. It fails without the fix by rendering exactly the `runs/…-10/survey/consan-gemm/` link you predicted. The end-to-end link check also runs this third mode now: 121 relative links, all resolving, with the re-run's own area still published.

Sweeping for the same defect: `build_run_index_html` already selects per run (`survey if current else survey_entries_from_published(run_out)`), and the `gate_label` print reads `runs[0]` consistently with the page. The one remaining `current_index or 0` — in the co-publish loop — was the same substitution in miniature, so it is gone too.

Fixed in 5f4df3a.

amd-vivekag:
Round 6 — the two remaining items from the latest pass, fixed in `1837b6f`. The other four suppressed comments in that review (the `clang-offload-bundler` PATH, the case-segment and `report_rel` URL syntax, and `survey_by_name` built from the combined list) plus the inline Tab 1/Tab 2 finding were already addressed by `5f4df3a`, so I've left those threads as they are.

**1. `inputs/` omitted the transitive sources decision 10 asks for** (`_publish_recipe_inputs`)

You're right, and this is the more embarrassing of the two because I saw the symptom earlier and talked myself out of it: a probe showed `survey_inputs=False` for `consan-gemm` and I filed it as "that recipe legitimately has no source inputs". It doesn't — the recipe only *names* the artifacts it consumes.

Confirmed by walking each recipe against its own published commands:

| recipe | shipped in `inputs/` | sources its commands read |
|---|---|---|
| `daily-consan-gemm` | *(empty)* | `gemm_shapes_unique.csv`, `kernels/consan_load.hip` |
| `daily-consan-tiny` | *(empty)* | `kernels/consan_load.hip`, `kernels/tiny_vecadd.hip` |
| `daily-consan-lds-dispatch` | *(empty)* | `kernels/lds_dispatch.hip`, `kernels/lds_reduce.hip` |
| `daily-waitcheck-gemm` | `gemm_shapes_unique.csv` | `gemm_shapes_unique.csv` |

So every ConSan area told the reader to compile a source it hadn't published, and only the waitcheck case was accidentally complete because its recipe names the CSV directly.

`rebuild_input_sources()` now derives the extra sources from the same `_GENCO_ISA_SOURCES` / `_BIN_SOURCES` / GEMM-extraction tables that `rebuild_plan` renders from, so the copied set and the published commands cannot disagree. It also folds in a binary's loaded code object, since rebuilding the binary means rebuilding that object first — `consan_gemm_load` needs `consan_load.hip` *and* the CSV its object is extracted with, which is why the gemm row above has two entries rather than one.

Two tests: a pure one that walks every `daily-*.yaml`, asserts each non-generated fixture path any command names is in the published set and exists on disk (so a new recipe or a changed command can't reintroduce the gap), and an end-to-end one asserting the ConSan survey area ships exactly those two files, records them in `env.json`, and links them from `index.html`.

**2. The "Results retention" section contradicted the run-area section above it**

Correct — it still said each run directory holds "that run's three raw `sanitizer_report.json` files and a `meta.json`" and documented only `--keep 30`, so the same document described two different storage models. Rewritten to state both bounds explicitly: reports for the newest 30 runs, bulk (logs, recipe copy, copied inputs) for the newest 7 via `--keep-logs`, guardrail and survey areas alike, and a note that pruning is one-way since raising `--keep-logs` later cannot bring deleted logs back.

Verification: 130 tests pass (2 new), ruff clean, workflow YAML parses, and the link check resolves all 139 relative links in both publish modes. Also run in a clean `git worktree` before pushing, after the round-3 test that passed only because gitignored build outputs happened to exist on my machine.

amd-vivekag:
Round 7 — two more, both third-order consequences of my own earlier fixes. Fixed in `3f1cdc9`.

**1. The `staged` cross-check compared only the final segment**

You're right, and this is the third pass on the same underlying issue, so I've stopped patching the comparison and removed the reason the two links can differ at all.

Comparing segments was never going to be sufficient: the run page builds `survey/<name>/` *relative to the run it is rendering*, so `runs/other/survey/foo/` matched `name="foo"` while the footer pointed into a different run. `survey_cases_from_spec` now takes the `rel` being rendered and requires `report_rel` to equal `<rel>/survey/<name>/sanitizer_report.json` exactly — which is precisely what `survey_cases_from_informational_dir` constructs, so the two renderers agree by construction rather than by comparison. Without a run context (`rel=None`) nothing can be verified, so no spec entry may claim an area.

That made `_report_rel_case_segment` dead, so it's deleted. The test now covers all three shapes: wrong case directory, right case under the wrong run (the one that survived the segment check), and a report not under `survey/` at all.

**2. Reconstructing a published run recomputed its reproduce command**

Also correct, and it's the same rule I applied to `REPRODUCE.md` in round 3 but missed on the reconstruction path added in round 5. `displayed_survey` rebuilds the newest run's Tab 2 list from the areas published under it, and `survey_cases_from_informational_dir` derived each `command` from the current `_survey_recipe_for` map — so a recipe rename would rewrite a historical run's advertised command while that area's `env.json` still recorded the one it ran.

The loader now prefers a published area's recorded command and falls back to the recipe map only for a fresh sweep directory, which has no manifest. A malformed manifest degrades to the map rather than raising.

Verification: 131 tests pass (1 new; both fixes confirmed to fail when reverted), ruff clean, 139 links resolve, and the suite also run in a clean worktree.

One process note: round 5 (`5f4df3a`) was pushed from a different session while this one was idle, and it had already fixed four of the six items I'd triaged from the earlier review. I fast-forwarded and re-read that commit before writing anything, so nothing here duplicates or reverts it.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -2372,6 +3736,30 @@ def main() -> int:
...
         help="in --history-root mode, render only the newest N runs (default 30)",
     )
+    ap.add_argument(
+        "--keep-logs",
+        type=int,
+        default=7,
+        help="keep sanitizer logs + the recipe copy and its inputs for only the "
+        "newest N runs (default 7), guardrail and survey areas alike. Older "
+        "retained runs keep their report, manifests and landing page; their bulk "
+        "is pruned so the data branch stays bounded (#384).",
```
vivekkhandelwal1: Worth qualifying "the data branch stays bounded": it's true of the checkout, not of the branch history. I tested it directly — commit a 400 KB blob, delete it the way `_prune_run_area_bulk` does, commit again, `git gc --aggressive --prune=now`: working tree goes to 0 bytes, `.git` stays at ~428 KB, because the blob is still reachable from the commit that introduced it. `sanitizer-results` accumulates ordinary commits indefinitely, so every gzipped log ever published is permanent even after its area is pruned.

Day to day this costs little (the publish job clones `--depth 1`, and Pages deploys from the workspace copy), so the exposure is remote repository size rather than CI time. The reason I'd flag it anyway is that it changes your deadline: combined with the test plan's note that real ConSan log sizes are unmeasured, a large log becomes unfixable-in-place the moment it lands, since shrinking the branch afterwards means rewriting history. Decision 6's size cap is cheap before the first publish and expensive after.

Suggest folding a `du -sh` of the staged case dirs into the same first-nightly check as the `.log.gz` curl, so you see the real number while capping is still just a code change.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -1064,18 +1130,1207 @@ def survey_cases_from_informational_dir(
...
+        if env.get("logs_published", True)
+        else "Every file published for this case. Its logs, recipe copy and inputs "
+        "were pruned for falling outside the nightly's log-retention window."
+    )
+    return (
+        "<!doctype html>\n"
+        "<html lang=en><head><meta charset=utf-8>\n"
+        '<meta name=viewport content="width=device-width, initial-scale=1">\n'
+        f"<title>{_esc(title)} &middot; {_esc(str(env.get('case', '')))}</title>\n"
+        f"<style>{_CSS}\n  .wrap {{ max-width:900px; }}\n"
```
vivekkhandelwal1: This inlines the whole dashboard `_CSS` into every run-area landing page. Measured: `_CSS` is 18,432 bytes and a rendered page is 20,574, so the stylesheet is ~90% of it — and the page only uses `.wrap`, `.navrow`, `.topbar`, `.panel`, `.kv`, `.repro`, `.table-wrap`, the chips and the three rules appended just below.

At `--keep 30` with five areas per run that's ~150 copies, about 2.6 MB, which is more than the rest of the published tree combined and well above the gzipped logs `--keep-logs` exists to bound. My six-run simulation was already at 1.07 MB total. It's byte-stable across re-publishes so it doesn't churn git, but every new run adds ~92 KB of duplicated CSS that is never deduplicated.

A single stylesheet at the dashboard root (`../../assets/run-area.css`, `../../../` from a survey area — `case_dir_up` already computes exactly that prefix) or a trimmed subset for these pages would recover almost all of it, with no change to how the pages render offline apart from needing the sibling file.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -1064,18 +1130,1207 @@ def survey_cases_from_informational_dir(
...
+                "commands": [
+                    *prelude("isa"),
+                    f"hipcc --genco --offload-arch={target} {_runnable(source)} -o tmp.o",
+                    # One command, because which branch is correct depends on the
+                    # ROCm build: --genco emits a raw object on some and a bundle
+                    # on others, and the recorded digest is of the unbundled one.
+                    "if head -c 24 tmp.o | grep -qF __CLANG_OFFLOAD_BUNDLE__; then "
+                    "clang-offload-bundler --type=o --unbundle --input=tmp.o "
+                    f"--targets=hipv4-amdgcn-amd-amdhsa--{target} --output={out}; "
+                    f"else cp tmp.o {out}; fi",
```
vivekkhandelwal1: The published command leaves `tmp.o` behind in whatever directory the reader pasted it into — which REPRODUCE.md has just told them is the repo root, so it lands as an untracked object file next to `pyproject.toml`. The workflow's own `genco_object` avoids this with `tmp="$(mktemp --suffix=.o)"` and a trailing `rm -f "$tmp"`.

Given the emphasis elsewhere on these commands being executable verbatim (down to matching `-O1 -g` so the recorded digest reproduces), it'd be worth matching that here too — either use `mktemp` like the workflow, or append `&& rm -f tmp.o` to the conditional so the temporary doesn't outlive the rebuild. Both branches of the `if` consume `tmp.o`, so a single cleanup at the end covers it.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -1064,18 +1130,1207 @@ def survey_cases_from_informational_dir(
...
+    The two sanitizers name the same kind of thing differently: waitcheck records
+    its binary as ``path``/``sha256``, ConSan records the repro binary as
+    ``command``/``command_sha256`` and its hook as ``hook``/``hook_sha256``, and
+    worklist kernels arrive from ``report_digests`` already keyed as
+    ``code_object:<basename>``. A run area names the artifacts it did not publish
+    by their recipe-relative path, so index by basename and look them up with one
+    rule -- keying only on ``code_object:`` left every ``fixtures/bin/`` repro
+    binary showing an em dash while its digest sat in the report unused.
+    """
+    index: dict[str, str] = {}
```
vivekkhandelwal1, line 1334: Noting for the follow-up you flagged in the description: this helper is precisely what the ConSan-backend em dash on Tab 2 needs. `summarize_case()` looks only for `path`/`sha256`, and the reason ConSan cards render `—` is that it records the repro binary as `command`/`command_sha256` — which is the exact key-naming mismatch `artifact_digest_index` already normalizes here.

Agreed it's out of scope for a PR this size, but it's gone from "a small separate fix" to roughly a two-line one now that the normalization exists. Worth filing so it doesn't get lost.

vivekkhandelwal1 (approved):
Reviewed carefully given the size (3,407 additions across the generator and its tests). Rather than read the resolved Copilot threads and take them on trust, I checked out the branch and tried to reproduce each of the earlier findings myself, plus the claims in the description. Everything I could test holds up, so this is an approve with follow-ups — none of my findings are correctness or security blockers.

## What I verified independently

**The earlier findings are genuinely fixed, not papered over.** I re-ran the two that mattered most:

- *Path traversal / link forgery via a `--survey` spec.* I fed the generator a spec with `name: "../../escape"`, `report_rel: "javascript:alert(1)"`, `report_rel: "//evil.example/x.json"`, `report_rel: "runs/x?a/survey/frag/..."`, and — the round-7 case — a `staged: true` entry whose `report_rel` pointed into a *different* run than the one being rendered. No area link was emitted for any of them, no dangerous `href` reached the page, and nothing escaped onto the filesystem. Requiring the full `\{rel}/survey/\{name}/sanitizer_report.json` match rather than just the trailing segment is the right closure, because it pins the area to one run instead of making the two renderers agree by coincidence.
- *Malformed nested reports aborting publication.* I planted `{"checks": null}` in the informational dir and `{"checks": null, "worklist": null}` as an orphan directly on the data branch, then published twice. Both publishes exited 0, the malformed incoming case did not persist into history, and the orphan degraded instead of taking the dashboard down. That was the sharpest bug of round 2 and it's properly closed at both ends.

**Retention does what the description says.** I simulated five sequential nightlies with `history_root == out_dir/runs` (the CI arrangement, which is the only way this is observable — you're right that a single publish can't show it). Logs were gzipped in place, and the recipe copy, `inputs/` and logs were pruned exactly as each run left the `--keep-logs` window, guardrail and survey areas alike. I also confirmed the in-place path is the one CI actually takes: `runs_dir="$tmp/dashboard/runs"` with `--out-dir "$tmp/dashboard"` makes `history_root` equal `out_dir/runs`, so `history_within_runs_out` is true, the stale-clear and copytree are both skipped, and the tree that gets gzipped/pruned is the one `git add -A` commits.

**The gzip determinism claim is real.** `mtime=0` with `filename=""` produces byte-identical output across time, and the OS byte is `255` (unknown) rather than a platform value, so it's stable across runners too. Following through on the consequence: I diffed already-pruned areas across a subsequent publish and they were byte-for-byte stable, so re-rendering 30 runs nightly genuinely doesn't churn the branch.

**Links resolve.** Independently walking every `href`/`src` in the published tree, 260 of 261 relative links resolve to a file or a directory with an `index.html`. The one that doesn't is the pre-existing `← back to CI dashboard` → `../`, which resolves to the Pages root in production and isn't yours.

**The rebuild commands match the workflow.** I diffed `_GENCO_ISA_SOURCES`, `_BIN_SOURCES` and `_BIN_OBJECT` against the actual fixture-build lines and they agree artifact by artifact — the `-O1 -g` on the two guardrail repro binaries, the `-DOBJECT` / `-DLDS_HSACO` defines and which object each names, `--top-n 3` vs `--top-n 0 --consan-object`, and the genco conditional right down to `head -c 24 | grep -qF __CLANG_OFFLOAD_BUNDLE__` and `--targets=hipv4-amdgcn-amd-amdhsa--gfx950`. Keeping these as per-artifact tables with the reasoning in a comment is the right call; a generic hint really would produce a digest mismatch.

**The workflow hop is sound.** `rocjitsu.json` is written at line 154 before any sweep touches `SANITIZER_OUT`, sits inside the uploaded `path: .sanitizer-nightly/out`, and is read back from `incoming/`. It can't disturb the gate: `compare_verdict_baselines.py` enumerates a fixed `_CASES` table rather than iterating the directory. `AORTA_CI_IMAGE` extracts correctly — `docker/Dockerfile.ci-gpu` is single-stage, so `tail -n1` gets the digest-pinned `FROM`. And `cp -a` of a whole case dir is bounded to what `pipeline.py` writes (`waitcheck/`, `consan/`, the report), so nothing unexpected rides along.

131 tests pass and `ruff check` is clean on both Python files.

## Follow-ups (all non-blocking)

**1. Pruning bounds the checkout, not the repository — and that changes when you have to act.** This is the one I'd most want on your radar. The description says `--keep-logs` keeps "the data branch bounded", and that's true of the working tree but not of the branch's history: publishing a 400 KB blob and then pruning it leaves the tree at 0 bytes but `.git` still ~428 KB after `git gc --prune=now`, because the blob stays reachable from the commit that added it. Since `sanitizer-results` accumulates normal commits forever, every gzipped log ever published is permanent. That interacts badly with your own note that real ConSan log sizes are unmeasured: if a log turns out large, decision 6's size cap has to land *before* the first publish, because afterwards you can only fix it by rewriting the branch. Concretely, I'd do the `du` on the first nightly's staged case dirs in the same pass as the `.log.gz` curl check you already planned, and treat a large log as something to cap immediately rather than next sprint. Details inline.

**2. The run-area landing page inlines the entire dashboard stylesheet.** `_CSS` is 18,432 bytes and it goes into every case's `index.html`, which makes it about 90% of a 20.5 KB page while the page uses only a handful of classes. At `--keep 30` × 5 areas that's roughly 150 copies ≈ 2.6 MB — more than everything else in the published tree combined, and considerably more than the gzipped logs the retention window exists to bound. It's byte-stable so it doesn't churn git, but each new run adds ~92 KB. A shared stylesheet next to the dashboard root, or a trimmed subset for these pages, would recover most of it.

**3. The published genco command leaves `tmp.o` in the repo root.** The workflow's own `genco_object` uses `mktemp` and `rm -f "$tmp"`; the published version writes `tmp.o` into the reader's CWD and never cleans it up. Small, but paste-and-run fidelity is the whole point of this section, and it drops an untracked object into the repo root.

**4. `summarize_case()` ignoring ConSan's backend keys is now a two-line fix.** You flagged the em-dash Backend on Tab 2 ConSan cards as deliberately out of scope, which is fair for a PR this size — but this PR adds the very helper that resolves it, since `artifact_digest_index()` already normalizes `command`/`command_sha256` against `path`/`sha256`. Worth an issue so it doesn't get lost now that the hard part exists.

**5. Cosmetic.** The Files caption reads "Logs are gzipped." for any in-window area, including a case that produced no logs at all.

Nothing here needs to block the merge. The first nightly is the real test as you say, and I'd prioritise measuring log sizes on it over the other items for the reason in follow-up 1.

scripts/sanitizers/gen_sanitizer_dashboard.py
amd-vivekag: Fixed in `45dd86e` — thanks, and thanks for measuring it rather than just flagging it. I reproduced your numbers: `_CSS` is 18,432 bytes, 81% of a 22.6 KB page, and my six-run fixture was carrying 295 KB of duplicated CSS across 16 pages, projecting to ~2.76 MB at `--keep 30` × 5 areas.

Both drill-down page types now link a single `assets/run-area.css`. You were right that `case_dir_up` already computes the prefix — the run area passes its existing `up`, and `build_run_index_html` gained the same parameter defaulting to `../../`. Your point about it needing the sibling file is why the asset is written by the same generator pass that renders the pages, so it cannot be absent.

Measured after: run-area pages 22.6 KB → 3.8 KB, whole published tree 664 KB → 312 KB. I swept the other emitters too — the per-run `runs/<id>/index.html` had the identical problem (30 more copies), so it links the sheet as well. The root dashboard keeps its inline copy: there's only ever one of it, and its empty-runs branch has to render the stale banner standalone, which is the case #378 specifically fixed.

The test asserts every drill-down page links the sheet, that the link resolves from that page's depth, and that no page still contains `_CSS`.

scripts/sanitizers/gen_sanitizer_dashboard.py
amd-vivekag: Fixed in `45dd86e`. Agreed — paste-and-run fidelity is the whole point of that section, and leaving an untracked `tmp.o` next to `pyproject.toml` undercuts it.

I took your `&& rm -f tmp.o` suggestion rather than `mktemp`: it keeps the command a single pasteable line, and the `&&` means a failed conversion still leaves `tmp.o` for inspection, which matches what the workflow's `genco_object` effectively does (it `return 1`s before its `rm -f`). As you noted, both branches consume `tmp.o`, so one cleanup at the end covers it. The test asserts the command that creates it is paired one-to-one with a command that removes it.

scripts/sanitizers/gen_sanitizer_dashboard.py
amd-vivekag: Fixed the claim in `45dd86e`, and thanks for testing it rather than reasoning about it — you're right, and I was being loose. Pruning empties the working tree but the blob stays reachable from the commit that added it, so on a branch that accumulates ordinary commits forever, everything ever published is permanent.

Both places that claimed otherwise now say so: the `--keep-logs` help text and the retention section in `docs/ci-nightly-eval.md`, which also records why it matters — the exposure is remote repository size, not CI time (the publish job clones `--depth 1`), but it moves the deadline on decision 6's size cap from "code change" to "history rewrite".

On the `du`: I went slightly further than a one-off first-nightly check and put `du -sh` of the staged case dirs into the publish job itself, printed before gzip on every run. Same number you asked for, but it can't be missed on the one night someone forgets to look, and it stays useful as the logs grow. The comment above it says why the deadline is asymmetric so whoever reads that output knows what to do with a big number.

scripts/sanitizers/gen_sanitizer_dashboard.py
amd-vivekag: Filed as ROCm/aorta#388, with the key-naming detail you spelled out — `summarize_case()` matching only `path`/`sha256` while ConSan records `command`/`command_sha256` — and a note that `artifact_digest_index()` in this PR already normalizes exactly that, so it's the two-line fix rather than the small separate one. Acceptance criteria cover the ConSan case, waitcheck staying unchanged, and an empty backend block still rendering the em dash rather than `None`.

Keeping it out of this PR as you suggested, but it won't get lost now.

amd-vivekag:
@vivekkhandelwal1 thank you — reproducing the earlier findings rather than trusting the resolved threads is a real service on a diff this size, and independently checking that CI actually takes the in-place path (`history_within_runs_out` true, so the tree that gets gzipped is the one `git add -A` commits) caught the thing I'd have most wanted a second pair of eyes on.

All five follow-ups are done in `45dd86e`. Replies are on the individual threads; this covers the cosmetic one from your summary, which had no thread:

**5. "Logs are gzipped." for a case with no logs** — fixed. The caption now has three states: pruned, in-window with a `.log.gz` in the listing, and in-window with none, which just reads "Every file published for this case." It keys off the actual file list rather than `logs_published`, so it can't claim a file that isn't there.

Recapping the other four:

- **CSS duplication** — both drill-down page types now link one `assets/run-area.css`. Your numbers reproduced exactly (18,432 bytes, 81% of a page); after the change run-area pages are 22.6 KB → 3.8 KB and the whole published tree 664 KB → 312 KB on a six-run fixture. I swept the per-run `runs/<id>/index.html` too, which had the same problem for another 30 copies.
- **Boundedness** — the claim is qualified in both the `--keep-logs` help text and the retention docs, with the reason it matters (the deadline asymmetry, not the bytes). I put `du -sh` of the staged case dirs into the publish job itself rather than a one-off check, so the number shows up every night and can't be missed on the one run nobody looks at.
- **`tmp.o`** — cleaned up with the `&& rm -f` you suggested, which also preserves it for inspection when the conversion fails, matching what `genco_object` effectively does.
- **`summarize_case()` / ConSan Backend** — filed as #388 with your key-naming detail and a note that `artifact_digest_index()` already normalizes it, so the two-line framing is captured. Kept out of this PR as you suggested.

On follow-up 1 specifically: agreed that measuring log sizes is the highest-priority item on the first nightly, ahead of the `.log.gz` CDN check. If the number is large, decision 6's size cap goes in before that publish lands rather than after.

135 tests pass (5 new), ruff clean, `bash -n` clean on every workflow `run` block, and 159 relative links resolve — the count went up because the stylesheet `href` is now a checked link too.

Issue #405, opened 2026-08-25
sanitizers: ConSan cannot instrument TokenSpeed Gluon attention kernels — register-saturated objects have no persistent SGPR window (upstream rocm-systems#11019)


vivekkhandelwal1:
**Updated 2026-09-01.** The original premise of this issue is now obsolete and the
title has been changed to match. Rewritten below.

## Where this stands

The two ConSan defects this issue was opened against are **fixed upstream** in
rocjitsu `bed84edd68` ([ROCm/rocm-systems#10955](https://github.com/ROCm/rocm-systems/issues/10955),
closed), and the fix is verified end to end on our own objects:

| Defect | Status |
|---|---|
| Barrier sites counted in the aggregate but never itemized | **Fixed** — itemized barriers now equal `barrier_discovered` (7/7 and 12/12 across two objects) |
| Every supported site reported `instrumentation_patch_missing` | **Fixed** — now correctly typed `outcome=resource_failed`, `lowering_reason=unsupported_resource_plan`, `resource_reason=no_legal_window` |

Verified as a controlled A/B on gfx950 / ROCm 7.0.2.2 against pre-fix bundle
`4227d40fb5b4`, on re-harvested objects whose inventories match the originals
exactly (232 access / 12 barrier, and the 77 / 12 gemm control). No regression on the
gemm path: 77/77 and 12/12, `analysis_complete=true`, identical on both bundles.

## What is actually left

**ConSan coverage for TokenSpeed is still gemm-only**, but for a different and
legitimate reason than this issue originally claimed. It is not a lowering bug. The
MOI Record/Replay engine needs a persistent hardware dispatch-ID SGPR pair, and the
attention kernels have no free SGPRs to give it:

| | attention `_mha_prefill` | gemm control |
|---|--:|--:|
| `numbered_sgpr` | **100** (of ~102 addressable) | 39 |
| `sgpr_spill_count` | **176** | 0 |
| `vgpr_count` / `agpr_count` | 289 / 33 | 57 / 0 |
| Occupancy | **1** | not limited |
| ConSan | `access=0/232`, `barrier=0/12` | `access=77/77`, `barrier=12/12` |

The kernel is at the hardware ceiling and already spilling, so `no_legal_window` is
literally true. No smarter allocation closes this, and there is no aorta-side
mitigation that preserves what we are measuring — the only local lever is to
recompile the kernels at lower register pressure, which for these kernels is
intrinsic to the tiling, so it would instrument a materially different binary than
[...]

PR #404, opened 2026-08-25
feat(tokenspeed): triage TokenSpeed kernels and serving on gfx950, and run its JIT kernels through Waitcheck and ConSan

vivekkhandelwal1:
## Summary

Wraps [TokenSpeed](https://github.com/lightseekorg/tokenspeed) — a third-party AMD-optimized inference engine — as an opaque subprocess under the existing `mode: probe` path, and runs its JIT-compiled kernels through the existing Waitcheck and ConSan pipelines.

**No workload class and no sanitizer-engine changes.** The point is to show aorta can triage an engine it knows nothing about using only what it already has: everything rides on the built-in `_subprocess` probe, `source.kind: kernel_list`, and the Triton ConSan loader from #403.

gfx950 only — TokenSpeed's platform check accepts gfx950/gfx1250 and raises on anything else, so MI300X cannot run this.

## Three execution routes, because no single one covers the kernels

| Route | Reaches | Gives you |
|---|---|---|
| Kernel probe (`tokenspeed-kernel-gemm-smoke.yaml`) | `gemm.mm` only (this smoke runs 2 solutions) | TFLOPs, latency, numerics |
| Suite probe (`tokenspeed-kernel-suites-smoke.yaml`) | **20 of 38 dispatched** | correctness per family, and a JIT cache for the sanitizers |
| Serving probe (`tokenspeed-serve-probe-smoke.yaml`) | — | bring-up triage, one real completion |

The benchmark harness only reaches `gemm.mm` because every other operator family is missing an input generator upstream. TokenSpeed's own op test suites build their inputs directly, so the suite probe reaches attention, MoE, quantization, sampling and transform — at the cost of correctness-only verdicts, since those suites assert rather than measure.

Coverage was measured, not assumed, and review corrected how. `map_kernel_test_coverage.py` originally instrumented `KernelRegistry.get_for_operator`, which returns *every* candidate for a family/mode — and upstream's `require` fixture calls it only to decide whether to skip, before narrowing candidates by dtype. That credited kernels the tests never ran. It now records `KernelRegistry.get_impl(name)`, the post-selection lookup a caller makes to obtain the callable for the one kernel it is about to run.

[...]

Issue #399, opened 2026-08-25
sanitizers: ConSan cannot reach JIT-compiled Triton/Gluon kernels — consan_command has no loader for a Triton .hsaco


vivekkhandelwal1:
## Summary

`mode: sanitizer` can run **Waitcheck** against TokenSpeed's Gluon kernels today with no changes. **ConSan** cannot, because the only supported way to scope ConSan to selected kernels is a caller-supplied `source.consan_command`, and the existing loader pattern assumes a hand-written HIP file — not a Triton/Gluon-generated code object.

This blocks the ConSan half of the TokenSpeed kernel integration. Waitcheck is unaffected and already useful.

## What works

TokenSpeed's Gluon GEMM kernels, harvested from the Triton JIT cache and fed in via `source.kind: kernel_list`, pass Waitcheck cleanly on gfx950:

```
sanitizer: waitcheck | state: ran | verdict: pass
    _mfma_lds_mediumm_kernel -> ran pass
    _mfma_lds_largem_kernel  -> ran pass
```

Both identities were auto-hashed by the recipe loader, so `code_object_sha256` is populated and the check is a real `state: ran`, not a `not_checked`.

## What does not work

Adding `consan` to the same recipe's `sanitizers:` list yields:

```
sanitizer: consan | state: not_checked | verdict: not_checked
  reason: consan_command_not_provisioned: targeted repro required;
          refused to wrap the whole application for 2 selected kernels
overall: not_checked (execution partial)
```

The fail-closed behaviour is correct — I am not asking for it to be relaxed. The gap is that there is no way to *satisfy* it for a JIT-compiled kernel.

## Why the existing workaround does not carry over

`recipes/sanitizers/consan-code-objects.yaml` documents the supported path: point `source.consan_command` at a small HIP loader (the `consan_app.py` pattern) that loads exactly the selected objects on-device. Per `docs/ci-nightly-eval.md`, those loaders are built per-object with a `-DOBJECT` define naming the object they load.

That works when the code object comes from a known HIP source. A Triton/Gluon object is different in ways that defeat a hand-written loader:

1. **The object does not exist until the kernel runs.** It is produced into `TRITON_CACHE_DIR` on first use, keyed by a content hash, and is specific to the image, the GPU target, and the compiled shapes. There is nothing to pin in a fixture directory.
[...]

PR #408, opened 2026-08-27
fix(sanitizers): report un-itemized ConSan sites as a coverage gap, not a parse error

amd-vivekag:
Addresses the aorta half of #405.

That issue reports two defects. **Defect 1** — ConSan discovers 232 access
sites on TokenSpeed's Gluon attention kernels, calls every one of them
*supported*, then lowers none of them, each reporting
`instrumentation_patch_missing` — needs a RocJITsu fix and is not actionable
here. **Defect 2** is ours, and it is what made defect 1 undiagnosable from a
report: the hook counts 12 barrier sites and itemizes none of them, our strict
coverage cross-check raised on the disagreement, and the run surfaced as

```
consan_output_parse_error: reader 1 barrier site count mismatch
```

An opaque parser complaint that reads like an aorta bug, for a run whose real
story was *"0 of 232 sites could be instrumented"*. All 20 attention runs
reported only this, and the actionable numbers never reached
`sanitizer_report.json`.

## What changed

A site kind the hook **counted but never itemized** is now a coverage *gap*
rather than malformed output. Malformed and internally-inconsistent evidence
still raises; so does *partial* itemization, which is lossy output with no
honest interpretation. What a gap gets instead is a named rejection:

```
consan_coverage_incomplete: verdict analysis_complete=false;
  verdict static_complete=false;
  access patched/supported mismatch: 0/232;
  barrier patched/supported mismatch: 0/12;
  access placement_or_lowering_failed: 232 instrumentation_patch_missing;
  reader 1 barrier sites not itemized: 0 of 12
```

Three things that the parse error used to discard now reach the report: the
per-object coverage entries (so the dashboard renders `0/232` instead of
nothing), the lowering reason every failed site gave, and any race finding from
the same run — a race found while coverage was incomplete is still a race.

That last one is a small correctness fix in its own right: on the parse-error
path `findings` was dropped along with everything else, so a Record/Replay
conflict in an un-itemized run came back as `error` with no findings rather
than `fail`.

## Points a reviewer will want flagged

**This relaxes a fail-closed guardrail, so the question is whether it can buy a
PASS. It cannot, and that is tested rather than asserted.** An un-itemized kind
always contributes a rejection reason, so `accepted` is always false. The
branch that worried me is the one where the aggregate looks *healthy* —
`barrier_patched=2` with no site to corroborate it — because it skips the
per-kind disposition/outcome cross-checks. `test_unitemized_sites_never_pass_even_when_the_counts_look_healthy`
pins exactly that shape, and I mutation-checked it: deleting the rejection
reason turns that run into a `PASS`, and the test fails. Per §49 of my own
review notes ("visibility is not a substitute for failing closed") — nothing
here is downgraded to a warning, and no wrong-but-plausible output is
persisted or blessable into a baseline.

**Verdict baselines are unaffected.** `verdict_baselines.json` compares
verdicts, execution status and finding shape, not reason strings, and the three
committed cases are fully-itemized healthy/racy paths.

**The `RJ_CONSAN_LOG` contract is unchanged.** At `kLogInfo` the hook emits the
aggregate line but no per-site records, so every kind reports as un-itemized
and the run still fails closed — the reason is now readable instead of a count
mismatch. `run_consan` still pins `kLogDebug=3`; the comment there and in
`daily-consan-gemm.yaml` were re-read against the new behaviour.

**Two drive-by cleanups, both inside the function being changed.** The
`coverage_site` branch now builds a typed `_SiteRecord` instead of a 5-tuple
addressed by index (it needed two more fields, and `site[4]` was already
opaque); and the loop-local `record` string was renamed `payload` because it
shadowed the `CoverageRecord` of the same name in the cross-check loop below.
The rename takes `consan_coverage.py` from 8 pre-existing mypy errors to 0.

## Test plan

- [x] `pytest tests/instrumentation/rocjitsu_sanitizers/ tests/sanitizers/` — 450 passed, 3 skipped
- [x] The 4 behavioural tests fail against `main` with the exact string from the
      issue (`consan_output_parse_error: reader 1 barrier site count mismatch`);
      the partial-itemization guard passes on both, which is the point — it
      proves the relaxation is bounded
- [x] Mutation check on the false-PASS guard (above)
- [x] `ruff` clean; `black`/`mypy` clean on `consan_coverage.py`; the
      pre-existing `black` findings in `consan.py` / `test_consan.py` are
      untouched and none fall in the added lines
- [ ] Not exercised on hardware — the fixtures are transcribed from the log
      excerpts in #405 rather than captured from a fresh gfx950 run

Defect 1 and this measurement are written up in
`docs/sanitizers/consan-405-unitemized-barrier-sites.md`, following the
`consan-4112` precedent. #405 stays open for the RocJITsu half.

### DIFF
--- a/docs/sanitizers/consan-405-unitemized-barrier-sites.md
+++ b/docs/sanitizers/consan-405-unitemized-barrier-sites.md
@@ -1,0 +1,133 @@
+# ConSan on TokenSpeed's Gluon attention kernels — nothing lowers, and an unitemized barrier count hid it
+
+Filed as [ROCm/aorta#405](https://github.com/ROCm/aorta/issues/405), found while
+wiring TokenSpeed's JIT kernels through the Triton ConSan loader
+([#403](https://github.com/ROCm/aorta/pull/403), closing
+[#399](https://github.com/ROCm/aorta/issues/399)). The loader works; this is
+about what ConSan does once it reaches the object.
+
+Two defects, and the second hid the first.
+
+| | Owner | Status |
+|---|---|---|
+| **1.** No site can be lowered on the attention kernels | RocJITsu (rocm-systems) | open |
+| **2.** Barrier sites are counted but not itemized, so aorta could not report defect 1 | aorta | **fixed** — this document's other half |
+
+## Measured on
+
+- gfx950 (cdna4), ROCm 7.0.2.2
+- rocjitsu `4227d40fb5b4ea76273589c56dac069af08b7aab`, artifact run 32745941408
+  — the same bundle
+  [`consan-4112-overlapping-anchor-patches.md`](consan-4112-overlapping-anchor-patches.md)
+  records as carrying both the `dc7c8e04` and `15275dad` fixes, so none of this
+  is a stale-artifact effect
+- `scripts/sanitizers/triton_consan_loader.py` in `load` mode, lenient policy
+- objects harvested from TokenSpeed's Triton cache
+  (`lightseekorg/tokenspeed-amd:nightly-20260714`)
+
+## Defect 1 — no site lowers on the attention kernels
+
+Gluon **gemm** kernels from the same JIT and the same harvest path instrument
+cleanly:
+
+```
+access_discovered=77  access_supported=77  access_selected=77  access_patched=77
+barrier_discovered=12 barrier_supported=12 barrier_patched=12
+analysis verdict static_complete=true  access=77/77 barrier=12/12
+```
+
+Gluon **attention** kernels patch nothing:
+
+```
+access_discovered=232 access_supported=232 access_selected=232 access_patched=0
+                      access_placement_or_lowering_failed=232
+barrier_discovered=12 barrier_patched=0 barrier_placement_or_lowering_failed=12
+coverage_site ... outcome=placement_or_lowering_failed lowering_reason=instrumentation_patch_missing
+```
+
+Every site is discovered, reported *supported*, then fails to lower with
+`instrumentation_patch_missing`. Uniformly: 20 of 20 harvested attention
+objects, 2502 sites, zero patched, no exceptions. Affected kernels are
+`_fwd_kernel`, `_mha_prefill` and `_mha_prefill_sliding`, each in several shape
+specializations. Since gemm objects from the same pipeline patch fine this is
+kernel-dependent rather than a provisioning problem; the attention kernels are
+the more aggressive Gluon code, and `ds_read_b64_tr_b16`, `ds_read2_b64` and
+heavy LDS staging dominate their site inventory.
+
+Waitcheck is unaffected and passes on all 20 attention objects, so the kernels
+are reachable — ConSan just cannot instrument them. **This half needs a RocJITsu
+fix and is not actionable in aorta.**
+
+## Defect 2 — the reporting gap that hid it
+
+On the failing path the hook emits `coverage_site` records for the access sites
+but none for the barrier sites, while still reporting `barrier_discovered=12`:
+
+| | `coverage_site` records emitted | aggregate `coverage` record says |
+|---|---|---|
+| gemm, patched | 77 access + 12 barrier | `access_discovered=77 barrier_discovered=12` |
+| attention, unpatched | 232 access + **0 barrier** | `access_discovered=232 barrier_discovered=12` |
+
+`consan_coverage.py` cross-checks itemized sites against the discovered counts
+and used to fail closed on any disagreement:
+
+```
+consan_output_parse_error: reader <N> barrier site count mismatch
+```
+
+The cross-check is right to exist — coverage that cannot be seen should not be
+trusted. But a run whose real story is *"0 of 232 sites could be instrumented"*
+surfaced as an opaque parser complaint that reads like an aorta bug, the
+actionable numbers never reached `sanitizer_report.json`, and the parse error
+discarded the rest of the parsed output along with them. All 20 attention runs
+reported only this.
+
+### What changed
+
+A site kind the hook *counted* but never itemized is now treated as a coverage
+**gap** rather than malformed output. Nothing is trusted that was not seen: the
+decision is still rejected, the check is still `error`, and the un-itemized kind
+is named with its count. What survives is the evidence:
+
+```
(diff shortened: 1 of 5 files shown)


vivekkhandelwal1 (approved):
Reviewed at `db4d1e6c`. The core judgement is right and the safety claim holds — I verified it independently rather than trusting the test. Two things to fix, neither a correctness risk.

## The relaxation is bounded, and I checked the boundary myself

The distinction the PR rests on is that a kind counted-but-never-itemized is a gap while partial itemization is malformed. I traced all four quadrants of `discovered` vs `retained`:

| `discovered` | itemized | behaviour |
|---|---|---|
| `> 0` | none | gap (named, rejected) |
| `> 0` | fewer than discovered | raises count mismatch |
| `> 0` | equal | full per-kind cross-checks run |
| `0` | any | raises count mismatch |

So the `continue` only skips reconciliation for the one kind that has nothing to reconcile against, and only when there is genuinely nothing. That is the narrowest possible carve-out.

**What I found more reassuring than the tests: the no-PASS property is structural, not incidental.** `accepted=not reasons`, and `reasons.extend(unitemized)` runs unconditionally, so a non-empty gap list forces rejection by construction. It cannot be defeated by a future reordering of the reason-building block.

I still mutation-checked it independently. Deleting `reasons.extend(unitemized)`:

```
test_unitemized_sites_never_pass_even_when_the_counts_look_healthy
E  AssertionError: assert <Verdict.PASS: 'pass'> is <Verdict.ERROR: 'error'>
```

That confirms your claim exactly — the healthy-aggregate shape does become a `PASS` without the reason, and the guard catches it. That was the right branch to worry about, since it is the one the `continue` leaves uncorroborated.

Also worth noting explicitly, because it is what makes the carve-out safe rather than merely narrow: the aggregate cross-checks *after* the per-kind loop are untouched. `applicable_code_objects`, `incomplete_code_objects` and the per-kind `patched/supported` aggregate-vs-coverage comparisons all still raise. So the hook cannot use an un-itemized kind to smuggle an inconsistent aggregate past the parser.

I reproduced the rest of the test plan: 450 passed / 3 skipped, and the four behavioural tests fail against `main` with the literal string from the issue —

```
consan_output_parse_error: reader 1 barrier site count mismatch
```

— while `test_partially_itemized_site_kind_is_still_a_parse_error` passes on **both** branches. That pairing is the actual proof that the relaxation is bounded, and it is good that you called it out as the point rather than listing it as one more green test.

The dropped-`findings` fix is real too. On `main` the same input returns `findings=(), coverage=()` — everything went out with the parse error, so a Record/Replay conflict in an un-itemized run really did come back as `error` with no findings.

---

## 1. The committed repro cannot be followed on `main`

`docs/sanitizers/consan-405-unitemized-barrier-sites.md:115` documents:

```bash
python3 src/aorta/workloads/tokenspeed/harvest_code_objects.py ...
```

That file does not exist in this repository, and there is no `tokenspeed` directory under `src/aorta/workloads/` at all. I searched the remote:

```
harvest_code_objects.py on any remote branch -> FOUND on feat/tokenspeed-probe-integration
"tokenspeed" mentioned anywhere on main      -> (no matches)
```

It lives only on the branch of **open PR #404**. So this doc lands in `docs/sanitizers/` as the durable record of the defect while its Reproducing section silently depends on unmerged work, and the word TokenSpeed appears nowhere else on `main` to give a reader a foothold. Someone hitting this in three months gets `No such file or directory` and no hint why.

Cheapest fix is a sentence saying the harvest step requires #404, or a link to it. I would not restructure the doc — the rest of it is accurate and the `consan-4112` precedent is followed properly. (The `#399`/`#403` cross-references both check out: closed and merged respectively.)

This is the same class as the worked example on #403, and it is worth being strict about precisely because the doc outlives the PR.

## 2. The gap message drops `load`, so multi-object runs emit indistinguishable duplicates

The message is built from `record.reader` alone, but the identity is `(reader, load)`. With two loads under one reader, both un-itemized:

```
reason: reader 1 barrier sites not itemized: 0 of 2
reason: reader 1 barrier sites not itemized: 0 of 2
distinct reason strings: 1 of 2
```

Two objects, two byte-identical lines, and no way to tell which object either refers to.

**Why this is newly a problem rather than pre-existing:** the raise path has the same ambiguity, but it fails fast, so exactly one message ever surfaced and there was nothing to disambiguate. The gap path *accumulates* across records, which is the whole point of the change — so the ambiguity becomes visible for the first time here, as duplicate lines in the output this PR exists to make readable.

The information is already at hand: `load` is part of `record.identity`, and the coverage entry you now persist carries it in `fields`. So the message contradicts its own record on whether `load` matters. `f"reader {record.reader} load {record.load} ..."` is the whole fix.

This is directly relevant to #405's shape rather than hypothetical — 20 attention objects all failing the same way is exactly the case that produces N identical lines.

## 3. One thing I checked and am satisfied with, noting it for the record

The un-itemized branch persists the hook's aggregate counts uncorroborated — `barrier_patched=2` with zero sites backing it reaches the coverage entry's `fields`. Against your §49 framing ("no wrong-but-plausible output is persisted"), that is strictly speaking persisted, so I chased whether it can mislead.

It cannot, materially. The dashboard cell renders only `access_patched/access_supported`, and in the #405 shape the access kind *is* itemized and fully cross-checked, so the `0/232` you get on screen is corroborated. The uncorroborated counts stay inside the JSON next to the gap reason and an `error` state, and cannot be blessed into a baseline.

The one shape where the rendered cell is itself uncorroborated is an *access*-un-itemized run — which is exactly the `RJ_CONSAN_LOG=1`/`kLogInfo` case you discuss. There the cell shows the hook's unverified pair beside an `error` execution state. Recording what the tool claimed is the right call and zeroing it would destroy evidence, so I am not asking for a change; but if the report ever grows a "coverage verified" marker, that pair is the thing it should mark.

---

CI is fully green including both GPU legs, and I am the first reviewer, so nothing here conflicts with another read. Approving: the fail-closed guarantee is intact and provable, and both items above are single-line fixes I am happy to see land as follow-ups on this branch rather than blocking it.

I did not verify the mypy 8-to-0 claim — mypy is not installed in my environment — so that one rests on `pre-commit`, which is green.

PR #415, opened 2026-09-01
feat(instrumentation): make Proton's AMD backends usable, with an example each

### LINKED ISSUE #412: proton: no examples or options for the AMD-specific backends, and pinning one under mode: cli captures an empty tree while the trial still exits 0
The `proton` collector ships two examples (`examples/profiling/proton/triton-vecadd`, `examples/profiling/proton/triton-softmax`) and both leave `backend: auto`. Nothing in the repo shows what the individual AMD backends are *for*, and the option schema cannot express most of what upstream Proton offers for them.

The ask is three AMD-specific examples — one per AMD backend, each demonstrating that backend's distinct benefit — plus the option-schema work that makes naming an AMD backend usable at all.

"At all" is load-bearing. While scoping this I measured that pinning an explicit AMD backend under the default `mode: cli` captures **nothing**, deterministically, and the trial still exits 0 with no proton metrics. Backend examples are meaningless until that is guarded, so it is in scope here.

## How this was measured

- gfx950 (CDNA4), ROCm 7.0.2
- a ROCm venv with Triton 3.7.1 and PyTorch
- payload: the repo's own `examples/profiling/proton/triton-vecadd/vecadd.py`
- Proton invoked directly the same way `mode: cli` invokes it (`python -m triton.profiler.proton ...`), so the collector itself is not in the picture

## Finding A — an explicit AMD backend under `mode: cli` captures an empty tree

Three runs of the shipped payload, differing only in `-b`:

```bash
PAYLOAD=examples/profiling/proton/triton-vecadd/vecadd.py

# 1. no -b: what aorta's default `backend: auto` renders
python -m triton.profiler.proton -n /tmp/proton-auto \
    --context shadow --data tree "$PAYLOAD" --size 1048576 --iters 20

# 2. explicit roctracer
python -m triton.profiler.proton -n /tmp/proton-roctracer -b roctracer \
    --context shadow --data tree "$PAYLOAD" --size 1048576 --iters 20

# 3. explicit instrumentation
python -m triton.profiler.proton -n /tmp/proton-instr -b instrumentation \
    --context shadow --data tree "$PAYLOAD" --size 1048576 --iters 20
```

| `-b` | `.hatchet` size | tree contents |
|---|---|---|
| absent (`backend: auto`) | 3091 bytes | 27 kernels with real timings |
| `roctracer` | 160 bytes | `ROOT` only, `metrics` empty |
| `instrumentation` | 160 bytes | `ROOT` only, `metrics` empty |

Deterministic, 3/3 for each.

**Root cause** is in Proton's CLI, not in aorta. `_select_backend()` is called only when `-b` is *absent*, and that call initialises the Triton HIP driver as a side effect. `roctracer` records nothing unless its queue interceptor is installed after the HIP runtime is up:

```python
# triton/profiler/proton.py, Triton 3.7.1
def run_profiling(args, target_args):
    backend = args.backend if args.backend else _select_backend()

    start(args.name, context=args.context, data=args.data, backend=backend, hook=args.hook)
```

**The decisive experiment** isolates it to ordering alone, via the Python API on one payload:

- `proton.start(..., backend="roctracer")` first, `import torch` only afterwards → 160 bytes, empty tree.
- `triton.runtime.driver.active.get_current_target()` first — exactly what `_select_backend()` does — then `proton.start(..., backend="roctracer")` → 1649 bytes with real kernel data.

Nothing else differs between the two. Triton `main` has the same asymmetry for `roctracer` and `rocprofiler`.

**Consequence in aorta.** `parse_summary_from_streams` returns only `proton_artifact_dir` when the tree yields no timings (`src/aorta/instrumentation/proton/_parse.py:185`), so the trial exits 0 with no proton metrics at all — the silent-unprofiled outcome the collector's design rejects elsewhere. `docs/profiling-collectors.md:534` already carries a troubleshooting entry for a ROOT-only profile, but it frames the pin as a *hazard* ("pin one only when you need to"). The measurement says it is unconditional under `mode: cli`: the pin is not risky, it is a no-op.

## Finding B — shipped Triton drops `--mode`, so two existing options are silent no-ops

Triton 3.7.1's CLI parses `-m/--mode` and never forwards it to `start()` — visible in the snippet above, which has no `mode=`. Triton `main` does pass it.

So `instrumentation_mode` and `granularity` (`docs/profiling-collectors.md:217-218`) render a `--mode` argument that every released Triton discards. They are honoured only in `mode: env`, where the payload calls `proton.start()` itself. Today's docs claim neither limitation.

## Finding C — the schema cannot express the AMD backend options that do exist

Upstream Proton on `main` supports `pcsampling` / `periodic_flushing` as *backend* modes on `rocprofiler`, `periodic_flushing` on `roctracer`, and `-k/--hook triton`. `src/aorta/instrumentation/proton/_options.py` has no `hook` key, and its only `--mode` producer is gated to `backend: instrumentation` (`_options.py:128-134`), which is a different concept from a backend mode. A recipe cannot ask for any of them.

## Finding D — `rocprofiler` exists in no released Triton, and newer ROCm does not change that

`rocprofiler` is a Triton-version feature, not a ROCm one. Querying the images directly, every one offers `-b {cupti,roctracer,instrumentation}` and no `rocprofiler`:

| image | Triton |
|---|---|
| `rocm/pytorch:rocm7.14_ubuntu26.04_py3.14_pytorch_release_2.12.0` | 3.7.1 |
| `rocm/pytorch:latest` | 3.6.0+rocm7.2.4 |
| an AMD-internal ROCm 7.2 workload image | 3.7.0 |

PyPI's latest `triton` is 3.7.1; PyTorch's ROCm nightly tops out at `pytorch-triton-rocm` 3.6.0. On Triton `main`, `-b` choices come from `libproton.get_available_profilers()` and `rocprofiler` is present with modes `[None, "pcsampling", "periodic_flushing"]`. Reaching it means a source build. The `backend` row at `docs/profiling-collectors.md:214` is right that pinning is a version commitment; it just does not say the commitment is currently unsatisfiable for `rocprofiler`.

### Examples

Three directories under `examples/profiling/proton/`, each with the payload / `recipe.yaml` / `README.md` shape `tests/test_examples.py` already enforces, and each standalone-runnable, self-checking and cheap by default like the existing two.

All three use **`mode: env`, not `mode: cli`** — forced by Finding A, and what makes them verifiable today: an env-mode payload calls `proton.start()` itself after `import torch`, so the runtime is already initialised.

- **`amd-roctracer/`** — whole-kernel timing across a multi-kernel pipeline. Shows kernel-level attribution over a real launch sequence, and is the portable AMD path that works on every shipped Triton.
- **`amd-instrumentation/`** — an intra-kernel hotspot inside one deliberately unbalanced kernel, with `granularity: warp`. Shows attribution *inside* a kernel, and that this is the only backend that coexists with `rocprof` (no HSA queue interception).
- **`amd-rocprofiler/`** — the rocprofiler-sdk path with `backend_mode: pcsampling`. Given Finding D this one ships documented-but-unverified, with the Triton-`main` requirement stated plainly and the `get_available_profilers()` check to run first.

### Option schema

In `src/aorta/instrumentation/proton/_options.py`:

- `BACKEND_MODES: dict[str, frozenset[str]]` — `rocprofiler: {pcsampling, periodic_flushing}`, `roctracer: {periodic_flushing}`, `cupti: {pcsampling, periodic_flushing}`.
- New keys `backend_mode` (validated against `BACKEND_MODES[backend]`) and `hook` (`triton`).
- `instrumentation_mode` / `granularity` and their `backend: instrumentation` gate stay exactly as they are — no breakage, and the two concepts stay distinct.
- Reject `backend_mode` together with `instrumentation_mode`: they render into the same Proton `--mode`.

In `src/aorta/instrumentation/proton/__init__.py`: `mode_argument` renders `backend_mode`, `build_argv_prefix` emits `-k <hook>`, `build_env` exports `AORTA_PROTON_HOOK`.

### The Finding A guard

`wrap_argv` raises `ProtonWrapError` for `mode: cli` with an explicit `backend` of `roctracer` or `rocprofiler`, stating the reason (Proton's CLI initialises the runtime only when `-b` is omitted, so the capture would come back empty) and naming `mode: env` as the route. `backend: auto` is untouched — it is the working path — and so is `instrumentation`, which is not queue-intercepting.

An empty capture that exits 0 is worse than a load-time error, which is the same reasoning that already makes `instrumentation_mode` on the wrong backend an error rather than accepted-and-ignored.

### Docs

- `docs/profiling-collectors.md`: `backend_mode` / `hook` rows; a "pinning an explicit AMD backend" subsection carrying Finding A and the `mode: env` requirement; correct the `backend` row (line 214) and sharpen the ROOT-only troubleshooting entry (line 534) from hazard to unconditional; note Finding B on the `instrumentation_mode` / `granularity` rows.
- `examples/profiling/README.md`: three new table rows, and extend the backend paragraph.
- Both existing example READMEs understate pinning the same way (`proton/triton-vecadd/README.md:94-101`, `proton/triton-softmax/README.md:96-98`); fix both.

### Tests

- `tests/instrumentation/test_proton.py`: `backend_mode` accept/reject per backend, `hook` rendering in both attach modes, the `backend_mode` + `instrumentation_mode` conflict, and the new guard (raises for explicit `roctracer`/`rocprofiler`, silent for `auto` and `instrumentation`).
- `tests/instrumentation/test_proton_smoke_gpu.py`: a GPU smoke test asserting an env-mode explicit-`roctracer` capture yields a **non-empty** tree. Asserting a `.hatchet` merely exists is exactly the trap Finding A falls into — that is the regression which would have caught it.

## Out of scope: two upstream Triton bugs

Findings A and B are defects in Triton's Proton CLI, not in aorta:

- `run_profiling` calls `_select_backend()` only when `-b` is absent, so an explicit backend loses the runtime initialisation it depends on and captures nothing.
- `-m/--mode` is parsed and never forwarded to `start()`.

Both are worth reporting upstream separately. This issue only works around A (guard plus `mode: env`) and documents B; neither upstream fix is a dependency for the work above.

amd-vivekag:
Closes #412.

## Why

Pinning `proton: {backend: roctracer}` was a no-op that looked like a success.
Triton's Proton CLI front-end calls `_select_backend()` **only when `-b` is
absent**, and that call is what initialises the HIP driver as a side effect — so
a queue-intercepting backend named explicitly attaches before the first HSA
queue exists and records nothing. `parse_summary` then degrades to
`proton_artifact_dir`, so the trial published no Proton metrics while exiting 0:
the silent-unprofiled outcome the collector's design rejects everywhere else.

Measured on the shipped `triton-vecadd` payload (8x gfx950, ROCm 7.0.2, Triton
3.7.1, torch 2.13.0+rocm7.2), 3/3 deterministic:

| Wrap | Result |
|---|---|
| no `-b` (`backend: auto`) | ~3 KB hatchet, 27 dispatches across 5 kernels |
| `-b roctracer` | 160-byte hatchet, bare `ROOT` frame, exit 0 |

Ordering alone accounts for it: driving Proton from the Python API with
`import torch` deferred past `proton.start()` reproduces the empty tree, and
initialising the driver first gives a populated one.

## What changed

`wrap_argv` refuses `mode: cli` with an explicit `roctracer` / `rocprofiler`,
naming the mechanism and `mode: env` as the route. `auto` is untouched —
omitting `-b` is what keeps that path working, which is why it stays the
default. `instrumentation` is deliberately **not** refused: it installs no queue
interceptor and a `-b instrumentation` CLI wrap of a scoped payload captures
correctly (verified). What it loses under `mode: cli` is `--mode`, not the
capture.

Two schema additions, validated against upstream `main`'s `proton.start()`
docstring rather than the installed Triton:

- **`backend_mode`** — Proton's per-backend `--mode`, keyed per backend
  (`rocprofiler`: pcsampling/periodic_flushing, `roctracer`:
  periodic_flushing, `cupti`: both), so `roctracer` + `pcsampling` fails at
  recipe load instead of inside Proton. Requires an explicit backend and is
  rejected beside `instrumentation_mode`, which renders the same argument.
- **`hook`** — Proton's `-k triton` launch hook, in both attach modes.

Three examples, one per AMD backend, all `mode: env` because that is what makes
an explicit backend capture anything:

| Example | Shows | Verified end to end |
|---|---|---|
| `amd-roctracer` | three Triton kernels in sequence; the portable AMD path | `proton_kernel_count` 60 (3 kernels x 20 iters), `proton_gpu_time_ms` 0.398, top kernel `bias_gelu_kernel` |
| `amd-instrumentation` | two named scopes inside one unbalanced kernel | `expensive` 8,608,876 cycles vs `cheap` 153,368 — a 56x split inside one kernel |
| `amd-rocprofiler` | `backend_mode: pcsampling` | capture **not** verified — see below |

Also corrects the "naming a backend is a version commitment" framing wherever it
appeared (collector docstring, options table, both existing example READMEs,
vecadd's recipe): on AMD it is an attach-mode commitment too.

## Things a reviewer should push back on if they disagree

- **`amd-rocprofiler` fails the trial on every released Triton.** No obtainable
  build has that backend: 3.7.1, 3.7.0 and 3.6.0+rocm7.2.4 all offer
  `-b {cupti,roctracer,instrumentation}`, PyPI's newest is 3.7.1, PyTorch's ROCm
  nightly tops out at 3.6.0, and even `rocm/pytorch:rocm7.14_...py3.14_...2.12.0`
  ships 3.7.1. The payload therefore checks availability up front and exits 2
  with a message naming the Triton-main requirement, rather than crashing inside
  Proton. Intentional per the repo's "a measurement that cannot be taken is a
  clean setup failure" stance, but it does mean the example is a documented
  failure today.
- **`amd-instrumentation` publishes no numeric metrics.** Instrumentation leaves
  carry `cycles` / `normalized_cycles`, never `time (<unit>)` or `count`, and the
  parser keys exclusively on the former, so a perfectly healthy intra-kernel
  capture yields only `proton_artifact_dir`. Documented in the README and in the
  artifact-layout section. Teaching `_parse.py` to read cycles is a separate
  change.
- **`build_argv_prefix` still renders `-b roctracer`** even though `wrap_argv`
  refuses to pin it. The split is deliberate — renderer vs. attach-decision
  point, so a caller that only wants to display the flags still can — and
  everything inside aorta reaches Proton through `wrap_argv`. Its docstring now
  says so, and says an out-of-tree consumer building its own attach must apply
  the policy itself.
- **`granularity` is unusable on Triton 3.7.1 in either attach mode** (found
  while building the example): the CLI drops `--mode`, and under `mode: env` the
  rendered `default:granularity=warp` is rejected at kernel exit with
  `RuntimeError: Only warp granularity is supported for now` — for warp, which is
  that backend's own default there. Documented in the options table and
  troubleshooting; the example sets `instrumentation_mode` and omits
  `granularity`. The option is left in the schema rather than removed, since it
  works on upstream `main`.

## CI gap found on the way

The GPU path filter in `.github/workflows/gpu-tests.yml` did not list the Proton
collector or its smoke test, so `pytest (GPU, MI350)` reported `skipping` on this
PR — the two regression tests for the empty-capture bug would never have run in
CI. Fixed here, and verified by replaying the filter's `case` patterns against
this branch's file list.

Two sibling gaps are left alone on purpose, since they predate this branch and
fixing them would put unrelated GPU results on a Proton diff:
`tests/instrumentation/test_rocprof_smoke_gpu.py` and
`tests/instrumentation/test_layer_numerics_smoke_gpu.py` are also missing from
the filter, along with their modules. They are one line each if you would rather
they came with this PR.

## Test plan

- [x] `ruff check` clean on all changed Python files
- [x] `pre-commit` clean
- [x] 479 CPU tests pass across `test_proton.py`, `test_collectors.py`,
      `test_examples.py`, `test_recipe.py`
- [x] Full CPU suite: 4304 passed, 95 skipped. The 17 `tests/ebpf/test_runner.py`
      failures are environmental (no `bpftrace` on the host) and reproduce
      identically on `main`.
- [x] 9 GPU smoke tests pass on gfx950, including
      `test_env_mode_pins_roctracer_and_gets_a_non_empty_tree` and
      `test_cli_mode_pin_would_still_capture_nothing` — the two regressions that
      would have caught this bug
- [x] All three examples run end to end through `aorta sweep run`, asserted on
      published metrics rather than on a `.hatchet` existing (a 160-byte empty
      tree is a file too)

Upstream Triton bugs found on the way, out of scope here and worth separate
reports: the `-b`-absent-only `_select_backend()` asymmetry (still present on
`main`), and 3.7.1's CLI parsing `-m/--mode` and never forwarding it to
`start()`.

### DIFF
--- a/.github/workflows/gpu-tests.yml
+++ b/.github/workflows/gpu-tests.yml
@@ -98,10 +98,17 @@ jobs:
             # without it, so a PR touching only the probe ran neither the audit
             # nor the GPU-gated tests in that file.
             'src/aorta/instrumentation/environment.py'
+            # The Proton collector attaches a profiler to a real GPU launch, so
+            # its behaviour is only observable on hardware: the CPU suite can
+            # assert the argv and the env bundle, but not that the capture came
+            # back with kernels in it. Listed with its smoke test because a
+            # backend pin that silently records nothing passes every CPU check.
+            'src/aorta/instrumentation/proton/*'
             'scripts/audit_env_knobs.py'
             'tests/hw_queue_eval/*'
             'tests/instrumentation/test_environment.py'
             'tests/instrumentation/test_env_knob_audit.py'
+            'tests/instrumentation/test_proton_smoke_gpu.py'
             'tests/instrumentation/rocjitsu_sanitizers/*'
             'tests/sanitizers/*'
             'tests/test_marker_partition.py'
--- a/docs/profiling-collectors.md
+++ b/docs/profiling-collectors.md
@@ -211,11 +211,13 @@ Precedence and scope:
 | Option | Values | Default | Notes |
 |---|---|---|---|
 | `mode` | `cli`, `env` | `cli` | See [Attach modes](#proton-attach-modes). |
-| `backend` | `auto`, `rocprofiler`, `roctracer`, `instrumentation`, `cupti` | `auto` | `auto` omits Proton's `-b` and lets Proton pick the backend matching the active runtime — `rocprofiler` where rocprofiler-sdk is available, `roctracer` otherwise. It is the only portable spelling and is why it is the default: **naming a backend is a version commitment.** `rocprofiler` is the preferred AMD backend upstream but Triton 3.7.x and earlier accept only `cupti`/`roctracer`/`instrumentation` and exit with an argparse `invalid choice: 'rocprofiler'` before the payload runs. `roctracer` is the deprecated AMD predecessor; `instrumentation` the intra-kernel path; `cupti` the NVIDIA one, accepted so a recipe stays portable even though aorta's examples are AMD. Pin a backend when you need to know exactly which one measured. |
+| `backend` | `auto`, `rocprofiler`, `roctracer`, `instrumentation`, `cupti` | `auto` | `auto` omits Proton's `-b` and lets Proton pick the backend matching the active runtime — `rocprofiler` where rocprofiler-sdk is available, `roctracer` otherwise. It is the only portable spelling and is why it is the default: **naming a backend is a version commitment, and on AMD it is an attach-mode commitment too.** `rocprofiler` is the preferred AMD backend upstream but Triton 3.7.x and earlier accept only `cupti`/`roctracer`/`instrumentation` and exit with an argparse `invalid choice: 'rocprofiler'` before the payload runs. `roctracer` is the deprecated AMD predecessor; `instrumentation` the intra-kernel path; `cupti` the NVIDIA one, accepted so a recipe stays portable even though aorta's examples are AMD. Pinning `roctracer` or `rocprofiler` means `mode: env` and a payload that drives Proton itself: under the default `mode: cli` such a pin captures an empty tree, and the collector refuses the combination — see [Pinning an explicit AMD backend](#pinning-an-explicit-amd-backend). `instrumentation` measures *inside* a kernel and publishes **no numeric metrics** — its leaves carry `cycles` / `normalized_cycles`, never `time (<unit>)`, so such a trial reports `proton_artifact_dir` and nothing else. |
+| `backend_mode` | `pcsampling`, `periodic_flushing` | (unset) | Proton's per-backend `--mode`. **Requires an explicit (non-`auto`) backend**, because which values are valid depends on which backend runs: `rocprofiler` takes both, `roctracer` only `periodic_flushing`, `cupti` both. Not valid on `backend: instrumentation`, whose modes are the `instrumentation_mode` key, and rejected together with `instrumentation_mode` — both render into Proton's single `--mode` argument. The two AMD backends it applies to need `mode: env` for the reason given in the `backend` row. |
 | `context` | `shadow`, `python` | `shadow` | How a kernel's time is attributed to a calling frame. |
 | `data` | `tree`, `trace` | `tree` | `tree` writes the `.hatchet` file the parser reads; `trace` writes a chrome trace instead, so it produces no numeric metrics. |
-| `instrumentation_mode` | `default`, `mma`, `pcsampling` | (unset) | **Requires `backend: instrumentation`.** |
-| `granularity` | `cta`, `warp`, `warp_2`, `warp_4`, `warp_8`, `warp_group`, `warp_group_2`, `warp_group_4`, `warp_group_8` | (unset) | **Requires `backend: instrumentation`.** |
+| `instrumentation_mode` | `default`, `mma`, `pcsampling` | (unset) | **Requires `backend: instrumentation`.** Takes effect only under `mode: env`: the shipped Triton CLI parses `-m/--mode` and never forwards it to `start()`, so it is a silent no-op under `mode: cli`. |
+| `granularity` | `cta`, `warp`, `warp_2`, `warp_4`, `warp_8`, `warp_group`, `warp_group_2`, `warp_group_4`, `warp_group_8` | (unset) | **Requires `backend: instrumentation`.** Unusable on Triton 3.7.1 in either attach mode: dropped by the CLI like `instrumentation_mode`, and under `mode: env` the rendered `default:granularity=warp` string fails at kernel exit with `RuntimeError: Only warp granularity is supported for now`. Warp is that backend's own default there, so leave it unset. |
+| `hook` | `triton` | (unset) | Renders Proton's `-k triton` under `mode: cli`; exported as `AORTA_PROTON_HOOK` under `mode: env`. Registers Proton's launch hook, which records Triton kernel launch metadata alongside the timing. |
 
 `instrumentation_mode` and `granularity` are rejected with an actionable
 error on any other backend rather than accepted-and-ignored: Proton would
@@ -223,8 +225,24 @@ silently produce a profile you did not ask for. Together they render into
 Proton's single `--mode` argument as
 `<instrumentation_mode>:granularity=<granularity>`, with
 `instrumentation_mode` defaulting to `default` when only `granularity` is
-set. When neither is set the collector omits `--mode` entirely and Proton
-keeps its own default.
+set. `backend_mode` renders into that same argument, which is why the two
+cannot be combined. When none of them is set the collector omits `--mode`
+entirely and Proton keeps its own default.
+
+Both intra-kernel knobs are **`mode: env` knobs today**, whatever the recipe
+says. Triton 3.7.1's Proton front-end parses `-m/--mode` and then calls
+`start()` without it, so the value the CLI wrap renders is dropped on the
+floor; upstream `main` forwards it. Under `mode: env` the payload reads
+`AORTA_PROTON_MODE` and passes it to `proton.start(mode=...)` itself, which
+is the only route by which either knob reaches Proton on a released Triton.
+`granularity` does not survive even that route on Triton 3.7.1: the rendered
+`default:granularity=warp` is rejected inside
+`libproton.exit_instrumented_op` (the bare `default` and other knobs such as
+`buffer_size` are accepted, and the typed
+`triton.profiler.mode.Default(granularity="warp")` object works). Warp
+granularity is that backend's default on that version anyway, so the
+`amd-instrumentation` example sets `instrumentation_mode` and omits
+`granularity` deliberately.
 
 ### Proton attach modes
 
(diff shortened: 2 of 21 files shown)


amd-vivekag:
Superseded by #416, which is the identical branch pushed to `ROCm/aorta` instead of my fork.

`pytest (GPU, MI350)` is gated on `head.repo.full_name == github.repository`, so it reports `skipping` for any fork branch — by design, since untrusted fork code must not reach the self-hosted runner. This change adds the two regression tests for the empty-capture bug and fixes the path filter that was omitting them, so letting the GPU suite actually gate it matters more than the PR number. Every recent PR here is same-repo anyway.

PR #423, opened 2026-09-01
feat(cia): add the Cluster Intelligence Agents

aviator19941:
First of three PRs porting the Cluster Intelligence Agents from the internal
`aorta_llm` repository. Plan and findings: `docs/plans/cia-agents-port.md`.

**This PR is the agents only.** No chat code is touched. Launch puts a workload
on a node, Watch reads the log it produces, Autopsy classifies the bundle and
cites what it read — all usable from a script or CI on a base install plus the
new `[cia]` extra.

## What changed on the way in

The port is not a file copy; four things had to change because the destination
is a public repository rather than one machine.

**One category vocabulary.** `aorta.agent` already defined `AUTOPSY_CATEGORIES`,
and the agents' list turned out to be a strict superset. The three
evidence-only categories (`gpu_race`, `numeric_silent`, `tooling_gap`) join the
shared set and the second list is gone. The two loops stay separate — one
searches for a mitigation, the other explains a cause — but a verdict now means
the same thing whichever produced it. The router's prompt is generated from the
set and its answer validated against it, so neither can drift from the other.

**Autopsy says when it classified without its router.** Its adapters reach a
real verdict on their own, so losing the reasoning step is invisible in the
category — you get a plausible answer either way. That already went unnoticed
once: after the agents moved from subprocesses to function calls, nothing set
the router's endpoint and correct-looking verdicts kept arriving for days. It
is now a warning *and* a recorded tooling gap, because a log line is gone by
the time anyone reads the bundle.

**No site's layout is a default.** `search_roots()` guessed `/apps/$USER`, which
exists at exactly one site. The home directory is the only guess left, and
`CIA_SEARCH_ROOTS` covers everything else.

**Launch is reached through one function.** Slurm is still the only backend, but
a scheduler-less one is now a branch in one place rather than an edit at every
call site.

Dropped: three argparse console scripts (aorta's Click CLI supersedes them) and
the standalone taxonomy schema.

## Tests

43 for the agents, plus 2 added to `tests/cli/test_chat_boundaries.py`. Two of
them are the point rather than coverage:

- `test_settings_neutral.py` fails if any absolute path, cluster hostname or
  partition name reappears in a shipped default.
- `test_router_degraded.py` fails if Autopsy ever again classifies without its
  router and says nothing.

`test_importing_the_agents_does_not_require_the_chat_extras` measures in a fresh
interpreter that Launch, Watch and Autopsy pull in no langchain, Chainlit,
chromadb or torch — the property that justifies `src/aorta/cia/` rather than
`src/aorta/chat/agents/`.

Full suite: 5,683 passing. The 48 failures are all pre-existing — I confirmed
each fails identically on the base commit `ca0e61f` (missing optional extras and
an absent `pytest-timeout` plugin).

## Known debt, deliberate

Watch and Autopsy reach a model through DSPy with their own configuration,
which is a second LLM config surface in a repo that deliberately unified on the
shared chat provider layer (Decision 7a). Converging them is a larger change
than the port, so it is deferred and written down as finding F2 in the plan —
along with the reason the degraded-router warning above belongs in this PR
regardless of when that lands.

## Next

PR 2 brings the tools and the LLM tool selection; PR 3 the streamed reasoning
and the three-part answer.


src/aorta/cia/autopsy/probe.py
```diff
@@ -0,0 +1,98 @@
...
+def _ssh(node: str, cmd: str, background: bool = False) -> subprocess.CompletedProcess | None:
+    full_cmd = ["ssh", "-o", "StrictHostKeyChecking=no", "-o", "ConnectTimeout=15",
+                f"root@{node}", cmd + (" &" if background else "")]
+    if background:
+        subprocess.Popen(full_cmd)
+        return None
+    return subprocess.run(full_cmd, capture_output=True, text=True, timeout=60)
+
+
+def run_aorta_probe(bundle_root: Path, job: JobRecord, head_node: str = "149.28.124.225") -> Path | None:
```
amd-vivekag: **[High]** A site's IP address is shipped as the default head node.

This is the exact thing F3 and `test_settings_neutral.py` exist to prevent, and the guard misses it: its hostname regex needs hyphens, so a bare IP passes. Two more survived the scrub for the same reason — `launch/job.py:35` (`e.g. 149.28.124.225`) and `cia/llm.py:32` (`the agent is on (chi2878)`). All three are in a public repo.

Fix: default to `""` and take the host from the caller / `CIA_SSH_HOST`, scrub the two comments, and widen the guard (see my note on the test).

```suggestion
def run_aorta_probe(bundle_root: Path, job: JobRecord, head_node: str = "") -> Path | None:
```

src/aorta/cia/autopsy/probe.py
```diff
@@ -0,0 +1,98 @@
...
+    # Find recipe and sidecar on the node
+    find = _ssh(
+        head_node,
+        f"ssh -o ConnectTimeout=10 {job.node} "
+        f"'find /root /mnt -name Residual-NaN-Repro.yaml 2>/dev/null | head -1; "
+        f"find /root /mnt -name Residual-NaN-Repro-sidecar.json 2>/dev/null | head -1'"
+    )
+    lines = (find.stdout + find.stderr).strip().splitlines() if find else []
+    recipe = next((l.strip() for l in lines if "Residual-NaN-Repro.yaml" in l and "smoke" not in l), "")
+    sidecar = next((l.strip() for l in lines if "sidecar.json" in l), "")
```
amd-vivekag: **[High]** The escalation path runs a hardcoded recipe, not the user's job.

It searches `/root` and `/mnt` for `Residual-NaN-Repro.yaml` and sweeps that, whatever the job under autopsy actually was. `_ssh` also hardcodes `root@`, ignoring the `CIA_SSH_USER` that `launch/cluster.py` established.

This is the failure the plan rejects in §1 — "a canned tool is worse than an absent one, because the model picks it and silently substitutes the demo". Here it substitutes the demo inside a verdict.

Fix: derive the recipe from `job.recipe` / the bundle manifest and refuse to escalate when it cannot be resolved, or drop the escalation from this PR until it is general.

src/aorta/cia/llm.py
```diff
@@ -0,0 +1,52 @@
...
+        cache=False,
+    )
+    dspy.configure(lm=lm)
+    _configured = True
+
+
+def ensure_configured(**kwargs) -> None:
+    """Call configure_dspy with kwargs only if not already configured."""
+    if not _configured:
+        configure_dspy(**kwargs)
```
amd-vivekag: **[High]** `ensure_configured` silently discards Autopsy's model choice.

`TriageRouter.__init__` asks for `model="claude-sonnet-4-6", max_tokens=2048`, but by then `LogFinder`/`LogWatcher` in `poll_jobs` have already set `_configured`, so the router runs on haiku at 1024 tokens. `configure_dspy` also lets `LITELLM_MODEL` override the caller's explicit `model`, so even a first call loses it.

This is the F2 failure class again: a wrong-but-plausible verdict with nothing saying the reasoning step was downgraded.

Fix: give each module its own LM instead of a global one — build the `dspy.LM` per module and `self.react.set_lm(lm)` (or wrap calls in `dspy.context(lm=...)`), and let the explicit `model` argument win over `LITELLM_MODEL`.

src/aorta/cia/launch/discovery.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def check_gpu_arch(host: str, node: str) -> dict[str, Any]:
+    """Get GPU arch and count for node, preferring scheduler gres over rocm-smi.
+
+    Many clusters refuse direct SSH to compute nodes, so Slurm's gres/features
+    are the only reliable source; rocm-smi is a fallback for when this process
+    is already running on the target node.
+    """
+    out = run_probe(host, f"sinfo -N -n {node} --noheader -o '%G %f' 2>/dev/null | head -3")
```
amd-vivekag: **[High]** `node` is model-supplied and lands in `shell=True` unquoted.

`check_gpu_arch` is a ReAct tool, so the model chooses `node`; `run_probe` runs it through `subprocess.run(..., shell=True)`. A returned value like `n1; curl …` executes. `read_cluster_configs` and `check_aorta_install` interpolate `search_roots()` the same way.

Fix: quote it — and prefer `node_exists`-style argv lists over `shell=True` for probes that take an identifier.

```suggestion
    out = run_probe(host, f"sinfo -N -n {shlex.quote(node)} --noheader -o '%G %f' 2>/dev/null | head -3")
```

src/aorta/cia/launch/discovery.py
```diff
@@ -0,0 +1,143 @@
...
+    if not out.strip() or "ERROR" in out:
+        out = run_probe(host, "rocm-smi --showproductname 2>/dev/null | grep -E 'Card Series|Card Model'")
+
+    count = 0
+    gres = re.search(r"gpu:[^:\s]*:(\d+)", out)
+    if gres:
+        count = int(gres.group(1))
+    elif "Card Series" in out:
+        count = out.count("Card Series")
+
```
amd-vivekag, line 119: **[Medium]** Every GPU that is not an MI25x is reported as gfx942.

When the probe returns nothing or `ERROR`, this still answers `gfx942` with no "unknown". MI300 happens to be right; MI355X (gfx950), Navi and an empty probe are all confidently wrong, and the planner has no way to tell a guess from a reading. `%f` in the `sinfo` query already carries the real arch string on most clusters.

Fix: parse the arch out of the features field, and return `""` when nothing was read so `uncertain_fields` can carry it.

src/aorta/cia/launch/discovery.py
```diff
@@ -0,0 +1,143 @@
...
+def read_cluster_configs(host: str) -> str:
+    """Read existing job scripts and scheduler config to learn how jobs run here."""
+    roots = " ".join(search_roots()) or "~"
+    out = run_probe(host, (
+        rf"find {roots} -maxdepth 4 \( -name '*.sbatch' -o -name '*.slurm' -o -name '*.sh' \) "
+        "2>/dev/null | head -8 | xargs head -25 2>/dev/null; "
+        "echo '--scheduler config--'; "
+        "scontrol show config 2>/dev/null | grep -iE 'ClusterName|SchedulerType|MaxJobCount'"
+    ))
+    return out[:1500]
```
amd-vivekag: **[Medium]** This ships the contents of the user's home-directory shell scripts to the LLM.

`find $HOME -maxdepth 4 -name '*.sh' | head -8 | xargs head -25` reads the first 25 lines of whatever it finds. Personal `*.sh` files are where `export …_TOKEN=` and `export …_API_KEY=` live, and those first lines are exactly where they sit. The prompt reaches whatever `LITELLM_API_BASE` points at.

Fix: send filenames and the `#SBATCH` directive lines only, and drop any line matching a secret-ish assignment before it goes into the prompt.

src/aorta/cia/launch/cluster.py
```diff
@@ -0,0 +1,252 @@
...
+        f"#SBATCH --output={log_path}",
+        f"#SBATCH --error={log_path}",
+        "#SBATCH --nodes=1",
+        f"#SBATCH --time={os.environ.get('CIA_TIME_LIMIT', '04:00:00')}",
+    ]
+    partition = os.environ.get("CIA_PARTITION", "")
+    if partition:
+        directives.append(f"#SBATCH --partition={partition}")
+    if node and node_exists(node):
+        directives.append(f"#SBATCH --nodelist={node}")
```
amd-vivekag: **[Medium]** A requested node is silently dropped when `node_exists` says no.

`node_exists` returns `False` both when the node is unknown *and* when Slurm is simply unreachable or `scontrol` times out. Either way the `--nodelist` directive vanishes with no warning and the job runs somewhere the caller did not ask for — which for a repro pinned to one machine means the result is about a different machine.

Fix: log a warning when a caller-supplied `node` is dropped, so the substitution is visible in the job record.

src/aorta/cia/watch/poll.py
```diff
@@ -0,0 +1,297 @@
...
+                    new_parts.append(f"=== {p.name} ===\n{text}")
+                    total_new += len(text)
+            save_cursors(job_dir, cursors)
+
+            if not new_parts:
+                continue
+
+            new_content = "\n\n".join(new_parts)
+            total_bytes = sum(cursors.values())
+            job_ctx = f"{job_context} elapsed_sec={int(time.time())} total_bytes_seen={total_bytes}"
```
amd-vivekag: **[Medium]** `elapsed_sec` is the Unix epoch, not elapsed time.

`int(time.time())` is ~1.7e9 here. The model is told the job has been running for 1.7 billion seconds on every poll, which is the opposite of the signal "stalled vs still starting" needs.

Fix: measure from the job's launch, e.g. `int(time.time() - job_started)`.

src/aorta/cia/watch/poll.py
```diff
@@ -0,0 +1,297 @@
...
+                    job_context=job_context,
+                    scheduler=job.scheduler,
+                    scheduler_job_id=job.scheduler_job_id,
+                    head_node=getattr(job, "head_node", ""),
+                )
+                # Also include explicit log_path from job record
+                explicit = Path(job.log_path) if job.log_path else None
+                if explicit and explicit.exists() and explicit not in found:
+                    found.insert(0, explicit)
+                job.watch_files = [str(p) for p in found]
```
amd-vivekag: **[Medium]** Discovered watch files are never persisted, so the finder re-runs every round.

`scan_active_jobs` re-reads `job.json` from disk each round, so this assignment is discarded. When discovery falls through to step 3 in `LogFinder.find`, that is a DSPy call plus an `rglob` per job per poll — every 30–60s for the life of the job.

Fix: `write_job_json(job, jobs_root)` after setting `watch_files`.

src/aorta/cia/watch/poll.py
```diff
@@ -0,0 +1,297 @@
...
+                fh.write(json.dumps(ev) + "\n")
+
+            if should_alert(healthy, confidence, confidence_threshold):
+                print(f"[watch] {job.job_id}: ALERT {signal} — triggering autopsy")
+                from aorta.cia.watch.bundle_writer import write_bundle
+                from aorta.cia.watch.trigger import trigger_autopsy
+                bundle = write_bundle(job, job_dir, evidence or new_content[:4000], signal)
+                trigger_autopsy(bundle, job, jobs_root)
+                break  # one alert per job per session
+
```
amd-vivekag, line 289: **[Medium]** This `break` leaves the job loop, and the comment's guarantee is not implemented.

Two separate problems. The `break` exits `for job in active`, so one job alerting skips every remaining job that round. And nothing records that a job has alerted, so the same job alerts and re-triggers autopsy again next round as soon as new bytes arrive — "one alert per job per session" is not enforced anywhere.

Fix: keep an `alerted: set[str]` across rounds, skip jobs in it at the top of the loop, and `continue` instead of `break`.

src/aorta/cia/autopsy/router.py
```diff
@@ -0,0 +1,178 @@
...
+        }
+    except Exception as e:
+        return {"error": str(e)}
+
+
+def read_evidence_file(uri: str, bundle_root: str) -> str:
+    """Read a specific evidence file from the bundle by its URI.
+    Returns up to 200 lines of the file content."""
+    try:
+        p = Path(bundle_root) / uri
```
amd-vivekag: **[Medium]** `read_evidence_file` escapes the bundle.

`uri` comes from the evidence list and from the model's own tool call, and `Path(bundle_root) / "../../../etc/passwd"` resolves outside `bundle_root`. The result goes straight into the router's context and can be quoted into the rationale.

Fix: resolve and contain it.

```suggestion
        p = (Path(bundle_root) / uri).resolve()
        if not p.is_relative_to(Path(bundle_root).resolve()):
            return f"[refused: {uri} is outside the bundle]"
```

src/aorta/cia/autopsy/adapters/stderr_watch.py
```diff
@@ -0,0 +1,103 @@
...
+
+from aorta.cia.autopsy.adapters.base import AdapterArtifact, BundleContext
+
+NAN_PATTERNS = [
+    re.compile(r"\bloss=nan\b", re.I),
+    re.compile(r"\bloss.*\bnan\b", re.I),
+    re.compile(r"non-?finite", re.I),
+    re.compile(r"NaN detected", re.I),
+    re.compile(r"residual.*nan", re.I),
+    re.compile(r"numeric_silent", re.I),
```
amd-vivekag: **[Medium]** These patterns fire on text that says the opposite.

`numeric_silent` matches the category name itself, so a log echoing an autopsy verdict or a mitigation name reads as a fresh NaN. `non-?finite` matches "no non-finite values found", and `\bloss.*\bnan\b` matches "checking loss for nan". Each false hit is a Watch alert plus an autopsy run.

Fix: anchor on a value rather than the word — require an assignment or comparison context (e.g. `loss\s*[=:]\s*-?nan`) — and drop the `numeric_silent` pattern.

src/aorta/cia/llm.py
```diff
@@ -0,0 +1,52 @@
...
+import os
+import ssl
+import certifi
+
+import dspy
+
+# LiteLLM fetches a remote price map on import and fails on corporate TLS.
+# Point it at the certifi bundle so imports are clean.
+os.environ.setdefault("SSL_CERT_FILE", certifi.where())
+os.environ.setdefault("REQUESTS_CA_BUNDLE", certifi.where())
```
amd-vivekag: **[Medium]** Importing this module rewrites TLS trust for the whole process, and `certifi` is undeclared.

`SSL_CERT_FILE` is process-global, so anything imported after this — including unrelated aorta code and the chat provider layer — verifies against certifi rather than the system store. At a site whose corporate CA is in the system bundle but not certifi, this breaks TLS that previously worked. `certifi` is also not in `dependencies` or the `cia` extra; it works today only because litellm happens to pull it.

Fix: set this inside `configure_dspy` rather than at import, and declare `certifi` in the `cia` extra.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

The port is careful where it says it is: the taxonomy unification, the degraded-router tests and `test_settings_neutral.py` are all doing real work, and the reasoning written into the docstrings is genuinely good. The problems are in the places the port did not revisit. `autopsy/probe.py` came across essentially untouched and carries a site IP, a hardcoded `root@` login and a hardcoded `Residual-NaN-Repro` recipe — the canned-demo pattern the plan explicitly rejects, sitting inside a verdict. Separately, `ensure_configured` is first-writer-wins, so Autopsy's request for a bigger model is discarded by whichever agent configured DSPy first, which is the same silent-downgrade class as F2.

**Findings:** 4 High · 11 Medium · 1 Low

**High**

- `autopsy/probe.py:24` — `head_node = "149.28.124.225"` is a shipped default; `chi2878` and the same IP also survive in comments in `cia/llm.py:32` and `launch/job.py:35`. `test_settings_neutral.py` catches none of the three.
- `autopsy/probe.py:33-41` — escalation sweeps a hardcoded recipe found under `/root` / `/mnt` rather than the job's own, and SSHes as `root@` regardless of `CIA_SSH_USER`.
- `cia/llm.py:49` — `ensure_configured` no-ops after the first caller, so `TriageRouter`'s `claude-sonnet-4-6` / 2048 tokens is silently dropped; `LITELLM_MODEL` also overrides an explicit `model`.
- `launch/discovery.py:64` — model-supplied `node` interpolated unquoted into `subprocess.run(shell=True)`.

**Two things that are not line-level**

The plan (§5) says to run these three sequentially — "merge one, branch the next off the updated base — rather than stacking … parallel review would mean reviewing against code that can still move" — and they are stacked. I reviewed all three anyway, but any change here moves #424 and #425 under review.

`watch_config.yaml` is loaded via `Path(__file__).parent`. Please confirm it is covered by `[tool.setuptools.package-data]`; if it is not, an installed wheel silently falls back to `{}` and the built-in defaults, with no error.

pyproject.toml
```diff
@@ -120,6 +120,14 @@ hw-queue-profiling = [
 ]
 
 # For aorta agent closed-loop (LiteLLM proposer)
+# The Cluster Intelligence Agents. DSPy is what Watch and Autopsy reach a model
+# through today; the shared chat provider layer is where that belongs, so this
+# extra is expected to shrink rather than grow -- see
+# docs/plans/cia-agents-port.md, finding F2.
+cia = [
+    "dspy-ai>=2.5; python_version >= '3.11'",
```
amd-vivekag: **[Medium]** On Python 3.10 this extra installs nothing and the failure surfaces as an ImportError.

`requires-python` is `>=3.10`, so `pip install amd-aorta[cia]` succeeds on 3.10 having installed no dspy, and the first `import aorta.cia.launch.planner` dies with `ModuleNotFoundError: dspy`.

Fix: keep the marker but make the requirement explicit at import — a `try/except ImportError` in `cia/llm.py` raising "the [cia] extra needs Python >= 3.11" — or drop the marker and let pip refuse the install.

tests/cia/test_settings_neutral.py
```diff
@@ -0,0 +1,100 @@
...
+    # ``f"/apps/{os.environ.get('USER')}"``, where the next character is a
+    # brace. Requiring a letter there would have missed it.
+    (re.compile(r"/apps/"), "an absolute path under /apps"),
+    (re.compile(r"/home/"), "an absolute path under /home"),
+    # cv350-rck-g03 and the like.
+    (re.compile(r"\b[a-z]{2,}\d{2,}-[a-z]{2,}\d?-[a-z]\d{2}"), "a cluster hostname"),
+    # Which partition is hardcoded matters less than that one is: an installed
+    # default no account can submit to fails the same way whatever it is named.
+    (re.compile(r"--partition=[A-Za-z]"), "a hardcoded Slurm partition"),
+    (re.compile(r'"(interactive|meta\d+)"'), "a Slurm partition name"),
```
amd-vivekag, line 54: **[Medium]** The guard misses the values actually in this diff.

`_SITE_SHAPED` requires hyphens for a hostname and has no IP pattern, so `149.28.124.225` (`autopsy/probe.py:24`, `launch/job.py:35`) and `chi2878` (`cia/llm.py:32`) all pass. `_source_files` also scans `src/aorta/cia` only, which means the tools PR #424 adds under `src/aorta/chat/tools/` are unguarded — and one of them defaults to `/opt/rocm-7.0.2.2/lib/llvm/bin`.

Fix: add a dotted-quad pattern and a bare `[a-z]{2,}\d{3,}` hostname shape, and widen `_source_files` to cover the chat tool modules too.

src/aorta/cia/watch/watch_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+watch:
+  model: claude-haiku-4-5        # fast + cheap for frequent polling
+  confidence_threshold: 0.70
+
+  # Adaptive poll interval — Watch self-tunes based on observed log activity.
+  # First poll: Haiku estimates step cadence, sets interval to ~2x step time.
+  # Backs off when log is quiet; re-arms when activity spikes.
+  poll_interval_sec: 60          # starting interval; updated dynamically
+  poll_interval_min_sec: 15      # never poll more often than this
+  poll_interval_max_sec: 600     # never wait longer than 10 minutes
```
amd-vivekag: **[Low]** This documents behaviour that does not exist.

`poll.py` reads `poll_interval_sec` and `confidence_threshold` and nothing else. `model`, `poll_interval_min_sec` and `poll_interval_max_sec` have no reader, and the comment promising "Haiku estimates step cadence, sets interval to ~2x step time" and "backs off when the log is quiet" describes an adaptive loop that is not implemented — the interval is a constant.

Fix: delete the three unread keys and the adaptive-polling comment, or open an issue and mark them `# not yet implemented`.

amd-vivekag (commented):

**Verdict: still fix-before-ship**

The first review was against `docs/plans/cia-agents-port.md`. This pass is against the aorta_llm integration contract that already landed in this repo: `docs/chat/{README,providers,configuration,redaction,extending}.md` and aorta_llm `ARCHITECTURE.md`.

The agents are useful headless — that part of F5 is right. What is not right is standing up a **second, unredacted, differently-configured LLM path** next to a package whose entire integration was "one provider layer, one config file, one redaction gate, tools bound to a root." A user who has `aorta chat` working has not configured CIA, and CIA will not notice.

**New findings:** 2 High · 1 Medium

**High**
- `cia/llm.py` — DSPy + `LITELLM_*` defaults, ignoring `get_chat_llm()` / `AORTA_CHAT_*` / `chat.toml` (`docs/chat/providers.md`, `docs/chat/configuration.md`).
- `cia/watch/watcher.py` — log tails and home-directory script heads leave the machine without Decision 16 redaction (`docs/chat/redaction.md`). Same for Launch discovery and Autopsy router.

**Medium**
- `launch/confirm.py` is unused, so chat-driven Launch is autonomous despite the chat-vs-agent split in `docs/chat/README.md`.

`docs/chat/*` is also untouched in this stack: no `[cia]` extra, no CIA knobs in the settings table, no warning that Watch/Autopsy do not use the chat provider. That documentation gap belongs with whichever PR first makes CIA reachable from `aorta chat` (#424), but the two LLM surfaces originate here.

src/aorta/cia/llm.py
```diff
@@ -0,0 +1,52 @@
...
+
+    lm = dspy.LM(
+        model=f"openai/{resolved_model}",
+        api_base=resolved_base,
+        api_key=resolved_key,
+        max_tokens=max_tokens,
+        cache=False,
+    )
+    dspy.configure(lm=lm)
+    _configured = True
```
amd-vivekag, line 339: **[High]** This is a second LLM stack beside the one aorta_llm integration locked.

`docs/chat/providers.md` is explicit: every graph node obtains a model from `get_chat_llm()`, selected by `llm_provider` and configured through `~/.config/aorta/chat.toml` / `AORTA_CHAT_*`. No node, tool, or entry point is supposed to branch on a provider. Watch, Autopsy, Launch discovery, and the log finder all go through DSPy + `LITELLM_API_BASE` / `LITELLM_API_KEY` / `LITELLM_MODEL` instead.

The practical result is a user who already has chat working (`aorta chat config init --profile openai`) still has a silent Watch/Autopsy that talks to `localhost:4000` with key `dummy`. That is the F2 failure the port plan deferred, but it is also a violation of the integration contract this PR is landing next to — not just debt to track.

Fix: even if the DSPy *modules* stay for this PR, `configure_dspy` must read the same Settings the rest of chat uses (`settings.remote_llm_base_url` / `vllm_base_url`, the matching key, the matching model) and refuse to default to a local proxy. Do not invent a `LITELLM_*` surface.

src/aorta/cia/watch/watcher.py
```diff
@@ -0,0 +1,115 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+import dspy
+
+from aorta.cia.llm import ensure_configured
```
amd-vivekag, line 10: **[High]** Watch's outbound prompts skip Decision 16 redaction.

`docs/chat/redaction.md` and `_send()` in `graph/nodes.py` exist because remote providers receive retrieved chunks and tool output, and those are rewritten (paths and IPs) before they leave. The comment on `_send` is "every node sends through here rather than calling ainvoke directly, so the gate cannot be bypassed by a node added later."

This module (and Launch discovery / Autopsy router) never goes through `_send`. Polling ships log tails, job context, and `find $HOME … | xargs head` output to whatever `LITELLM_API_BASE` points at, unredacted. On a remote provider that is the exfiltration path the integration docs exist to close.

Fix: run Watch/Autopsy text through `aorta.chat.redaction.redact_text` before it is sent, or — better — drive those calls through `get_chat_llm` + `_send` so the gate is the same one. If cia must not import chat (F5), extract redaction to a shared helper both packages call.

src/aorta/cia/launch/confirm.py
```diff
@@ -0,0 +1,47 @@
...
+
+    return answers
+
+
+def proceed_gate(summary: str) -> bool:
+    """Always-show confirmation before executing any SSH command."""
+    if not sys.stdin.isatty():
+        # Non-interactive: print summary but proceed (CI/script mode)
+        print(f"\n[deploy] Auto-proceeding (non-interactive):\n{summary}")
+        return True
```
amd-vivekag: **[Medium]** The human-in-the-loop gate from the aorta_llm integration never runs.

`docs/chat/README.md` draws the line: `aorta chat` is conversational and watched; `aorta agent` is autonomous and writes artifacts. `confirm.py` is the only thing in this package that would keep Launch on the chat side of that line (`proceed_gate` before SSH, `confirm_gate` when the planner is unsure). Nothing imports it — I grepped. `launch()` in `__init__.py` goes straight to `submit_sbatch`.

A chat turn that reaches Launch therefore submits a cluster job with no confirmation, which is the agent shape, not the chat shape.

Fix: either call `proceed_gate` from `launch()` when stdin is a TTY (and refuse when it is not, unless an explicit `CIA_CONFIRM=never` is set), or delete this file so the next reader does not think the gate exists.

amd-vivekag:

aviator19941:
> Please rebase these onto `main` as the baseline — it now has the latest chatbot fixes, so using `main` should pick those up and avoid landing on a stale chat stack.
> 
> Happy to help if anything conflicts.

Yes thanks for reminding me, will do.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The remediation since the earlier review is substantial: site-specific defaults, canned-recipe escalation, per-module LM settings, discovery quoting, GPU architecture parsing, elapsed time, alert deduplication, TLS behavior, package data, and shared provider settings are materially improved. The remaining blockers are at the trust and delivery boundaries: Watch can permanently discard an unprocessed failure chunk, model-callable tools can escape their intended roots, probe values still reach remote shells, and several failure paths can be reported as clean or never escalate.

**Findings:** 7 High · 5 Medium · 1 Low

**Still open / partial from the last review**
- Discovered `watch_files` are still not persisted, so fallback discovery can repeat every poll.
- Containment is partial: the model can choose `bundle_root`, adapters bypass the resolver, and Watch exposes arbitrary path arguments.
- `[cia]` now installs on Python 3.10, but its shared chat configuration/redaction seam is not actually 3.10/base-install compatible.

I independently confirmed the current Copilot findings, including manifest traversal, sanitizer error states falling through to clean, unsafe probe/scp construction, unvalidated environment names, unrestricted Watch file reads, the taxonomy regression, malformed-matrix aborts, and the dropped `JobRecord`. I did not duplicate those inline.

- `pre-commit`: `end-of-file-fixer` modifies `src/aorta/cia/autopsy/adapters/__init__.py`.
- CPU Python 3.10–3.14 installs the slim test environment but collects `tests/cia`; DSPy is absent, causing collection failures and cascading failures across every matrix leg. Python 3.10 also cannot import the shared `tomllib`-based settings seam.
- Add a dedicated CIA test lane that installs `.[tests,cia]` (or explicitly exclude dependency-requiring CIA tests from the slim CPU lane), while retaining source-only dependency-boundary tests there. This is caused by this PR's test/dependency partition, not baseline noise.

**Architecture integration**
The long-term shape should move provider settings and redaction into a Python-3.10-compatible core seam shared by chat, agents, and CIA. Bind model tools to trusted job/bundle roots outside their callable schemas, commit log cursors only after successful processing, and dispatch Autopsy through a bounded persisted queue so one slow classification cannot stop every watched job.

src/aorta/cia/watch/poll.py
```diff
@@ -0,0 +1,347 @@
...
+            total_new = 0
+            for p_str in job.watch_files:
+                p = Path(p_str)
+                cursor = cursors.get(p_str, 0)
+                text, new_cursor = read_new_bytes(p, cursor)
+                cursors[p_str] = new_cursor
+                if text:
+                    new_parts.append(f"=== {p.name} ===\n{text}")
+                    total_new += len(text)
+            save_cursors(job_dir, cursors)
```
amd-vivekag: **[High]** A transient model failure permanently discards this log chunk.

The cursor is saved before `watcher.forward()`; if that call raises, the loop continues with the cursor already advanced, so a one-time NaN/OOM/fault is never assessed again.

Fix: commit candidate cursors only after assessment and event persistence succeed, with a bounded retry or visible dead-letter path.

src/aorta/cia/llm.py
```diff
@@ -0,0 +1,292 @@
...
+    found under a home directory, and Autopsy ships bundle evidence.
+
+    Returns *text* unchanged when the chat settings are unavailable, which is
+    the only honest thing to do -- but that is also why the gate below is not
+    the only protection: the probes were narrowed to stop collecting what
+    should not be sent in the first place.
+    """
+    try:
+        from aorta.chat.redaction import redact_text
+    except Exception:  # pragma: no cover - depends on what is installed
```
amd-vivekag: **[High]** The advertised `[cia]`-only path silently disables redaction.

`aorta.chat.redaction` imports `aorta.chat.config`, but `[cia]` does not declare `pydantic-settings` and Python 3.10 also lacks `tomllib`; this broad fallback therefore sends Watch/Autopsy evidence unchanged.

Fix: extract settings/redaction into a core dependency of `[cia]` (with a 3.10 TOML backport), or refuse remote LM use when the redaction gate cannot load.

src/aorta/cia/autopsy/router.py
```diff
@@ -0,0 +1,209 @@
...
+    into the router's context and could be quoted into the rationale, which is
+    written to the report and sent to a model.
+
+    The policy itself lives with BundleContext, which is what defines a bundle;
+    this is the same check the adapters get, spelled for a string root.
+    """
+    return _resolve_in_bundle(Path(bundle_root), uri)
+
+
+def read_evidence_file(uri: str, bundle_root: str) -> str:
```
amd-vivekag, line 84: **[High]** The model still chooses the containment root.

`read_evidence_file(uri, bundle_root)` is a ReAct tool, so a call with `bundle_root="/"` makes any host file appear contained even though adapters now use a trusted `BundleContext`.

Fix: bind the resolved bundle root in a closure/tool object outside the model-callable schema and expose only a logical evidence URI.

src/aorta/cia/watch/poll.py
```diff
@@ -0,0 +1,418 @@
...
+
+            # Assessed, and the event is on disk. Only now is this chunk done.
+            save_cursors(job_dir, cursors)
+
+            if should_alert(healthy, confidence, confidence_threshold):
+                print(f"[watch] {job.job_id}: ALERT {signal} — triggering autopsy")
+                from aorta.cia.watch.bundle_writer import write_bundle
+                from aorta.cia.watch.trigger import trigger_autopsy
+                bundle = write_bundle(job, job_dir, evidence or new_content[:4000], signal)
+                trigger_autopsy(bundle, job, jobs_root)
```
amd-vivekag, line 282: **[High]** One Autopsy can stop monitoring every other active job for hours.

The `JobRecord` fix makes the four-hour production escalation reachable, but this synchronous call still runs inside Watch's serial job loop.

Fix: persist alert/in-flight state, enqueue Autopsy on a bounded worker, and continue polling other jobs.

src/aorta/cia/launch/planner.py
```diff
@@ -0,0 +1,221 @@
...
+        beside = run_probe(host, rf"ls {shlex.quote(str(Path(recipe_path).parent))}/*sidecar*.json 2>/dev/null")
+        candidates = [l.strip() for l in beside.splitlines() if l.strip().endswith(".json")]
+    if not candidates:
+        anywhere = run_probe(host, rf"find {roots} -maxdepth 6 -name '*sidecar*.json' 2>/dev/null | head -5")
+        candidates = [l.strip() for l in anywhere.splitlines() if l.strip().endswith(".json")]
+
+    ticket = stem.removesuffix("-smoke")
+    sidecar_path = next((c for c in candidates if ticket and Path(c).name.startswith(ticket)), "")
+    if not sidecar_path and candidates:
+        sidecar_path = candidates[0]
```
amd-vivekag, line 106: **[Medium]** Falling back to the first sidecar can pair a recipe with an unrelated environment.

The planner prompt says that exact pairing fails with `UnknownEnvironmentError`, so this fallback manufactures the failure it warns against.

Fix: return an empty sidecar and require confirmation when no owned/ticket-matching sidecar is found.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The latest remediation fixes most of the previous review correctly: manifest paths, sanitizer failure states, probe copying, environment-name validation, Python 3.10 environment settings, Watch file-tool scope, category eligibility, malformed matrices, dropped `JobRecord`, cursor commit timing, and headless redaction are materially improved. I did not approve because two High issues remain and required CI is red.

**Findings:** 2 High · 2 Medium

**Still open / introduced by the fixes**
- **High:** Autopsy still lets the ReAct model choose `bundle_root`, so its evidence tool can redefine `/` as the trusted root and read arbitrary host files.
- **High:** passing the `JobRecord` made production escalation reachable, but it now runs synchronously inside Watch's serial job loop; one low-confidence alert can stop monitoring every other job for up to four hours.
- **Medium:** fallback log discovery is still not persisted, and planner sidecar discovery still selects the first unrelated sidecar when no owned match exists.

Prior reconciliation: approximately 27 prior inline findings are fixed correctly; containment is still partial and the `watch_files` persistence finding remains open. All prior Copilot findings were rechecked rather than assumed fixed.

**CI**
- `pre-commit` still fails because `src/aorta/cia/autopsy/adapters/__init__.py` has no final newline; `git diff --check` also shows CRLF/whitespace across `autopsy/reporter.py`.
- CPU Python 3.10–3.14 collects `tests/cia` without installing `[cia]`, producing 18 collection errors on py3.12 and similar failures in every matrix leg. Chat CI is green.
- Add a CIA-enabled test lane or install `[cia]` for the tests that import DSPy, while retaining source-only optional-boundary tests in the slim CPU lane.

**Architecture integration:** the shared provider/redaction direction is now substantially better, but trust roots must be bound outside model-callable schemas and long-running Autopsy work needs a persisted bounded queue. Until those ownership boundaries and CI are fixed, integration is not solid enough to approve.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The remediation is substantial, but the integration is still unsafe at the filesystem, provider, classification, and scheduling boundaries. Autopsy still lets the model redefine its evidence root, Watch accepts model-selected paths outside the job, a persisted output path reaches a remote shell unquoted, the shared provider loses its backend identity, and generic workload failures are classified as numeric corruption. Autopsy escalation also remains synchronous inside Watch's serial loop.

**Findings:** 6 High · 3 Medium

**Still open from prior reviews**
- **High:** model-controlled Autopsy `bundle_root`.
- **High:** synchronous escalation can stop monitoring all jobs for up to four hours.
- **Medium:** discovered `watch_files` are not persisted.
- **Medium:** unrelated-sidecar fallback remains.

**CI:** required CPU tests and pre-commit are red; CPU collects `tests/cia` without `[cia]`, and the EOF hook still modifies `autopsy/adapters/__init__.py`. The branch is also behind `main`.

**Architecture integration:** shared taxonomy, core redaction, and per-module model budgets are good improvements. The integration is not solid until trust roots are bound outside model schemas, provider identity is preserved end-to-end, classifications require category-specific evidence, and long-running Autopsy work moves behind a bounded durable queue.

src/aorta/cia/autopsy/router.py
```diff
@@ -0,0 +1,209 @@
...
+    """Read a specific evidence file from the bundle by its URI.
+    Returns up to 200 lines of the file content."""
+    try:
```
amd-vivekag: **[High]** The model still controls the evidence trust root.

Calling this tool with `bundle_root="/"` makes `/etc/passwd` pass the containment check.

Fix: bind the resolved bundle root in a closure/context and expose only `uri` in the ReAct tool schema.

src/aorta/cia/watch/poll.py
amd-vivekag: **[High]** Autopsy blocks the serial Watch loop for every job.

A low-confidence verdict can enter the four-hour production sweep here, so no later job is monitored.

Fix: persist the alert and enqueue Autopsy into a bounded durable worker queue with timeout/retry state.

src/aorta/cia/watch/log_finder.py
```diff
@@ -0,0 +1,259 @@
...
+        if not listing.strip():
+            return by_ext[: self.max_files]
+
+        try:
+            pred = self._discovery(
+                job_dir_listing=listing,
+                job_context=job_context or f"job_dir={job_dir}",
+            )
+            result = []
+            for p_str in (pred.relevant_files or []):
```
amd-vivekag, line 226: **[High]** A model-selected absolute path is accepted as a log with no root check.

The next stage reads it and grants its parent as a Watch tool root, so a hallucinated `/etc/passwd` escapes the job boundary.

Fix: resolve every returned path under trusted job/search roots before accepting it, including symlink checks.

src/aorta/cia/autopsy/probe.py
```diff
@@ -0,0 +1,180 @@
...
+        )
+        return None
+
+    sweep = ["nohup aorta sweep run", f"--recipe {shlex.quote(recipe)}"]
+    if sidecar and check_recipe_mode(recipe).get("mode") != "sanitizer":
+        sweep.append(f"--mitigations-file {shlex.quote(sidecar)}")
+    sweep.append(f"--output {shlex.quote(aorta_output)}")
+
+    cmd = (
+        f"ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no {shlex.quote(job.node)} "
```
amd-vivekag, line 127: **[High]** `aorta_output` is still unquoted in the remote-shell redirection.

This field comes from the plan/job record; shell characters here execute on the compute node even though the `--output` argument above is quoted.

Fix: quote the complete log path independently, or remove the nested shell construction.

src/aorta/cia/llm.py
```diff
@@ -0,0 +1,365 @@
...
+    if resolved is None and not api_base:
+        raise ProviderNotConfigured(
+            "The agents reach a model through the same configuration as the "
+            "rest of aorta chat, and none is available. Run `aorta chat config "
+            "init`, or set AORTA_CHAT_VLLM_BASE_URL. On Python 3.10 the profile "
+            "file cannot be read at all -- only AORTA_CHAT_* is honoured there."
+        )
+    settings_base, settings_key, settings_model = resolved or ("", "", "")
+
+    _use_certifi_bundle()
```
amd-vivekag, line 337: **[High]** Shared settings lose the selected provider identity.

A `litellm`/Anthropic profile is reduced to URL/key/model and rebuilt as `openai/<model>`, routing a Claude model through the wrong backend.

Fix: preserve and map `llm_provider`, or adapt the shared provider object instead of reconstructing it.

src/aorta/cia/autopsy/adapters/aorta_matrix.py
```diff
@@ -0,0 +1,303 @@
...
+    for cell in cells:
+        name = cell.get("name", "")
+        mitigations = set(cell.get("mitigations") or [])
+        failure_rate = float(cell.get("failure_rate") or 0.0)
+        failed = int(cell.get("failed_count") or 0)
+        hints = " ".join(cell.get("failure_hints") or [])
+        exit_counts = cell.get("exit_status_counts") or {}
+        workload_failed = int(exit_counts.get("workload_failed") or 0)
+
+        is_repro = failure_rate > 0 or failed > 0 or workload_failed > 0
```
amd-vivekag, line 203: **[High]** Every failed workload cell is treated as numeric corruption.

OOMs, launch failures, and unrelated nonzero exits all satisfy this predicate; the `NAN_HINTS` check below has no effect.

Fix: require structured numeric evidence before emitting `numeric_silent`, and leave non-numeric failures unknown or explicitly categorized.

src/aorta/cia/watch/poll.py
```diff
@@ -0,0 +1,418 @@
...
+                    # sanitizer sweep is a source checkout whose test fixtures
+                    # are *designed* to read like failing logs. Watch then
+                    # faithfully diagnoses somebody's test data.
+                    continue
+                if declared is not None:
+                    # The job told us where it writes. Nothing discovery finds
+                    # can be more authoritative than that.
+                    job.watch_files = [str(declared)]
+                else:
+                    job.watch_files = [
```
amd-vivekag, line 539: **[Medium]** Discovered watch files are never persisted.

Each round reloads a fresh `JobRecord`, so fallback discovery and its model call repeat indefinitely.

Fix: atomically write `watch_files` back to `job.json` once resolved.

src/aorta/cia/launch/planner.py
```diff
@@ -0,0 +1,221 @@
...
+    if recipe_path:
+        beside = run_probe(host, rf"ls {shlex.quote(str(Path(recipe_path).parent))}/*sidecar*.json 2>/dev/null")
+        candidates = [l.strip() for l in beside.splitlines() if l.strip().endswith(".json")]
+    if not candidates:
+        anywhere = run_probe(host, rf"find {roots} -maxdepth 6 -name '*sidecar*.json' 2>/dev/null | head -5")
+        candidates = [l.strip() for l in anywhere.splitlines() if l.strip().endswith(".json")]
+
+    ticket = stem.removesuffix("-smoke")
+    sidecar_path = next((c for c in candidates if ticket and Path(c).name.startswith(ticket)), "")
+    if not sidecar_path and candidates:
```
amd-vivekag: **[Medium]** An unrelated sidecar is still selected when no owned match exists.

That makes a valid recipe fail against another recipe's environments.

Fix: return no sidecar and require confirmation when no candidate matches the recipe/ticket.

src/aorta/cia/llm.py
```diff
@@ -0,0 +1,365 @@
...
+    imports pydantic and stdlib and nothing from the chat extras, so this keeps
+    the agents runnable on a base install.
+
+    With *configured_only*, None also means "nothing here was actually set".
+    Every field has a default, so answering with one would silently outrank a
+    deployment that had configured the agents some other way.
+    """
+    global _warned_no_settings
+    try:
+        from aorta.chat.config import settings
```
amd-vivekag, line 128: **[Medium]** The advertised shared profile is unavailable on a base + `[cia]` install.

`aorta.chat.config` needs dependencies outside `[cia]`, so this falls back to environment-only settings even on Python 3.11+.

Fix: move profile parsing to a dependency-neutral core seam or include its direct dependencies and report the actual import failure.

src/aorta/cia/llm.py
```diff
@@ -0,0 +1,423 @@
...
+    a proxy -- both speak the OpenAI protocol whatever is behind them -- and
+    wrong for a litellm profile addressing a vendor directly, whose model names
+    carry their own vendor: ``anthropic/claude-3-5-sonnet`` became
+    ``openai/anthropic/claude-3-5-sonnet`` and went to the wrong backend.
+
+    A name that already names its vendor is left alone. Otherwise an endpoint
+    is what decides: with one, the request goes to something OpenAI-shaped; a
+    litellm profile without one is routed by litellm's own rules.
+    """
+    if "/" in model:
```
amd-vivekag: **[High]** A slash in a vLLM model is mistaken for a provider prefix.

The default is `deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct`; returning it unchanged bypasses the `openai/` routing required by the configured OpenAI-compatible vLLM endpoint.

Fix: decide `vllm`/`openai` endpoint routing before the generic slash check, and only preserve an existing prefix when it is a real LiteLLM provider prefix.

src/aorta/cia/autopsy/adapters/aorta_matrix.py
```diff
@@ -0,0 +1,381 @@
...
+import logging
+import re
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+from aorta.cia.autopsy.adapters.base import AdapterArtifact, BundleContext
+
+MITIGATION_ISOLATION = frozenset({"tf32_off", "deterministic"})
+NAN_HINTS = re.compile(r"nan|residual|non.?finite|inf", re.I)
```
amd-vivekag: **[High]** `inf` matches ordinary words such as `inference` and `infrastructure`.

Now that this regex gates `numeric_silent`, a failed inference cell beside a clean mitigation is still reported as numeric corruption at high confidence.

Fix: match complete numeric tokens, e.g. `\b(?:nan|inf(?:inity)?|non[- ]?finite)\b`, and add inference/infrastructure regressions.

src/aorta/cia/watch/poll.py
```diff
@@ -0,0 +1,642 @@
...
+    """The rounds themselves, so the pool above owns its own lifetime."""
+    while max_rounds is None or rounds < max_rounds:
+        rounds += 1
+        active = scan_active_jobs(jobs_root)
+
+        for job in active:
+            # Both halves matter: the set covers this process, the file covers
+            # a restart. Only the set existed, so a watcher that came back
+            # re-diagnosed what it had already paid for.
+            if job.job_id in alerted:
```
amd-vivekag: **[High]** A failed Autopsy is never retried while this Watch process stays alive.

The job is added to `alerted` after submission and this check runs before the persisted `failed` state; the retry test only passes by starting `poll_jobs` again.

Fix: clear the in-memory claim when the Future fails, or derive eligibility from tracked Future/state so `failed` can re-enter the bounded retry path in the same process.

tests/cia/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+# rather than skipping. Skipping is right: a base install genuinely cannot run
+# these, and that is a shipped configuration rather than a fault.
+#
+# Same shape as tests/chat/conftest.py, which skips itself when the chat extra
+# is absent. The import-boundary tests live in tests/cli/ for the same reason
+# they do there: they are pure AST and must run everywhere.
+CIA_EXTRA_INSTALLED = importlib.util.find_spec("dspy") is not None
+
+if not CIA_EXTRA_INSTALLED:  # pragma: no cover - exercised on a base install
+    collect_ignore_glob = ["test_*.py"]
```
amd-vivekag, line 29: **[High]** Current green CI skips every CIA test.

The CPU lane lacks `[cia]`, the chat lane does not run `tests/cia`, and no workflow installs the CIA extra, so none of the new security/concurrency regressions executes in CI.

Fix: add a CIA-enabled lane that installs `.[tests,cia]`, verifies DSPy imports, and runs `tests/cia` on the supported Python range.

src/aorta/cia/watch/poll.py
```diff
@@ -0,0 +1,642 @@
...
+    #: is diagnosed once per session however much more it goes on to write.
+    alerted: set[str] = set()
+    #: Consecutive assessment failures per job, so a chunk is retried rather
+    #: than dropped, and a job whose assessment always fails is eventually let
+    #: go rather than blocking its own progress for ever.
+    failures: dict[str, int] = {}
+
+    print(f"[watch] polling {jobs_root} every {interval}s")
+
+    pool = ThreadPoolExecutor(
```
amd-vivekag, line 418: **[Medium]** Limiting workers does not bound `ThreadPoolExecutor`'s task queue.

One alerting round can enqueue every job behind two four-hour workers; queued records can age past the five-hour lease before execution.

Fix: add a bounded submission capacity/backpressure policy and persist enough queue state for prompt restart recovery.

src/aorta/cia/llm.py
```diff
@@ -0,0 +1,423 @@
...
+        from aorta.chat.config import settings
+    except ImportError as exc:
+        if not _warned_no_settings:
+            _warned_no_settings = True
+            # The reason is reported, not assumed. This used to state the 3.10
+            # one whatever had actually happened, so an install failing because
+            # a package was missing was told it needed Python 3.11 -- on 3.13.
+            # A diagnostic that names the wrong cause is worse than none: it is
+            # followed.
+            if sys.version_info < (3, 11):
```
amd-vivekag: **[Medium]** The shared TOML profile remains unavailable on supported Python 3.10.

The extra now declares pydantic correctly, but this branch still degrades to environment-only settings solely because stdlib `tomllib` is absent.

Fix: use `tomli` on 3.10 through a dependency-neutral settings seam, or explicitly narrow CIA's supported Python contract.

src/aorta/cia/watch/cursors.py
```diff
@@ -0,0 +1,60 @@
...
+def read_new_bytes(path: Path, cursor: int, max_bytes: int = 32768) -> tuple[str, int]:
+    """Read up to max_bytes of new content from path starting at cursor.
+
+    Returns (new_text, new_cursor).
+    Resets cursor to 0 on file rotation (size shrank).
+    Returns ("", cursor) if nothing new.
+    """
+    try:
+        size = path.stat().st_size
+    except FileNotFoundError:
```
amd-vivekag, line 44: **[Medium]** One unreadable log can terminate monitoring for every job.

Only `FileNotFoundError` from `stat()` is handled; permission/NFS errors from `stat()` or `open()` escape the per-job loop.

Fix: catch `OSError` per watched file, emit a visible Watch error, and leave that file's cursor uncommitted for retry.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The latest remediation correctly fixes the model-controlled Autopsy root, model-selected log escape, remote-shell redirection, sidecar guess, and `watch_files` persistence. The queue, provider, classification, and profile fixes are partial: failed workers are suppressed by the in-memory alert set, the default namespaced vLLM model bypasses OpenAI-compatible qualification, and `inf` still turns ordinary inference/infrastructure failures into numeric evidence.

**Findings:** 4 High · 3 Medium · 0 Low

**Still open / partial from prior reviews**
- **High:** Autopsy failure retries do not run in the same Watch process.
- **High:** provider preservation breaks the default `deepseek-ai/...` vLLM model.
- **High:** matrix numeric evidence still false-matches `inference` / `infrastructure`.
- **Medium:** the executor limits workers but not queued tasks; its state file is not a durable bounded queue.
- **Medium:** Python 3.10 still cannot read the shared TOML profile.

Across the 42 prior inline comments: 34 are fixed correctly, 8 are partial, and none is wholly untouched. Of the September 16 review specifically, root containment, log containment, shell quoting, watch-file persistence, and sidecar ownership are fixed correctly; queueing, provider mapping, matrix classification, and profile parity are partial. I independently rechecked Copilot's nine comments: seven are fixed correctly; Python/profile compatibility and `JobRecord` escalation remain partial through the issues above.

**CI:** pre-commit, required CPU, Python 3.10-3.14 CPU, and chat checks are green; GPU checks are skipped. However, `tests/cia/conftest.py` now skips the entire CIA suite in CPU CI and no CIA-enabled workflow runs it, so the green checks do not exercise this feature. The branch is also behind `main`.

**Architecture:** the shared taxonomy/redaction direction and the new root bindings are sound. Approval still needs correct provider qualification, category-specific evidence, a genuinely bounded/recoverable Autopsy queue, and a CI lane that executes the CIA suite.

src/aorta/cia/watch/poll.py
```diff
@@ -0,0 +1,816 @@
...
+        "attempts": previous_attempts,
+        "next_attempt": attempt,
+    }
+    if signal:
+        fields["signal"] = signal
+    if confidence is not None:
+        fields["confidence"] = confidence
+
+    if not capacity.acquire(blocking=False):
+        record_autopsy_state(
```
amd-vivekag: **[High]** A failed deferral write permanently drops this diagnosis.

`record_autopsy_state` swallows `OSError`; this path still returns `False`, and the caller adds the job to `alerted`. No worker owns it, no durable record exists, and the committed log cursor supplies no new alert to recover from.

Fix: make state writes atomic and report success; only claim the job after either a worker accepts it or `deferred` is durably persisted, with an unwritable-state regression test.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The seven findings from the last pass are materially remediated: provider qualification, numeric token matching, same-process retries, the CIA CI matrix, Python 3.10 profiles, and per-file read failures are fixed correctly; queue admission is bounded and persisted in the normal path. Two core acceptance defects remain, so I am not approving.

**Blocking findings:** 2 High · 0 Medium · 0 Low

- **High:** a full-capacity deferral can still be lost when its state write fails; no worker owns it and the in-memory alert set suppresses recovery.
- **High:** the plan's required hardware Launch→Watch→Autopsy verification is not selected by GPU CI and has not run.

**Prior reconciliation:** of 49 prior inline findings, 47 are fixed correctly and 2 queue-family findings are partial; none is untouched. The two partial comments are the September 16 durable-queue requirement and the September 17 bounded/persisted admission finding. All nine Copilot comments are fixed correctly.

**CI:** pre-commit, CPU, chat, and the new CIA Python 3.10–3.14 matrix are green; a representative CIA leg ran 694 tests. GPU and workload-regression checks are skipped because the GPU path filter does not include CIA. The branch is current and mergeable.

**Architecture:** provider/redaction and filesystem trust roots are now sound, but queue durability on persistence failure and the promised hardware end-to-end evidence remain merge blockers.

### Follow-up issues (non-blocking)

1. **Title:** `CIA: add cooperative cancellation and prompt Watch shutdown`
   **Scope:** thread a stop signal through Watch, queued Autopsy, and production probe waits; cancel not-started work and mark it `abandoned`.
   **Acceptance:** Ctrl-C/API cancellation returns within a bounded interval, starts no new sweep, and leaves every accepted job in a recoverable terminal/deferred state.

2. **Title:** `CIA: stop mutating process-global TLS trust when building an LM`
   **Scope:** move certifi/custom-CA selection from `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` mutation into provider-client configuration.
   **Acceptance:** building a CIA LM leaves both environment variables unchanged while explicit custom CA, system trust, and opt-in certifi paths remain covered by tests.

docs/plans/cia-agents-port.md
```diff
@@ -0,0 +1,287 @@
...
+| `tests/chat/test_answer_shape.py` | the answer carries all three parts, and the evidence part names the tool and cites a file and line |
+| `tests/chat/test_no_callback_unchanged.py` | with no callback supplied the non-streaming path is byte-identical — the CLI is unaffected |
+
+## 6. Cross-cutting verification
+
+Once per PR, and again before the branch merges:
+
+1. **Headless.** Launch→Watch→Autopsy driven from a script with the chat extras
+   uninstalled. Catches any accidental dependency from `cia` on `chat` (F5).
+2. **On hardware** (`gpu`, `slow`), for each of the three instruments:
```
amd-vivekag, line 278: **[High]** This stated hardware acceptance check still never runs for CIA changes.

The GPU workflow does not select `src/aorta/cia/**` or a CIA end-to-end test, and the current GPU checks are skipped; all 694 CIA tests are CPU/mocked, so they cannot prove sanitizer → Watch → Autopsy actually executes on hardware.

Fix: add a marked hardware smoke covering these three bullets and include its code/test paths in the GPU relevance gate before merge.

amd-vivekag (approved):
No blocking findings remain — approving; please track cooperative CIA cancellation and per-client TLS trust isolation as separate follow-ups.

PR #424, opened 2026-09-02
feat(chat): choose a diagnostic tool from what the user pasted

aviator19941:
Second of three. Stacked on #423 — **review that one first**; GitHub will
retarget this to `feat/aorta-agent` when it merges. Plan:
`docs/plans/cia-agents-port.md`.

PR 1 landed the agents. This adds the tools that act on user-supplied code, and
the selector that decides which one to reach for.

## Three tools, all of which read what the user pasted

| tool | what it returns |
| --- | --- |
| `triage_kernel_source` | compiles a pasted HIP kernel, runs it under a sanitizer, reports accesses two waves made to the same shared-memory bytes with nothing ordering them |
| `triage_assembly_source` | assembles pasted instructions and reads the code object without executing it, reporting results consumed before a wait guarantees they landed |
| `triage_workload` | runs pasted training code or a command, with Watch reading the log and Autopsy classifying what it leaves behind |

**Five tools were deliberately not ported.** They run fixed built-in workloads
and ignore whatever was pasted. This is measured rather than assumed: in the
source repo, given a pasted RMSNorm with a non-finite loss, the selector chose
the fixed-workload tool — which then ran a *different*, hardcoded RMSNorm. Its
own rationale promised "the exact step and register state where it occurs",
truthfully, just not of the user's workload. A tool like that is worse than an
absent one, because the model picks it and silently substitutes the demo.

## Selection

One call, ranking tools by the evidence each returns. No symptom keywords, no
routing table — the model is shown what each tool produces and asked whether
that would answer the question in front of it.

It is **advisory**. An unreadable answer, a hallucinated tool name, or no model
at all leaves the agent seeing every tool; every failure widens rather than
narrows, because narrowing on a bad reading is how a chatbot confidently runs
the wrong thing. The one rule enforced in code is structural: a tool that reads
pasted source cannot run when nothing was pasted, and that is not a judgement.

The catalogue is **generated from the registry** rather than written beside it.
`load_chat_tools()` already returns a `BaseTool` per name and `@tool` takes its
description from the docstring, so there is no descriptor table to drift — and a
tool contributed through the `aorta.chat_tools` entry point is rankable without
being named in this package. There is a test for that.

## Two bugs found while testing

A fence marker counted as an instruction, so an empty code block assembled into
a valid kernel with nothing in it, which then reported no hazards — a false
clean, indistinguishable from a real one.

Two tool descriptions named a symptom (`triage_kernel_source` said "whether it
has a race"; my own `triage_workload` said "artifacts a crash left"). That is
precisely the coupling this design exists to remove, and a test now fails if a
description carries one. Worth noting what the test does *not* ban: five
existing tools say "Use this for exact searches" and similar, which is capability
guidance the act node legitimately reads. The hazard is symptom vocabulary, not
the imperative mood.

## Tests

53 offline, plus 5 marked `integration`.

The live ones run against a real model and are the only ones that answer the
question that matters — given a problem in a user's own words, is the top pick a
tool that reads it. Four pasted problems including **a transpose bug written
after the catalogue was**, plus a throughput question that must *not* start a
diagnosis (`inf` is a substring of `inference`, which is the false positive a
keyword matcher cannot avoid). All five pass.

They live in `tests/integration/` rather than `tests/chat/`, because that suite
deletes every `AORTA_CHAT_*` variable before each test so a developer's endpoint
cannot leak into a unit test. Correct for those, and the opposite of what a live
test needs.

`tests/chat/test_e2e_graph.py` gained a scripted selector, since `select` runs
between the router and the plan and its sequences script one model per node.
Otherwise the suite is unchanged: the 8 files still failing all fail identically
on `feat/aorta-agent` for want of the RAG and sqlite extras.

## Next

PR 3: the streamed reasoning and the three-part answer.


src/aorta/chat/graph/nodes.py
```diff
@@ -324,6 +325,102 @@ async def router_node(state: AgentState) -> dict[str, Any]:
...
+    except Exception as exc:
+        logger.warning("Selector unavailable (%s); the agent will see every tool.", exc)
+
+    candidates, dropped = enforce_requirements(
+        proposed[:MAX_CANDIDATES], has_pasted_source=_looks_like_pasted_source(text)
+    )
+    if dropped:
+        why = f"{why} Dropped {', '.join(dropped)}: needs source in the message.".strip()
+    logger.info("Selector: %s", candidates or "no candidate")
+    return {"candidate_tools": candidates, "selection_rationale": why}
```
amd-vivekag, line 615: **[High]** Nothing reads `candidate_tools`, so the selector is decoration.

I grepped the whole tree: `candidate_tools` and `selection_rationale` are written here, declared in `state.py`, asserted in tests, and (in #425) displayed in the UI. `plan_node` and both act paths never look at them — the model still receives the full tool list and picks unaided.

So the ranking this PR is titled after costs an LLM call per action turn and changes nothing about which tool runs. "Advisory" should still mean the recommendation reaches the model; right now it reaches only the screen. It also means `test_selector_live.py` proves the selector ranks well, not that a good ranking ever affects a run.

Fix: put the ranking into the act/plan prompt — e.g. append "Recommended tools, best first: … (because: `selection_rationale`)" to the system message — and keep the full list available so a bad ranking still only widens.

src/aorta/chat/plugins.py
```diff
@@ -39,6 +39,13 @@
     read_run_matrix,
     search_run_artifacts,
 )
+from aorta.chat.tools.cluster import (
+    list_cluster_jobs,
+    read_autopsy_report,
+    triage_assembly_source,
+    triage_kernel_source,
+    triage_workload,
+)
```
amd-vivekag: **[High]** This eager import makes `dspy` a hard requirement of `aorta chat`.

The chain is `aorta.chat.plugins` → `chat.tools.cluster` → `aorta.cia.triage` → `cia.watch.poll` → `cia.watch.watcher` → `import dspy`, all at module scope. `cia` is not in `chat-cli` or `chat-ui` (only in `all`), so `pip install amd-aorta[chat-ui]` then `aorta chat` raises `ModuleNotFoundError: dspy` at import. On Python 3.10 the `[cia]` extra installs nothing at all, so even asking for it does not help.

#423 has `test_importing_the_agents_does_not_require_the_chat_extras` for exactly this property in the other direction; this PR breaks the mirror image of it untested.

Fix: import `chat.tools.cluster` lazily inside `load_chat_tools()` and skip the five tools on ImportError, or add `amd-aorta[cia]` to `chat-cli` — plus a test that `aorta.chat.plugins` imports without dspy installed.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

The reasoning behind this PR is the strongest part of it — dropping the five canned tools because a measured run showed the selector substituting a hardcoded RMSNorm is the right call, and generating the catalogue from the registry instead of a descriptor table is the right shape. Two things stop it from delivering that.

First, the selector's output is never consumed. `candidate_tools` is written by `selector_node`, declared in state, asserted in tests, and displayed in #425's UI — but `plan_node` and both act paths never read it, so the model still sees every tool and picks unaided. The ranking costs an LLM call per turn and changes nothing about which tool runs. Second, removing the canned tools left the seams open: `_wrong_tool_hint` still instructs the model to "call run_nan_demo now" for a tool that does not exist, and about 100 lines of dead scaffolding from those five tools is still in `cluster.py`.

The rest are concrete: `dspy` becomes a hard import requirement of `aorta chat` without being in the chat extras, `run_triage` returns `1` from three error paths its callers `.get()` on, and `triage_workload` builds a filename out of a model-supplied label.

**Findings:** 5 High · 10 Medium · 0 Low

**High**

- `chat/graph/nodes.py:421` — nothing reads `candidate_tools`; the selector does not affect tool choice.
- `chat/plugins.py:42` — eager import chain pulls `dspy`, which is not in `chat-cli` / `chat-ui`, so `aorta chat` fails to import without the `[cia]` extra (and cannot get it at all on Python 3.10).
- `chat/tools/cluster.py:252` — `_wrong_tool_hint` directs the model to the unported `run_nan_demo`.
- `chat/tools/cluster.py:591` — model-supplied `label` used unsanitised as a path, and without a uniquifier.
- `cia/triage.py:310` — `return 1` (three sites) where callers do `result.get("ok")`, surfacing as `AttributeError`.

One question I could not answer from the diff: `triage_workload` and `run_terminal_command` will execute pasted code inside `containerize()`'s `docker run --security-opt seccomp=unconfined --ipc=host` with `/apps` and `$HOME` bind-mounted read-write. `cia/launch/confirm.py` landed in #423 but I do not see it on this path. Is there a confirmation gate before a paste runs, or is that assumed to be the deployment's problem? Worth stating explicitly in the docs either way.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,670 @@
...
+    """
+    if not any(marker in source for marker in _PYTORCH_MARKERS):
+        return ""
+    return (
+        " This is a PyTorch model, not a HIP kernel, and ConSan needs a __global__ "
+        "function to compile. If you are chasing a NaN or a non-finite loss, call "
+        "run_nan_demo instead: it runs this model and its kernel on an MI355X, traps "
+        "the first non-finite value with a device-side assert, and returns the root "
+        "cause. Do not ask the user for kernel source -- call run_nan_demo now."
+    )
```
amd-vivekag, line 312: **[High]** This tells the model to call a tool that does not exist.

`run_nan_demo` is one of the five canned tools this PR deliberately did not port — it is nowhere in the tree. So a user pasting PyTorch into `triage_kernel_source` gets back an instruction ("Do not ask the user for kernel source -- call run_nan_demo now") that can only produce a hallucinated tool call, and the turn dead-ends in exactly the way this hint was written to prevent.

The removal looks half-finished elsewhere in this file too: `KNOWN_RECIPES`, `WAITCHECK_RECIPE`, `_format_waitcheck`, `_NAN_CACHE`, `_WAITCHECK_CACHE` and the four `_NAN_*` regexes have no readers — roughly 100 lines of leftovers from the same five tools.

Fix: redirect to `triage_workload`, which is the general replacement, and delete the dead leftovers.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,670 @@
...
+        )
+
+    name = label or "workload"
+    if command.strip():
+        return _run_triage(["--command", command.strip()], name)
+
+    staged = settings.jobs_root / "staged"
+    staged.mkdir(parents=True, exist_ok=True)
+    script = staged / f"{name.replace(' ', '_')}.py"
+    script.write_text(source, encoding="utf-8")
```
amd-vivekag, line 599: **[High]** A model-supplied `label` becomes a filename with no sanitising.

`name` is `label or "workload"` straight from the tool call. `replace(" ", "_")` leaves `/` and `..` intact, so `label="../../../../.bashrc"` writes user-pasted `source` outside `staged`. Two runs with the same label also overwrite each other's script mid-flight — the kernel and assembly paths both timestamp the stem, this one does not.

Fix: derive the stem from the label but keep it inside the directory and make it unique.

```suggestion
    stem = re.sub(r"[^A-Za-z0-9_.-]", "_", name)[:60] or "workload"
    script = staged / f"{stem}-{datetime.now().strftime('%Y%m%d-%H%M%S')}.py"
```

src/aorta/cia/triage.py
```diff
@@ -0,0 +1,556 @@
...
+        f"; {shlex.quote(venv_bin('python'))} "
+        f"{shlex.quote(str(Path(__file__).resolve().parent / 'echo_findings.py'))} "
+        f"{shlex.quote(str(Path(aorta_output) / 'sanitizer_report.json'))} || true"
+    )
+
+    if compiled_from_source:
+        src = Path(args.source).expanduser().resolve()
+        if not src.is_file():
+            print(json.dumps({"ok": False, "error": f"source not found: {src}"}))
+            return 1
```
amd-vivekag: **[High]** `run_triage` returns `1` where its callers expect a dict.

Three error paths do this (here, 319, 362) — source missing, no `__global__` kernel found, recipe missing. `_run_triage` in `chat/tools/cluster.py` then calls `result.get("ok")` on an `int` and the broad `except Exception` turns a clear "source not found" into `Error: triage failed: AttributeError: 'int' object has no attribute 'get'`. `main()` has the same problem. These also `print(json.dumps(...))` to stdout, which is now the chat server's stdout rather than a subprocess pipe.

Fix: return the dict the signature promises.

```suggestion
            return {"ok": False, "stage": "source", "error": f"source not found: {src}"}
```

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,670 @@
...
+
+    Args:
+        source: The assembly text, exactly as the user pasted it.
+        label: Human label for the run.
+
+    Returns:
+        The hazards found in the supplied assembly, or a clean result.
+    """
+    try:
+        prepared = prepare_asm(source)
```
amd-vivekag: **[Medium]** The wrapped assembly targets gfx950 even when `CIA_GPU_ARCH` says otherwise.

`prepare_asm` defaults `arch="gfx950"` and is called with no `arch`, so the generated `.amdgcn_target "amdgcn-amd-amdhsa--gfx950"` is fixed, while the clang command below passes `-mcpu={_ARCH}`. Set `CIA_GPU_ARCH=gfx942` and the two disagree — the assemble either fails with a confusing target mismatch or the analysis is about the wrong ISA.

Fix: pass it through.

```suggestion
        prepared = prepare_asm(source, arch=_ARCH)
```

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,670 @@
...
+        "the first non-finite value with a device-side assert, and returns the root "
+        "cause. Do not ask the user for kernel source -- call run_nan_demo now."
+    )
+
+
+# Triaging one kernel costs a multi-minute cluster job, and the same kernel gets
+# submitted more than once per conversation: the model re-checks its own answer and
+# the critic can re-run a turn. Reusing the verdict for an identical kernel keeps a
+# demo to one job per kernel instead of one per attempt.
+_TRIAGE_CACHE: dict[tuple[str, int, int], str] = {}
```
amd-vivekag: **[Medium]** These caches are process-global, not per-conversation, and never evict.

The comment and the user-facing text both say "in this conversation", but a module-level dict in a Chainlit server is shared by every session: user B pasting the same kernel is told a job was reused when it was someone else's run. Keyed on full pasted source with no cap, they also grow for the process's lifetime — `_ASM_CACHE` and `_TRIAGE_CACHE` hold every kernel anyone ever submitted.

Fix: key on the session (Chainlit's `cl.user_session`, or a cache passed in via state) and bound it with an LRU, and fix the wording if it stays global.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,670 @@
...
+        # and it points at the package while recipes sit beside src/.
+        ("AORTA_PATH", _default_aorta_root()),
+        ("CIA_JOBS_ROOT", str(settings.jobs_root)),
+    ):
+        if value:
+            argv += ["--env", f"{key}={value}"]
+
+    pool = ThreadPoolExecutor(max_workers=1)
+    future = pool.submit(run_triage, argv)
+    pool.shutdown(wait=False)
```
amd-vivekag: **[Medium]** A new executor per call, abandoned on timeout, blocks interpreter exit.

`ThreadPoolExecutor` threads are non-daemon and `concurrent.futures` installs an atexit hook that joins them, so a `triage_timeout` expiry leaves a thread running `run_triage` — up to 4 hours if it reaches `run_aorta_probe`'s wait loop — and the chat server cannot shut down until it finishes. Every timed-out call adds another. In #425 each triage also holds an `asyncio.to_thread` worker, so it is two threads per call.

Fix: use one module-level executor with a small bounded `max_workers`, and make the abandoned work interruptible (a stop event the poll loops check) rather than relying on the process outliving it.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,670 @@
...
+from aorta.chat.config import settings
+from aorta.cia.triage import _default_aorta_root, run_triage
+from aorta.chat.tools.harness.assembly import AsmHarnessError, prepare_asm
+from aorta.chat.tools.harness.kernel import WAVEFRONT, HarnessError, prepare_source
+
+_PROJECT_ROOT = Path(__file__).resolve().parents[2]
+_ARCH = os.environ.get("CIA_GPU_ARCH", "gfx950")
+# The assembler lives with ROCm on the compute nodes, not on the login node.
+_ROCM_LLVM = os.environ.get("ROCM_LLVM_BIN", "/opt/rocm-7.0.2.2/lib/llvm/bin")
+
```
amd-vivekag, line 55: **[Medium]** A pinned ROCm patch version as the default assembler path.

`/opt/rocm-7.0.2.2/lib/llvm/bin` exists on one machine. Anywhere else `triage_assembly_source` fails at the clang invocation, and the message the user sees is "did not assemble" with empty assembler output — a tooling gap reported as a source problem.

This file is also outside what `tests/cia/test_settings_neutral.py` scans (`src/aorta/cia` only), so the F3 guard does not see it.

Fix: default to `/opt/rocm/lib/llvm/bin` (the version-independent symlink), fall back to `shutil.which("clang")`, and say plainly when neither is present.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,670 @@
...
+    clang = f"{_ROCM_LLVM}/clang"
+    build = (
+        f"{shlex.quote(clang)} -target amdgcn-amd-amdhsa -mcpu={shlex.quote(_ARCH)} "
+        f"{shlex.quote(str(asm_path))} -o {shlex.quote(str(obj_path))}"
+    )
+    proc = subprocess.run(
+        ["srun", "--nodes=1", "-t", "5", "bash", "-c", build],
+        capture_output=True, text=True, timeout=settings.waitcheck_timeout,
+        stdin=subprocess.DEVNULL,
+    )
```
amd-vivekag: **[Medium]** `srun` is assumed present and this call is unguarded.

No slurm means `FileNotFoundError`, and a slow queue means `subprocess.TimeoutExpired`; neither is caught, so both escape as a raw traceback out of a LangChain tool. The `_run_waitcheck` call below is wrapped for exactly these two — this one is not.

Fix: wrap it in the same `except (subprocess.TimeoutExpired, OSError)` and return a message saying the assembler could not be reached, and fall back to running clang locally when `srun` is absent.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,670 @@
+"""Tools that let the assistant drive the Cluster Intelligence Agents.
+
+These wrap the Launch -> Watch -> Autopsy pipeline so a user can go from "why did
+my kernel race?" to a classified failure bundle without touching a terminal. The
+agents live in their own virtualenv, so everything here shells out to that venv's
+console scripts rather than importing them into the chatbot process.
```
amd-vivekag: **[Medium]** This docstring describes the opposite of what the module does.

"The agents live in their own virtualenv, so everything here shells out to that venv's console scripts rather than importing them into the chatbot process" — but line 26 imports `run_triage` and `_run_triage` calls it in-process, as its own docstring says ("this is a call rather than a subprocess"). `cia/triage.py`'s module docstring is stale the same way: "Executed with the cluster-intelligence-agent virtualenv" and "Emits a single JSON object on stdout; progress goes to stderr so the caller can show it live", which is why the `print`/`log` calls still exist.

Fix: rewrite both to describe the in-process call, and route `log()` through `logging` now that stderr is the chat server's.

src/aorta/chat/graph/nodes.py
```diff
@@ -324,6 +325,102 @@ async def router_node(state: AgentState) -> dict[str, Any]:
...
+"""
+
+#: Anything below is prose. Fences and these markers are what tell source from
+#: a description of source, which decides whether a pasted-source tool is even
+#: possible -- see enforce_requirements.
+_CODE_MARKERS = ("```", "__global__", "__shared__", "def ", "class ", "s_load", "v_mov", "import ")
+
+
+def _looks_like_pasted_source(text: str) -> bool:
+    """Whether the message carries code, as opposed to describing some."""
```
amd-vivekag, line 528: **[Medium]** Only the latest message is checked, so a follow-up turn drops the source tools.

`_looks_like_pasted_source` sees `messages[-1]`. A user pastes a kernel, the agent asks a clarifying question, the user replies "yes, run the sanitizer on it" — no marker, so `enforce_requirements` drops `triage_kernel_source` and `triage_assembly_source` for the turn that actually asked for them. The rationale then says "needs source in the message", which is wrong: the source is in the conversation.

Also `"import "`, `"def "` and `"class "` match prose like "I import the model and define a class", which errs the harmless way but makes the signal noisy.

Fix: scan the recent human messages, not just the last one, and prefer the structural markers (fences, `__global__`, `__shared__`) over the English-word ones.

src/aorta/cia/triage.py
```diff
@@ -0,0 +1,556 @@
...
+        launcher="sbatch",
+        env_vars=job_env_vars,
+    )
+
+    log(f"job_id={job_id} recipe={recipe.name if recipe else '(raw command)'} label={args.label or '-'}")
+    reconcile_stale_jobs(jobs_root)
+    log("── Launch ──")
+
+    saved = os.environ.get("CIA_TOLERATE_NONZERO")
+    os.environ["CIA_TOLERATE_NONZERO"] = "1"
```
amd-vivekag: **[Medium]** Mutating `os.environ` is not thread-safe, and this runs on a worker thread.

`_run_triage` submits `run_triage` to an executor, so two concurrent triages race on `CIA_TOLERATE_NONZERO` — the second's `finally` can restore the value the first is still relying on. The `env = os.environ.copy()` / `env["CIA_TOLERATE_NONZERO"] = "1"` at 378-379 is also dead: `env` is never used again, and the `run()` helper it was for has no callers.

Fix: pass the flag explicitly — give `build_sbatch_script` a `tolerate_nonzero: bool` parameter instead of reading the environment — and delete the dead `env` and `run()`.

src/aorta/chat/tools/capabilities.py
```diff
@@ -0,0 +1,90 @@
...
+
+from __future__ import annotations
+
+#: Tools that read something the user supplied in the message. Proposing one of
+#: these with nothing pasted is impossible, not merely unlikely, so it is the
+#: single filter applied to the model's ranking.
+_REQUIRES: dict[str, frozenset[str]] = {
+    "triage_kernel_source": frozenset({"pasted_source"}),
+    "triage_assembly_source": frozenset({"pasted_source"}),
+}
```
amd-vivekag, line 47: **[Medium]** The one rule enforced in code is the one a plugin tool cannot declare.

The module's premise is that the catalogue is generated so "a tool contributed through the `aorta.chat_tools` entry point is rankable without being named in this package" — true for ranking, but `_REQUIRES` is a central dict keyed by hardcoded name. A third-party tool that reads pasted source has no way to say so, and will be proposed with nothing to read.

Fix: read the requirement off the tool instead of a table here — a `requires` attribute or metadata key on the `BaseTool`, with `_REQUIRES` kept only as the fallback for the two built-ins.

src/aorta/chat/tools/harness/assembly.py
```diff
@@ -0,0 +1,147 @@
...
+    Engineers paste their symptoms around the code, and the model deciding what
+    to pass along is a judgement it sometimes gets wrong. Prose reaching the
+    assembler produces a wall of syntax errors that says nothing about the bug,
+    so a fenced block, when present, wins over everything around it.
+    """
+    blocks = _FENCED.findall(text)
+    for block in blocks:
+        if _looks_like_instructions(block):
+            return block.strip("\n")
+    return text
```
amd-vivekag: **[Medium]** With no fenced block, prose reaches the assembler after all.

The docstring says a fence "wins over everything around it", and it does — but the fallback returns the whole message, and `_looks_like_instructions` accepts any line that is not blank, a fence, a comment or a label. "the kernel hangs after the second load" passes, gets wrapped in a descriptor and hits clang, producing the wall of syntax errors this function exists to avoid. `test_harness.py` covers "prose around a fenced block", which is the case that works.

Fix: when no fenced block qualifies, require the unfenced text to look like instructions with a stricter test (a known mnemonic prefix, or a majority of lines parsing as `op dst, src`) and raise `AsmHarnessError` asking for the code otherwise.

src/aorta/chat/graph/nodes.py
```diff
@@ -324,6 +325,102 @@ async def router_node(state: AgentState) -> dict[str, Any]:
...
+        response = await llm.ainvoke(
+            [
+                SystemMessage(
+                    content=_SELECTOR_PROMPT.format(
+                        catalogue=catalogue(), limit=MAX_CANDIDATES
+                    )
+                ),
+                HumanMessage(content=text),
+            ]
+        )
```
amd-vivekag, line 596: **[High]** The new node bypasses `_send`, which is the Decision 16 gate.

`_send` exists so "the gate cannot be bypassed by a node added later." This is that later node. `selector_node` calls `llm.ainvoke(...)` with the user's latest message verbatim. On `llm_provider = openai|litellm` that message — which for this PR is pasted HIP/assembly/training code, plus whatever prose sits around it — leaves the machine with no path/IP rewrite.

The rest of the graph (router, plan, act, critic, answer) all go through `_send`. There is no reason this one should not.

Fix:

```suggestion
        response = await _send(
            llm,
            [
                SystemMessage(
                    content=_SELECTOR_PROMPT.format(
                        catalogue=catalogue(), limit=MAX_CANDIDATES
                    )
                ),
                HumanMessage(content=text),
            ]
        )
```

src/aorta/chat/config.py
```diff
@@ -139,6 +139,30 @@ class Settings(BaseSettings):
     remote_llm_auth_header: str = ""
     remote_llm_extra_headers: Annotated[dict[str, str], NoDecode] = {}
 
+    # --- Cluster Intelligence Agents ---
+    # Every one of these is empty or a duration by default. A guessed path or a
+    # guessed node is worse than an unset one: it does not fail, it runs
+    # somewhere nobody meant and reports nothing useful.
+    #: Where job records and bundles are written. Empty means the agents'
+    #: own default. Must be readable from every node that runs work.
+    jobs_path: str = ""
```
amd-vivekag, line 159: **[High]** Two prefixes for the same knobs, which is the collision `docs/chat/configuration.md` exists to prevent.

The integration docs lock one prefix: every chat setting is `AORTA_CHAT_<NAME>` or a key in `chat.toml`. This PR adds `jobs_path` / `cia_demo_node` / `rocjitsu_*` here (so they become `AORTA_CHAT_JOBS_PATH`, …), while Launch/Watch/Autopsy still read `CIA_JOBS_ROOT`, `CIA_DEMO_NODE`, `CIA_PARTITION`, `CIA_SSH_USER`. `jobs_root` even defers to `default_jobs_root()`, which reads `CIA_JOBS_ROOT`.

A user following `docs/chat/configuration.md` will set `jobs_path` in the profile and watch Launch ignore it. The port plan's F3 said these knobs belong on this Settings object *instead of* a second env dialect, not in addition to one.

Fix: one name. Either Launch reads `settings.jobs_root` (and CIA_* becomes an alias that Settings maps), or drop these fields and document `CIA_*` — not both. And add them to the settings table in `docs/chat/configuration.md`; that file is still the ten-tool, no-cluster description.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,670 @@
...
+
+# Triaging one kernel costs a multi-minute cluster job, and the same kernel gets
+# submitted more than once per conversation: the model re-checks its own answer and
+# the critic can re-run a turn. Reusing the verdict for an identical kernel keeps a
+# demo to one job per kernel instead of one per attempt.
+_TRIAGE_CACHE: dict[tuple[str, int, int], str] = {}
+
+
+@tool
+def triage_kernel_source(
```
amd-vivekag, line 316: **[High]** These tools break the sandbox the aorta_llm integration documented as a hard bound.

`docs/chat/extending.md`: "Bound what you touch. The built-in file tools resolve every path relative to the configured source root and refuse anything that escapes it." aorta_llm `ARCHITECTURE.md` §5: path containment, command allowlist, blocked `ssh`/`curl`, 60s timeout, cwd locked to `AORTA_PATH`.

`triage_kernel_source` / `triage_assembly_source` / `triage_workload` do the opposite: they write under `jobs_root`, `sbatch`/`srun`/`ssh` onto the cluster, `docker run --security-opt seccomp=unconfined` with `/apps` and `$HOME` mounted, and execute pasted source. `confirm.py` from #423 is not on this path. `run_terminal_command` would reject every one of those operations.

That may be the product, but then it is a new class of tool and the integration docs have to say so — including that a chat turn can submit a multi-minute GPU job with no confirmation. Shipping them as ordinary `@tool` entries next to `list_files` is how an LLM with `llm_tool_mode=native` also gets them with no extra gate, because native mode "sends every tool's schema through the provider's function-calling API."

Fix: keep them off the default tool list until a confirmation / allowlist setting is on (`AORTA_CHAT_ALLOW_CLUSTER_JOBS=false` by default); when enabled, document the bound (jobs_root only, no SSH as root, no `seccomp=unconfined` unless asked). Do not rely on the selector to be the sandbox.

src/aorta/cia/triage.py
```diff
@@ -0,0 +1,556 @@
...
+    try:
+        slurm_id, err = submit_sbatch(
```
amd-vivekag: **[Medium]** This driver goes around the Launch seam #423 added.

`src/aorta/cia/launch/__init__.py` says `launch()` is the seam and "callers should not reach for `submit_sbatch` directly." This is the one production caller, and it imports `submit_sbatch`. The local-backend work F4 was meant to enable then has two call sites again.

Fix: call `launch(...)`.

amd-vivekag (commented):

**Verdict: still fix-before-ship**

Against `docs/chat/{providers,configuration,redaction,extending,README}.md` and aorta_llm `ARCHITECTURE.md`, not just the port plan.

The selector is wired into the graph as if it were a peer of router/plan/act, then it skips the one function (`_send`) that makes a new node safe. The cluster tools are registered as ordinary builtins, which under `llm_tool_mode = native` means they are offered to the model with the same status as `list_files`, while they violate every sandbox rule in those docs. And the knobs this PR adds to `Settings` (`AORTA_CHAT_JOBS_PATH`) are not the knobs Launch actually reads (`CIA_JOBS_ROOT`).

**New findings:** 3 High · 2 Medium

**High**
- `selector_node` calls `llm.ainvoke` and skips Decision 16 redaction.
- CIA fields on `Settings` vs `CIA_*` in the agents: two prefixes for one fact.
- Cluster tools have no `AORTA_PATH` bound, no confirmation, and ride the native tool-calling path.

The `docs/chat/*` files are not in this diff. They need to be, or the integration docs become a description of a product that no longer exists.

src/aorta/chat/plugins.py
```diff
@@ -52,6 +59,13 @@
 #: :func:`load_chat_tools` raises if they drift, since that is an aorta bug.
 BUILTIN_CHAT_TOOLS: dict[str, BaseTool] = {
     "list_files": list_files,
+    # Diagnostics: each runs what the user supplied on a GPU node and
+    # classifies what came back.
+    "triage_kernel_source": triage_kernel_source,
```
amd-vivekag: **[Medium]** `docs/chat/extending.md` still says chat ships ten tools, and `docs/chat/installation.md` still lists three extras.

This PR adds five builtins and a `[cia]` extra that `chat-cli` / `chat-ui` do not depend on. The integration guides are the contract operators read. Untouched, they describe a sandbox that no longer exists and an install that cannot import this module (see the eager `dspy` chain).

Fix: update `extending.md` (tool count, cluster-tool bound, "errors as strings"), `installation.md` (`[cia]`, the 3.11 floor, that `chat-ui` alone is not enough), and `configuration.md` (the new Settings keys, `CIA_*` aliases if any). The `aorta chat tools` listing will otherwise be the first place anyone learns these exist.

amd-vivekag:

src/aorta/chat/tools/harness/kernel.py
```diff
@@ -0,0 +1,248 @@
...
+            args.append("static_cast<int>(kElements)")
+
+    body = "\n".join(
+        [
+            "// Harness generated by AORTA so the pasted kernel can be launched under ConSan.",
+            "int main() {",
+            f"  constexpr size_t kElements = {count};",
+            *setup,
+            "",
+            f"  {name}<<<{grid}, {block}>>>({', '.join(args)});",
```
amd-vivekag, line 187: **[Medium]** The generated harness always launches one-dimensional geometry.

The PR's transpose-style workload can use `threadIdx.y`; running it with only scalar block/grid sizes analyzes materially different behavior and can report a false clean.

Fix: support `dim3` geometry inferred or supplied for x/y/z, or reject kernels that reference y/z dimensions.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,578 @@
...
+        return _run_triage(["--command", command.strip()], name)
+
+    staged = settings.jobs_root / "staged"
+    staged.mkdir(parents=True, exist_ok=True)
+    script = staged / f"{name.replace(' ', '_')}.py"
+    script.write_text(source, encoding="utf-8")
+    # {bundle} is substituted by the driver, so a workload that writes
+    # artifacts puts them where Autopsy will look.
+    return _run_triage(
+        ["--command", f"{shlex.quote(sys.executable)} {shlex.quote(str(script))} {{bundle}}"],
```
amd-vivekag, line 603: **[Medium]** Appending the bundle as an unnamed argument changes the submitted program's CLI.

A pasted script using strict `argparse` fails before its workload runs, so Autopsy diagnoses the wrapper-induced usage error rather than the user's failure.

Fix: expose the bundle through a named environment variable and run the submitted source with its original argv.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

Generating the catalogue from the live registry and dropping canned demo tools are good architectural choices. The implementation still does not deliver the PR's central behavior: the selector's result is displayed but never consumed by plan/native-act/text-act, so the extra LLM call cannot affect which tool runs. At the same time, cluster execution is registered unconditionally beside read-only tools and bypasses the chatbot's opt-in, containment, redaction, and execution boundaries.

**Findings:** 7 High · 11 Medium

**Still open from the last review**
- All prior High items remain: unused selector output, eager DSPy import, nonexistent `run_nan_demo` recovery, model-controlled staging path, and `run_triage()` returning integers to dict callers; the selector still bypasses `_send`, and dangerous cluster execution still has no operator gate.
- The process-global caches, environment race, follow-up source loss, assembly architecture mismatch, prose-to-assembler path, and docs/config integration gaps also remain.
- Fixed correctly: the hard-coded ROCm patch path and direct bypass of the Launch seam.

I independently confirmed every current Copilot finding, including its suppressed assembly-cache finding, and did not duplicate those inline.

- `pre-commit`: inherited missing final newline in `src/aorta/cia/autopsy/adapters/__init__.py`.
- Chat Python 3.11–3.14: importing `aorta.chat.plugins` eagerly imports CIA/DSPy, but `chat-cli`/`chat-all` do not install `[cia]`; representative runs have 31 failures plus 22–25 collection errors.
- CPU Python 3.10–3.14: CIA tests are collected without DSPy, and `tests/integration/test_selector_live.py` imports `langchain_core` at collection time in an environment without chat dependencies.
- Make CIA-backed tools optional/lazy behind an explicit setting and extra, add a CIA-enabled chat lane, and keep live integration tests out of slim CPU collection (or defer their optional imports). The red matrix is caused by this stack, not unrelated baseline failures.

**Architecture integration**
Treat cluster diagnostics as a separate capability class, disabled by default, with explicit admission/confirmation, trusted roots, bounded concurrency, cancellation, and audit records. The selector should steer the act prompt/tool binding while retaining a fail-wide fallback. Move all outbound model traffic through one redaction gate, and update `docs/chat/*` so the documented sandbox and install extras match the product that ships.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

This revision correctly fixes many earlier issues: ordinary chat now starts without DSPy, workload staging is contained and unique, `run_triage()` returns its documented type, caches are per-session and bounded, assembly uses the selected architecture, waits are interruptible, environment mutation is gone, and the Launch seam is used. Approval is still unsafe because four High architecture/security issues remain and every required CI aggregate is red.

**Findings:** 4 High · 8 Medium

**Still open from prior reviews**
- **High:** `src/aorta/chat/graph/nodes.py:587` still calls the selector through `llm.ainvoke`, sending pasted source around `_send` and the redaction gate.
- **High:** `src/aorta/chat/plugins.py:92-120` registers model-authored cluster execution merely because `[cia]` is installed, with no separate admission/confirmation control.
- **High:** `src/aorta/chat/tools/cluster.py:660-675` still lets `read_autopsy_report` escape `jobs_root` through absolute, traversal, or symlinked paths.
- **High:** `src/aorta/chat/graph/nodes.py:978-997` reconstructs text-mode framing without `_recommendation`; because text mode is the default, the headline selector still does not steer default execution. Fix by starting from `_act_messages(state)` and then appending `TOOL_DESCRIPTIONS`.

**Other actionable findings**
- `src/aorta/cia/triage.py:103-118`: cancellation stops local waiting but never cancels the Slurm allocation. Add cancellation to the scheduler seam and issue `scancel` (or backend equivalent).
- Workload source still receives an extra positional `{bundle}` argument; valid strict-`argparse` reproducers fail before reproducing.
- Job listing slices staging directories before checking for `job.json`; multidimensional HIP geometry and fenced HIP extraction remain unsupported; assembly caches transient failures; hardware output still names MI355X without observing it.

Human reconciliation: 14 of 22 prior findings are fixed correctly, 3 are partial, and 5 remain open. Copilot reconciliation: 5 fixed, 2 partial, 10 still applicable; confirmed items were not duplicated inline.

**CI**
- CPU inherits #423's missing-DSPy CIA collection failures and adds optional-import errors.
- Chat Python 3.11–3.14 has 18 failures plus 14 errors on representative legs: the lane does not install `[cia]`, while tests assume diagnostic tools are present. The redaction chokepoint independently catches the direct selector `ainvoke`.
- `pre-commit` inherits #423's EOF/line-ending failure.
- Split no-CIA and CIA-enabled tests explicitly, add a `[chat-*,cia]` lane for supported integration, and keep optional imports out of collection-only environments.

**Architecture integration:** registry-derived capabilities and session-scoped caches are sound. The integration is not solid until cluster execution becomes an explicit capability with admission control, every path is rooted consistently, default text mode consumes selection, and scheduler cancellation owns the full job lifecycle. `docs/chat/*` must also describe the extra, execution boundary, settings, and progress behavior.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,722 @@
...
+    Args:
+        job_id: The CIA job id, e.g. cia-20260819-232554-6ec890.
+
+    Returns:
+        The report JSON, including category, confidence, rationale and evidence.
+    """
+    try:
+        job_dir = resolve_within(settings.jobs_root, job_id, JOBS_ROOT_LABEL)
+    except ValueError as exc:
+        return f"Error: {exc}"
```
amd-vivekag, line 899: **[High]** Containment stops at `job_dir`, not the file being read.

A workload-writable `bundle/report.json` symlink can point outside `jobs_root`, and `read_text()` follows it.

Fix: resolve and validate the complete report path against the resolved jobs root immediately before opening it.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,722 @@
...
+    if cached is not None:
+        return (
+            "(Reusing the triage already run for this exact kernel in this "
+            "conversation — no second cluster job was submitted.)\n\n" + cached
+        )
+
+    staging = settings.jobs_root / "chat-kernels"
+    staging.mkdir(parents=True, exist_ok=True)
+    stamp = datetime.now().strftime("%Y%m%d-%H%M%S")
+    src_path = staging / f"{prepared.kernel}-{stamp}.hip"
```
amd-vivekag: **[High]** Kernel staging is still shared and only unique to one second.

Two sessions submitting the same kernel name can overwrite this file before the first job reads it; assembly has the same defect at lines 528-533.

Fix: create a per-invocation UUID/mkdtemp directory and write with exclusive creation.

src/aorta/chat/tools/harness/kernel.py
```diff
@@ -0,0 +1,356 @@
...
+    for param in params:
+        if param.is_pointer:
+            var = f"h_{param.name}"
+            setup.append(f"  {param.base_type} *{var} = nullptr;")
+            setup.append(
+                f"  if (hipMalloc(&{var}, kElements * sizeof({param.base_type})) != hipSuccess)\n"
+                f"    return 1;"
+            )
+            setup.append(
+                f"  if (hipMemset({var}, 0, kElements * sizeof({param.base_type})) != hipSuccess)\n"
```
amd-vivekag: **[High]** Zero-filled generated inputs can produce a false-clean sanitizer result.

A race behind `if (input[i] > 0)` never executes under this harness even though it exists for real inputs.

Fix: accept explicit initialization/arguments, or mark generated-harness clean results inconclusive and require a caller-supplied `main()` for data-dependent kernels.

src/aorta/cia/triage.py
```diff
@@ -0,0 +1,678 @@
...
+    log.info("── Watch ──")
+    log.info(f"poll_jobs(rounds={args.watch_rounds})")
+    watcher = threading.Thread(
+        target=poll_jobs,
+        kwargs={"jobs_root": jobs_root, "max_rounds": args.watch_rounds, "stop": stop},
+        daemon=True,
+    )
+    watcher.start()
+
+    state = wait_for_job(slurm_id, timeout=args.job_timeout, stop=stop)
```
amd-vivekag, line 533: **[Medium]** Internal job timeout still leaves the Slurm allocation running.

Only `ABANDONED(...)` reaches `scancel`; `TIMEOUT_WAITING(...)` can hold the node until its four-hour limit.

Fix: cancel and persist terminal state for both abandonment and timeout paths.

src/aorta/chat/graph/nodes.py
```diff
@@ -431,6 +487,138 @@ async def router_node(state: AgentState) -> dict[str, Any]:
...
+    whether that would answer this question, so a problem described in words
+    nobody anticipated still reaches the right instrument.
+
+    Advisory: on any failure the act node still sees every tool. The only thing
+    enforced is structural -- a tool that reads pasted source cannot run when
+    nothing was pasted.
+    """
+    from aorta.chat.tools.capabilities import MAX_CANDIDATES, catalogue, enforce_requirements
+
+    text = str(state["messages"][-1].content)
```
amd-vivekag: **[Medium]** Multi-turn selection still sends only the latest reply to the model.

Requirement checks find an earlier paste, but a follow-up like `32, go ahead` gives the selector no source to rank.

Fix: send the same bounded recent human-message window used by `_conversation_has_source`.

src/aorta/cia/triage.py
```diff
@@ -0,0 +1,678 @@
...
+        log_path=log_path,
+        script_path=job_dir / "launch.sbatch",
+        working_dir=args.aorta_root,
+        env_vars=record.env_vars,
+        node=args.node,
+        # The sanitizer positive control exits non-zero by design ("guardrail
+        # not clean"), so the batch script swallows the code and the verdict
+        # comes from the sanitizer report. Per call, not per process: triages
+        # run concurrently and share one environment.
+        tolerate_nonzero=True,
```
amd-vivekag: **[Medium]** Raw workloads also suppress nonzero job exits.

That can record a failed user program as `COMPLETED`; only sanitizer positive controls need this behavior.

Fix: set `tolerate_nonzero` only for recipe/source sanitizer paths.

src/aorta/cia/triage.py
```diff
@@ -0,0 +1,678 @@
...
+    log.info(f"submitted slurm job {slurm_id}")
+
+    bundle = job_dir / "bundle"
+    report_path = bundle / "report.json"
+
+    # Watch has to start while the record still says 'running', because the job
+    # registry is what makes it eligible for monitoring at all.
+    log.info("── Watch ──")
+    log.info(f"poll_jobs(rounds={args.watch_rounds})")
+    watcher = threading.Thread(
```
amd-vivekag, line 521: **[Medium]** Every concurrent triage starts another watcher over every active job.

Two watchers can assess and trigger Autopsy for the same job concurrently.

Fix: use one watcher service, or scope each poller to its submitted job with an atomic alert claim.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The selector now reaches both act modes through the redaction gate, CIA imports are lazy, cluster submission is operator-opt-in, configuration names converge, and workload staging/return types are materially improved. Four merge blockers remain: long tools still block the shared async loop (`chat/graph/nodes.py:946,1075`), final report reads can escape through nested symlinks, kernel/assembly staging collides across sessions, and generated zero-filled harness inputs can turn data-dependent races into false-clean results.

**Findings:** 4 High · 4 Medium

**Prior reconciliation:** the four High issues from the last review (selector redaction, opt-in registration, direct report traversal, and default-text selection) are substantially addressed; report containment is only partial because the final component is not revalidated. Scheduler cancellation is partial, and multi-turn source selection still loses the earlier paste.

#425 contains fixes for off-loop execution and staging uniqueness, but #424 must remain independently safe before it merges.

**CI:** CPU and chat matrices pass; pre-commit remains red and GPU checks are skipped.

**Architecture integration:** the registry-derived selector, common redaction seam, opt-in capability class, session cache, and Launch abstraction are sound. Integration is not solid until execution is async/cancellable end-to-end, every final artifact read is contained, and generated harnesses cannot claim clean without representative inputs.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The latest fixes correctly move tools off the event loop, contain final report reads, isolate ordinary staging calls, cancel timed-out jobs, preserve raw-workload exits, show the selector earlier source, and scope/claim Watch work. Two merge blockers remain: a complete assembly unit can put path separators into the new staging prefix, escaping `jobs_root`, and the generated-harness caveat still misses common data-dependent paths.

**Findings:** 2 High · 4 Medium · 0 Low

**Still open / partial from prior reviews**
- **High:** generated-harness honesty is partial; direct pointer guards are detected, but scalar guards and indirect pointer-derived guards still return an unqualified pass.
- **Medium:** multidimensional geometry remains partial because `blockIdx.y/z` is treated as block geometry while the grid stays one-dimensional.
- **Medium:** `docs/chat/installation.md:7-15` still documents only three chat extras and never tells an operator to install `[cia]`.

**Prior reconciliation:** of the 29 prior human inline findings, 26 are fixed correctly, 2 are partial, and 1 is unaddressed. For the September 16 review specifically, 6 of 7 inline findings are fixed correctly and the harness finding is partial; the summary-only event-loop blocker is fixed. Copilot's 17 inline findings are independently reconciled as 16 fixed correctly and 1 partial (multidimensional geometry); its suppressed assembly-cache duplicate is fixed.

**Architecture:** the selector/redaction seams, opt-in capability class, per-session cache, off-loop execution, scheduler lifecycle, final-path containment, and per-job Watch claim are now sound. Merge safety still fails on the assembly staging escape and on clean-result claims from incompletely modelled harness inputs.

**Stack/CI:** #424 is mergeable and GitHub reports CLEAN. Its base #423 is one commit ahead but merges cleanly; all current pre-commit, CPU, and chat checks pass. GPU pytest and workload regression are skipped, so hardware behavior is not CI-validated.

src/aorta/chat/tools/cluster.py
```diff
@@ -0,0 +1,802 @@
...
+    is worse than no verdict.
+
+    A finer timestamp narrows that window without closing it. ``mkdtemp``
+    creates with ``O_EXCL`` and retries on a clash, so the directory is this
+    call's or the call raises, and the plain filenames inside it cannot
+    collide because nothing else can reach them.
+    """
+    parent.mkdir(parents=True, exist_ok=True)
+    stamp = datetime.now().strftime("%Y%m%d-%H%M%S")
+    return Path(tempfile.mkdtemp(prefix=f"{name}-{stamp}-", dir=parent))
```
amd-vivekag: **[High]** The assembly kernel name can escape this staging root.

A complete unit supplies `prepared.kernel` via `.amdhsa_kernel\s+(\S+)`; a name such as `../../outside` becomes the `mkdtemp` prefix, and `tempfile` creates it outside `parent`.

Fix: sanitize/validate `name` to one filename component before using it as either a prefix or filename, then verify the created directory resolves under `parent`.

src/aorta/chat/tools/harness/kernel.py
```diff
@@ -0,0 +1,410 @@
...
+        warned about a launch that is nothing like single-wave.
+        """
+        return self.wrapped and self.threads <= WAVEFRONT
+
+
+#: A branch whose condition reads one of the kernel's own inputs. Not an
+#: exhaustive parse -- it wants the shape ``if (input[i] > 0)``, which is what
+#: makes a race data-dependent, and would rather name a kernel that turns out
+#: to be fine than stay quiet about one that is not.
+_GUARD = re.compile(r"\b(?:if|while)\s*\(([^)]*)\)", re.S)
```
amd-vivekag, line 384: **[High]** The clean-result caveat still misses common data-dependent paths.

This only sees direct pointer indexing inside `if`/`while`; `active = input[i] > 0; if (active) ...`, scalar `mode == 1`, ternaries, and `switch` remain unqualified false-clean runs.

Fix: treat every generated-input clean result as inconclusive unless representative arguments are explicitly supplied, or require a caller-provided `main()` instead of relying on a regex to prove independence.

src/aorta/chat/graph/nodes.py
```diff
@@ -431,6 +543,261 @@ async def router_node(state: AgentState) -> dict[str, Any]:
...
+    # The context is copied across by hand because run_in_executor, unlike
+    # to_thread, does not. Without it the per-conversation ToolCache and the
+    # redaction notice would be unbound inside the worker, and the cache would
+    # silently fall back to the process-wide one -- which is the cross-session
+    # bug it was introduced to fix.
+    loop = asyncio.get_running_loop()
+    context = contextvars.copy_context()
+    started = time.monotonic()
+    try:
+        return await loop.run_in_executor(
```
amd-vivekag: **[Medium]** Cancelling the chat task does not stop the running tool.

Cancelling the awaited asyncio future returns through `finally` and announces `done`, but a running executor callable continues; a disconnected session can therefore leave a triage and cluster allocation alive until its internal timeout.

Fix: give long tools a cancellation handle propagated into `run_triage`, and emit `done` only after the worker actually exits.

src/aorta/chat/tools/harness/kernel.py
```diff
@@ -0,0 +1,410 @@
...
+MIN_ELEMENTS = 4096
+
+_KERNEL_SIG = re.compile(r"__global__\s+(?:[\w:]+\s+)*?void\s+(\w+)\s*\(", re.MULTILINE)
+_SHARED_DECL = re.compile(r"__shared__\s+[\w:]+\s+\w+\s*\[\s*([^\]]+?)\s*\]")
+_DEFINE = re.compile(r"^\s*#\s*define\s+(\w+)\s+(\w+)\s*$", re.MULTILINE)
+#: A kernel that reads threadIdx.y is written for a two-dimensional block, and
+#: launching it one-dimensionally does not fail -- it runs with that index
+#: pinned at zero, touching a different set of addresses than the author wrote.
+_USES_DIM = {
+    "y": re.compile(r"\b(?:threadIdx|blockDim|blockIdx)\.y\b"),
```
amd-vivekag: **[Medium]** Grid dimensions are still mistaken for block dimensions.

`blockIdx.y/z` passes this check when `block_y/z` is supplied, but line 259 still launches a scalar grid, so those indices remain zero and a different execution can look clean.

Fix: track block and grid axes separately, add `grid_y/grid_z`, and reject either missing geometry rather than mapping `blockIdx` to `block_*`.

src/aorta/chat/plugins.py
```diff
@@ -72,12 +72,75 @@
...
+    Absent is a shipped configuration, not a fault: chat without [cia] is a
+    working assistant over the codebase and past runs, minus the tools that
+    submit jobs. It says so once rather than leaving the reader to notice five
+    tools missing from a catalogue.
+    """
+    from aorta.chat.config import settings
+
+    try:
+        from aorta.chat.tools import cluster
+    except ImportError as exc:
```
amd-vivekag, line 137: **[Medium]** This hides every built-in import bug as a missing optional extra.

An incompatible or broken internal import silently removes all diagnostic tools and tells the operator to install `[cia]`, contrary to this registry's rule that broken built-ins raise.

Fix: suppress only the expected missing DSPy module; re-raise other `ImportError`/`ModuleNotFoundError` failures.

amd-vivekag (commented):
## Re-review summary

**Verdict: ship**

No merge-blocking findings remain. The five inline findings from the last review are fixed correctly: assembly-derived staging names are contained, every generated-input clean result is qualified as inconclusive, chat cancellation drains the running triage and scheduler work, grid/block geometry is separated, and only the expected missing-DSPy import is suppressed.

**Blocking findings:** 0 High · 0 Medium

**Prior reconciliation:** 33 of 34 prior human inline findings are fixed correctly; none remain partial. The one unaddressed item is the installation-guide gap below, which is independently deliverable and does not make the current implementation unsafe or incorrect because runtime diagnostics already name `[cia]` and configuration.md documents the capability. All 17 Copilot inline findings and its suppressed assembly-cache duplicate are fixed correctly. GitHub still shows old threads as unresolved mechanically, but their code concerns no longer apply at this head.

**Follow-up issues to create (non-blocking)**

1. **Document installing CIA-backed chat diagnostics**
   - Scope: update `docs/chat/installation.md` and the chat README to explain combining `chat-cli`/`chat-ui`/`chat-all` with `[cia]`, and link the existing cluster configuration/security sections.
   - Acceptance: include pip and editable-install commands; state Python support and `allow_cluster_jobs=false` default; distinguish read-only report tools from submitting tools; show `aorta chat tools` verification and the missing-extra behavior.

2. **Add hardware smoke coverage for CIA chat diagnostics**
   - Scope: add a GPU-gated smoke lane for kernel, assembly, and workload triage without expanding this PR's implementation.
   - Acceptance: each tool acts on the supplied artifact; sanitizer/waitcheck reports `ran` rather than `not_checked`; cancellation releases the allocation; produced job/report artifacts validate against their schemas; the aggregate check reports clearly when hardware is unavailable.

**Architecture:** merge-safe. The registry-derived selector, redaction chokepoints, opt-in execution capability, contained per-invocation staging, per-session caches, async cancellation, scheduler lifecycle, honest harness verdicts, separated launch geometry, and bounded/per-job Watch admission now form coherent boundaries.

**Stack/CI:** #424 is based directly on current #423 (`0f6eeefb`), is 113 commits ahead, and GitHub reports CLEAN/MERGEABLE. Pre-commit, CPU, chat, CIA Python 3.10-3.14, and the combined chat-all+CIA lane pass. GPU pytest and workload regression are skipped, which is why hardware automation is proposed separately.

I am posting COMMENT rather than APPROVE because the prior installation-doc finding is not yet fixed or represented by a tracking issue; it is safe to merge once that follow-up is recorded.

amd-vivekag (approved):
## Approval

**Verdict: ship**

My September 18 review reached the same verdict but was posted as COMMENT, because the installation-doc finding was neither fixed nor recorded as a tracking issue. Both follow-ups I named have since been delivered in the PR itself rather than deferred, so the reason for withholding is gone.

**Follow-up 1 — document installing CIA-backed chat diagnostics.** `8647fdf0` adds "Add CIA-backed cluster diagnostics" to `docs/chat/installation.md` and the matching entry point in `docs/chat/README.md`. Against the acceptance criteria I wrote: published and editable installs for all three interfaces, the 3.10–3.14 headless range with the narrower `chat-cli` 3.11–3.14 and `chat-ui`/`chat-all` 3.11–3.13 ceilings, the `allow_cluster_jobs = false` default, the read-only pair separated from the three submitting tools, and `aorta chat tools` / `--json` verification with the missing-extra behaviour and its remedy. `tests/chat/test_cia_installation_is_documented.py` pins each of those, so the guide cannot drift back to the three-extra description. The three cross-links resolve: `configuration.md#the-cluster-diagnostic-tools`, `extending.md#the-exception-and-why-it-is-one`, and `redaction.md` all exist at this head.

**Follow-up 2 — hardware smoke coverage.** The GPU-gated lane arrived on the base branch; `a025cd86` is this PR's adaptation of it to the `tolerate_nonzero` parameter that replaced the process-global environment flag, which is the right direction — the smoke test exercises the new signature rather than the old side effect.

**Re-verification at `a025cd86`:** the only two commits on this branch since my last reviewed head (`3846c562`) are those two; everything else in the range came in through the merges from `main` after #423 landed. All 22 checks pass, including both GPU legs this time — `pytest (GPU, MI350)` and `workload regression (GPU, MI350)` now run rather than skip, which closes the hardware gap I flagged as unvalidated in the last two reviews. GitHub reports CLEAN / MERGEABLE against `main`.

**One non-blocking nit, for whenever this file is next touched:** the `if "tolerate_nonzero" in inspect.signature(launch).parameters` guard in `tests/cia/test_cia_hardware_smoke.py` is always true at this head, since `launch()` declares the parameter with a default. A test feature-detecting its own branch's API will read as a genuine compatibility requirement to the next person, and the conditional would silently stop passing the flag if the parameter were ever renamed. Passing it unconditionally is the clearer statement. Not worth a push on its own.

Older threads above are still mechanically unresolved; their code concerns do not apply at this head.

PR #425, opened 2026-09-02
feat(chat): show the work, and run tools off the event loop

aviator19941:
Last of three. Stacked on #424, which is stacked on #423 — **review those
first**; GitHub retargets each as its parent merges. Plan:
`docs/plans/cia-agents-port.md`.

## Showing the work

A diagnostic tool holds the act node for minutes. Steps that appear only when a
node *finishes* therefore say nothing for the entire part of a run the user is
waiting through — which is indistinguishable from a hang, and was reported as
one during the demo.

Two mechanisms, both proven in the source repo before being ported:

- `invoke_agent` takes an optional callback and streams the graph
  (`stream_mode=["updates", "values", "custom"]`), so each node renders a step
  as it completes.
- **A tool announces itself before it blocks**, through the custom stream. This
  is the one that matters, and the ordering is asserted directly rather than
  implied: an event *after* the call would be a log line; an event *before* it is
  what lets the UI name what it is waiting on while it waits.

Omitting the callback awaits the graph exactly as before. The CLI depends on
that, and a test asserts the streamed and awaited paths return the same reply,
the same state keys and the same history.

## Running tools off the event loop

Both act paths called `_execute_tool` inline from an async function. That is
free for `list_files` and fatal for a tool that blocks for minutes: the loop
stalls, Chainlit stops answering, and the browser reports the backend as
**unreachable** rather than busy.

This was survivable before the port and is not survivable now — PR 2 added
tools that take 150 to 350 seconds. `_execute_tool_async` puts them on a
worker thread, and a test asserts the tool body does not run on the caller's
thread.

## The answer

Three labelled parts whenever a diagnostic tool has run: **the bug** (in the
user's own code), **how we found it** (which tool, and the evidence — the
signal, the file and line, the confidence), and **the fix** (quotable verbatim).

The middle part is the one that gets dropped, and without it a verdict is an
assertion: the reader cannot tell a collision that was *observed* from a hazard
*inferred* off a path that may never execute, and those deserve different
responses. Which is also why the confidence is reported rather than rounded up —
0.55 and 0.95 are both real answers about real bugs, and presenting them
identically discards the only signal that says how hard to trust them.

The tests assert the instruction, not a model's compliance with it; an offline
suite cannot do the latter. What they can do is fail when the instruction is
dropped, which is how it would actually be lost. One of them also fails if the
three-part rule ever smuggles routing back in ("for NaN, use…"), since that
would make the selector decoration.

## Tests

14 offline. The 5 `integration` selector tests from #424 still pass against a
real model after these act-path changes.

Suite: 875 passing. The 46 failures are the same 8 files that fail on
`feat/aorta-agent` for want of the RAG and sqlite extras — confirmed in a
worktree, unchanged by this PR.


src/aorta/chat/graph/nodes.py
```diff
@@ -57,6 +60,7 @@
 
 RETRIEVED CONTEXT:
 {context}
+12. When a diagnostic tool has run, answer in three labelled parts: the bug (what is wrong, in the user's own code); how we found it (which tool, and the evidence it returned -- the signal, the file and the line, the confidence); and the fix (the change, quotable verbatim). Report the confidence the tool gave rather than rounding it up: a static finding on a path that may never execute is worth less than a collision that was observed, and saying so is the difference between a report an engineer can act on and one they have to re-derive.
```
amd-vivekag: **[High]** Rule 12 is outside the rules — it lands after the retrieved context.

The prompt ends `RULES: 1…11`, then `RETRIEVED CONTEXT:`, then `{context}`, and rule 12 is appended after all of it. So the three-part instruction is separated from the list it is numbered into and sits below however many kilobytes of retrieved code chunks got interpolated — the worst position in the prompt for an instruction you need followed.

`test_answer_shape.py` cannot catch this: every assertion is a substring check against `SYSTEM_PROMPT`, which passes wherever the text sits. That is the gap that let it through, given the PR's own framing that the tests "fail when the instruction is dropped".

Fix: move the line up to sit after rule 11 and before `RETRIEVED CONTEXT:`, and add an assertion that it precedes the context marker — e.g. `assert SYSTEM_PROMPT.index("three labelled parts") < SYSTEM_PROMPT.index("RETRIEVED CONTEXT")`.

.chainlit/config.toml
```diff
@@ -0,0 +1,178 @@
...
+# Whether to persist user environment variables (API keys) to the database
+# Set to true to store user env vars in DB, false to exclude them for security
+persist_user_env = false
+
+# Whether to mask user environment variables (API keys) in the UI with password type
+# Set to true to show API keys as ***, false to show them as plain text
+mask_user_env = false
+
+# Authorized origins
+allow_origins = ["*"]
```
amd-vivekag, line 23: **[High]** `allow_origins = ["*"]` on a UI that compiles and runs pasted code.

This commits a wildcard CORS policy for a server whose tools submit cluster jobs, run `run_terminal_command`, and compile user-supplied HIP. Any page a developer visits can talk to a locally-running instance. `mask_user_env = false` three lines up also renders API keys as plain text in the UI.

The PR body does not mention this file at all — 178 lines of generated Chainlit config arriving unannounced in the same diff as the streaming work, and `.chainlit/` is not in `.gitignore`.

Fix: set `allow_origins` to the origins the UI is actually served from (`["http://localhost:8000"]` for local use), set `mask_user_env = true`, and either explain the file in the PR body or gitignore it and document the settings that matter.

src/aorta/chat/session.py
```diff
@@ -35,27 +37,51 @@ def extract_reply(messages: list[BaseMessage]) -> str:
...
+            # "updates" names the node that just ran and "custom" carries a
+            # tool announcing itself; "values" is the accumulated state, so the
+            # last one matches what ainvoke would have returned.
+            result = {}
+            async for mode, chunk in agent_graph.astream(
+                initial, stream_mode=["updates", "values", "custom"]
+            ):
+                if mode == "updates":
+                    for node, delta in chunk.items():
+                        await on_step(node, delta or {})
```
amd-vivekag, line 84: **[Medium]** An `on_step` failure now aborts the whole run.

These awaits are unguarded, so anything the callback raises propagates out of `astream` and kills the query. In the UI `show_step` calls Chainlit, which fails when the session has gone away — a user closing the tab or navigating during a five-minute cluster job takes the run down with it, and the tool keeps burning the job with nobody to receive the answer. The `ainvoke` path this replaces had no such coupling, so it is a regression for anyone passing a callback.

Fix: make progress reporting best-effort.

```suggestion
                        try:
                            await on_step(node, delta or {})
                        except Exception:
                            logger.warning("progress callback failed", exc_info=True)
```

src/aorta/chat/session.py
```diff
@@ -35,27 +37,51 @@ def extract_reply(messages: list[BaseMessage]) -> str:
...
-                "iteration": 0,
-            }
-        )
+        if on_step is None:
+            result = await agent_graph.ainvoke(initial)
+        else:
+            # "updates" names the node that just ran and "custom" carries a
+            # tool announcing itself; "values" is the accumulated state, so the
+            # last one matches what ainvoke would have returned.
+            result = {}
```
amd-vivekag, line 78: **[Medium]** If no `values` chunk arrives, the reply is silently empty.

`result` stays `{}` and `extract_reply({}.get("messages", []))` yields an empty string that gets appended to history as an `AIMessage` — the user sees a blank answer with no error, and the transcript records it as one. The awaited path could not produce this, so it is only reachable through the new branch.

Fix: track whether a `values` chunk was seen and raise (or return a clear "the graph produced no state" message) instead of falling through with `{}`.

src/aorta/chat/graph/nodes.py
```diff
@@ -325,6 +329,26 @@ async def router_node(state: AgentState) -> dict[str, Any]:
...
+
+    The announcement is what makes a five-minute tool visible: steps otherwise
+    appear only when a node finishes, so the one node that takes real time is
+    the one that says nothing.
+    """
+    try:
+        get_stream_writer()({"tool": _normalise_tool_name(tool_name), "args": kwargs})
+    except RuntimeError:
+        pass  # not being streamed; nothing to announce to
+    return await asyncio.to_thread(_execute_tool, tool_name, kwargs)
```
amd-vivekag, line 510: **[Medium]** `to_thread` work cannot be cancelled, and it stacks on #424's abandoned executor.

`asyncio.to_thread` runs on the default executor with no cancellation path: cancelling the awaiting task returns control but the tool keeps running to completion, so an aborted turn still holds a cluster job and a thread. Combined with `_run_triage`'s own per-call `ThreadPoolExecutor`, one triage occupies two non-daemon threads, and both loop's shutdown handlers join them — so the server cannot exit until a multi-minute tool finishes.

The default executor's `max_workers` is `min(32, cpu+4)`, which is not a meaningful bound for 150–350s tools; enough concurrent turns and unrelated `to_thread` work starves.

Fix: run these on one explicitly-sized executor via `loop.run_in_executor`, and give the diagnostic tools a cooperative cancel (a stop event their poll loops check) so an abandoned turn releases its job.

src/aorta/chat/graph/nodes.py
```diff
@@ -325,6 +329,26 @@ async def router_node(state: AgentState) -> dict[str, Any]:
...
+    while a cluster job runs. Calling one inline stalls the loop, which under
+    the web UI means the server stops answering and the browser reports the
+    backend as unreachable rather than busy.
+
+    The announcement is what makes a five-minute tool visible: steps otherwise
+    appear only when a node finishes, so the one node that takes real time is
+    the one that says nothing.
+    """
+    try:
+        get_stream_writer()({"tool": _normalise_tool_name(tool_name), "args": kwargs})
```
amd-vivekag, line 507: **[Low]** The announcement carries `args`, which nothing reads.

`show_step` in `ui/app.py` uses only `delta.get("tool")`. Meanwhile `kwargs` for `triage_kernel_source` is the user's entire pasted kernel, so every tool start pushes the full source through the custom stream for no consumer.

Fix: send just the tool name, or a short summary of the argument keys, and add the payload back when something renders it.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

The two mechanisms are right and the tests are pointed at the right properties — asserting that the tool event is observed *while* the tool is still running, rather than that it eventually appears, is the distinction that matters, and the same-reply-either-way test on the awaited path is the right guard for the CLI. `_execute_tool_async` fixes a real bug: with #424's 150–350s tools, an inline call really would make the browser report the backend as unreachable.

Two things to fix before it lands. Rule 12 is appended *after* `RETRIEVED CONTEXT:` and `{context}`, so the three-part instruction sits below the interpolated code chunks instead of in the rules list it is numbered into — and `test_answer_shape.py` is a set of substring checks, so it passes regardless of position. And `.chainlit/config.toml` arrives with `allow_origins = ["*"]` and `mask_user_env = false`, unmentioned in the PR body, on a UI that runs pasted code on GPU nodes.

The rest is error handling around the new streaming path: `on_step` is awaited unguarded, so a Chainlit failure mid-run kills the query, and `result` defaults to `{}`, which turns a missing `values` chunk into a silently empty reply.

**Findings:** 2 High · 3 Medium · 1 Low

**High**

- `chat/graph/nodes.py:63` — rule 12 placed after the retrieved context, outside `RULES`; the test cannot detect it.
- `.chainlit/config.toml:23` — `allow_origins = ["*"]` (and `mask_user_env = false`) committed without mention.

src/aorta/chat/session.py
```diff
@@ -35,27 +37,51 @@ def extract_reply(messages: list[BaseMessage]) -> str:
...
+            async for mode, chunk in agent_graph.astream(
+                initial, stream_mode=["updates", "values", "custom"]
```
amd-vivekag, line 80: **[Medium]** Streaming the graph is a contract change the aorta_llm integration docs still forbid.

aorta_llm `ARCHITECTURE.md` §2.1 / §4: the web UI shows a `Thinking...` placeholder and then *replaces* it with one complete answer — "not streamed token by token." `invoke_agent` was the single convergence point for UI and CLI, and it `ainvoke`d. This PR splits that: callback → `astream`, else `ainvoke`.

The CLI path is preserved (good; the test is right). The docs are not. `docs/chat/README.md` still describes the old delivery, and nothing says that a UI session now emits per-node steps and that tool args ride a custom stream.

Fix: update the integration docs to match (Chainlit steps, custom stream for tool-start, CLI unchanged). Keep token-level streaming off, which this still does.

src/aorta/chat/graph/nodes.py
amd-vivekag: **[Medium]** The three-part diagnostic answer is bolted onto a prompt that still describes the old aorta_llm product.

Rule 1 still says "Only answer questions about the AORTA codebase. Politely refuse anything else." Rule 3 still lists only the sandboxed file/search/run-artifact tools. The new rule 12 then asks for a labelled bug/evidence/fix about *the user's own code* from a diagnostic tool that this prompt never named.

A model following the integration prompt to the letter will refuse a pasted-kernel question that #424 just taught the selector to route. A model following rule 12 will diagnose it. That is two products in one system prompt.

Fix: rewrite rules 1 and 3 to match what chat can now do (codebase Q&A *and* a pasted-workload diagnosis), name the three cluster tools, and keep the refuse-unrelated-work rule for things that are neither. Then put rule 12 with the other rules, not after `{context}`.

src/aorta/chat/ui/app.py
```diff
@@ -63,8 +95,28 @@ async def on_message(message: cl.Message):
     thinking_msg = cl.Message(content="Thinking...")
     await thinking_msg.send()
```
amd-vivekag, line 123: **[Low]** `Thinking...` is still sent, then removed, *and* steps are appended — so the UI no longer matches either documented shape.

ARCHITECTURE.md: one placeholder, then one replacement message. This PR: placeholder + a stack of `cl.Step` + a final message after `thinking_msg.remove()`. Fine as UX, but `docs/chat/` has no description of steps, and a reader of the architecture doc will think the UI still works the old way.

Fix: drop the placeholder once steps exist (the first step *is* the thinking indicator), and document the step list in `docs/chat/README.md`.

amd-vivekag (commented):

**Verdict: still fix-before-ship** (the High items remain the ones from the first pass; these are design-contract gaps on top)

#423/#424 already break the aorta_llm integration's single provider, single redaction gate, and tool sandbox. This PR then changes the *delivery* contract those docs describe — `Thinking...` then one replacement message, UI and CLI sharing `ainvoke` — without updating `docs/chat/` or the system prompt that still says chat only answers questions about the AORTA codebase.

**New findings:** 0 High · 2 Medium · 1 Low

**Medium**
- `invoke_agent` `astream` vs ARCHITECTURE.md §2.1 (placeholder-then-replace, not streamed). CLI path is kept; the docs are not.
- Rule 12 added onto a prompt whose rule 1 still refuses non-codebase questions, which is the diagnostic product fighting the aorta_llm product.

Please treat `docs/chat/*` as part of this stack's diff. The integration guides are how the aorta_llm merge was reviewed; leaving them describing the pre-CIA assistant is how the next port will re-learn these collisions.

amd-vivekag:

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

Moving synchronous tools off the event loop is necessary, and the optional callback preserves the CLI path cleanly. The prompt-placement fix is correct. This still cannot ship safely: `get_stream_writer` violates the declared LangGraph minimum, and concurrency turns #424's shared staging file and process-global environment mutation into cross-session races.

**Findings:** 2 High · 3 Medium · 3 Low

**Prior-review reconciliation**
- Fixed correctly: the diagnostic answer rule is now before retrieved context, with placement tests; the UI notice doubles accept the callback.
- Still open: callback failures abort the graph while work continues, executor work has no cooperative cancellation/admission control, the system prompt describes the old codebase-only product, streamed args are unused, docs still describe the pre-CIA UI, and `Thinking...` remains redundant with steps.
- I confirmed all five current Copilot findings: dependency floor, concurrent shared-state races, cwd-dependent/unpackaged Chainlit config, Markdown fence breakage, and the immediately-completed tool step. I did not duplicate them inline.

- `pre-commit`: inherited EOF fixer failure from #423.
- Chat Python 3.11–3.14: inherited eager CIA/DSPy import from #424; every required leg is red with broad collection failures.
- CPU Python 3.10–3.14: inherited CIA tests without `[cia]` plus the live selector test without LangChain.
- Therefore this PR's new streaming tests are not validated by the required green stack. Fix the optional dependency/test partition in #423/#424, raise `langgraph` to the first version supporting `get_stream_writer` and custom streams, then rerun the full stacked matrix.

**Architecture integration**
Keep the callback convergence point, but replace stringly typed node/delta events with a small typed progress-event contract carrying correlation IDs and start/completion states. Run long diagnostics through one bounded executor/job manager with cancellation and per-session staging; do not nest the default executor around a per-call executor. Update `docs/chat/*` with the expanded diagnostic scope, step lifecycle, and unchanged CLI behavior.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The callback convergence point, unchanged CLI path, and off-loop intent are good foundations, and the answer-rule placement fix is correct. Approval is not safe: three High blockers remain, the progress lifecycle is misleading, and the entire stacked required matrix is red.

**Findings:** 3 High · 5 Medium · 4 Low

**Still open High items**
- `src/aorta/chat/graph/nodes.py:27` imports `get_stream_writer`, but `pyproject.toml:163` still permits LangGraph 0.2.0; valid 0.2.0–0.2.68 installs fail at import. Raise the floor to `>=0.2.69` and test the minimum.
- `src/aorta/chat/graph/nodes.py:510` enables concurrent execution while kernel and assembly staging still use second-resolution shared names (`cluster.py:361-363`, `499-502`). Same-named submissions in one second can overwrite each other's source/object. Use per-invocation UUID/temp directories with exclusive creation.
- `.chainlit/config.toml:23` still allows every browser origin for an unauthenticated UI that can submit cluster work. Restrict origins and make the shipped UI reliably load that policy.

**Other open items**
- `session.py:84,86`: callback failures abort the graph while work can continue.
- `ui/app.py:129-135`: the tool step exits immediately and appears complete before the tool runs.
- Cancellation is split across nested executors and does not propagate to the scheduler.
- The system prompt still describes a codebase/run assistant while later rules require user-code diagnosis.
- Chainlit config is cwd-dependent and not packaged; Markdown fences, unused streamed arguments, stale docs, and redundant `Thinking...` remain.

Prior reconciliation: 4 earlier findings are fixed correctly, 2 are only partial, and the remaining lifecycle/dependency/UI findings are still open. All current Copilot comments were rechecked; the concurrency finding narrows to kernel/assembly staging but is not resolved.

**CI**
- `pre-commit` inherits #423's EOF/line-ending failure.
- CPU Python 3.10–3.14 inherits the CIA dependency/test-partition failures.
- Chat Python 3.11–3.14 inherits #424's missing-`[cia]` registration/collection failures (18 failures + 14 errors on representative legs).
- CI uses a newer LangGraph, so it does not test or catch this PR's minimum-version import break.

**Architecture integration:** moving blocking work off the event loop is necessary, but a durable design needs typed start/completion progress events, per-request staging, one bounded executor/job manager, end-to-end cancellation through Slurm, and UI transport failures decoupled from graph survival. The integration is not yet solid.

src/aorta/chat/tools/cluster.py
```diff
@@ -301,7 +317,11 @@ def _run_triage(extra_args: list[str], label: str) -> str:
         return (f"Triage failed at stage {result.get('stage', '?')}: "
                 f"{result.get('error', 'unknown error')}")
 
-    return _format_result(result, label)
+    # A --recipe run is a sanitizer run by construction: write_asm_recipe asks
+    # for waitcheck and sets on_missing_backend=fail. So a report is owed, and
+    # its absence is a failure rather than a clean sheet.
+    expect_sanitizer = "--recipe" in extra_args
```
amd-vivekag: **[High]** Kernel triage does not require the sanitizer it claims to run.

`triage_kernel_source` passes `--source`, so a missing sanitizer report skips the new DID-NOT-RUN warning and an unrelated Autopsy result can still be cached as completed.

Fix: require sanitizer completion for both `--source` and `--recipe`, and cache only when that expected report exists.

src/aorta/chat/ui/chainlit_config.toml
```diff
@@ -0,0 +1,201 @@
...
+#
+# The origins this UI is actually served from, not "*". The tools behind it
+# submit cluster jobs, compile pasted HIP and can run shell commands, so a
+# wildcard means any page a developer happens to have open can talk to a
+# local instance and start work on a GPU node.
+#
+# Serving on another host or port means adding it here. `aorta chat ui --port`
+# and the PORT variable in the launcher both change the port, and neither can
+# reach into this file.
+allow_origins = [
```
amd-vivekag, line 42: **[High]** The shipped origin policy rejects the default UI.

These origins use port 8010, while `aorta chat ui` defaults to 8000, so the default browser socket/API origin is not allowed.

Fix: derive/validate origins from the effective host and port, or align both defaults and add a cross-contract test.

src/aorta/chat/graph/nodes.py
```diff
@@ -487,6 +558,98 @@ async def router_node(state: AgentState) -> dict[str, Any]:
...
+    # The context is copied across by hand because run_in_executor, unlike
+    # to_thread, does not. Without it the per-conversation ToolCache and the
+    # redaction notice would be unbound inside the worker, and the cache would
+    # silently fall back to the process-wide one -- which is the cross-session
+    # bug it was introduced to fix.
+    loop = asyncio.get_running_loop()
+    context = contextvars.copy_context()
+    started = time.monotonic()
+    try:
+        return await loop.run_in_executor(
```
amd-vivekag: **[Medium]** Cancelling this coroutine does not stop the worker or scheduler job.

The `finally` immediately emits `done=True` while `_execute_tool` can keep running for the full triage timeout.

Fix: pass a cancellation token through tool execution to `run_triage`, issue scheduler cancellation, and emit completion only when the worker actually ends.

src/aorta/chat/ui/app.py
```diff
@@ -62,6 +219,123 @@ async def on_start():
...
+        except UnicodeDecodeError:
+            notes.append(
+                f"`{name}` is not text. A .hsaco has to be disassembled first; "
+                "attach the listing."
+            )
+            continue
+        if not text.strip():
+            notes.append(f"`{name}` is empty.")
+            continue
+        if len(raw) > _INLINE_ATTACHMENT_BYTES:
```
amd-vivekag, line 314: **[Medium]** Every large accepted source file is routed as assembly.

`.hip`, `.cpp`, and `.py` uploads over 32 KiB are told to use `source_file`, but only `triage_assembly_source` accepts that argument.

Fix: route staging by suffix and add file arguments to kernel/workload tools, or restrict this path to assembly listings.

.github/workflows/chat-tests.yml
```diff
@@ -275,21 +275,77 @@ jobs:
...
+          # declared minimum cannot run the code that declares it.
+          from langgraph.config import get_stream_writer
+          from aorta.chat.graph import nodes
+          from aorta.chat import session
+          print("get_stream_writer and the graph import on the floor")
+          PY
+
+      - name: The custom stream mode works on the floor
+        run: |
+          python -m pytest tests/chat/test_langgraph_floor.py tests/chat/test_streamed_steps.py -q
```
amd-vivekag, line 353: **[Medium]** The new floor gate cannot run its own tests.

CI reaches this line and fails with `No module named pytest`, so custom-stream compatibility at the declared floor remains unvalidated.

Fix: install `.[chat-cli,tests]` together with the pinned LangGraph floor.

tests/chat/test_prompt_describes_the_product.py
```diff
@@ -0,0 +1,170 @@
...
+    Rule 1 was widened to take pasted GPU code while the sentence above it
+    still said "the AORTA codebase" and nothing else. A model reading the two
+    in order is introduced to one product and then given the rules of another,
+    and the identity sentence is the one it leans on when the rules are
+    ambiguous -- which is how a diagnostic request gets politely refused by
+    the assistant built to diagnose it.
+    """
+
+    def test_it_says_it_diagnoses_when_it_can(self, prompt):
+        assert "diagnoses" in _identity(prompt(cluster_jobs=True))
```
amd-vivekag, line 157: **[Medium]** This CIA-dependent assertion runs in the supported no-CIA chat matrix.

Unlike the surrounding tests it does not skip without DSPy; the staging-concurrency tests have the same boundary error, producing five failures in each chat-only leg.

Fix: move CIA assertions to an explicit `[cia]` lane or make them dependency-neutral and test no-CIA behavior separately.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

This revision correctly fixes the LangGraph floor declaration, per-invocation staging, normal tool-step lifecycle, callback isolation, Markdown fencing, packaged Chainlit config, prompt identity, and streaming documentation. Two High correctness/availability defects remain: kernel triage can still present a missing sanitizer as a completed diagnosis, and the hardened CORS configuration rejects the launcher's default port.

**Findings:** 2 High · 5 Medium

**Prior reconciliation:** the concurrency/staging and prompt-placement High items are fixed correctly. The CORS item is partial: wildcard/key exposure is fixed, but the configured origin does not match the CLI. Executor cancellation is also partial: normal completion is correlated, while cancelling the coroutine closes the step before the worker/Slurm job stops.

The outer four-worker tool pool also blocks two workers waiting on the inner two-worker triage pool, so a burst can consume the capacity intended for quick tools (`chat/tools/cluster.py:290`).

**CI:** pre-commit and CPU pass, but all chat matrix legs, the LangGraph floor job, and the Chat aggregate are red. The floor job never runs its behavior tests because `pytest` is not installed; the chat-only matrix also runs CIA-dependent tests without `[cia]`.

**Architecture integration:** the progress callback seam, context propagation, bounded staging, and normal start/completion events are sound. The stack is not solid until sanitizer completion is part of the result contract, UI origin policy shares configuration with the launcher, and cancellation/admission are owned by one end-to-end job manager.

vivekkhandelwal1:
Reading through this stack I noticed one thing the chat does not do: it keeps nothing. The browser holds its history in `cl.user_session` with no data layer behind it, and the CLI prints `--json` to stdout and remembers only the lines that were typed. So a turn that ranked the tools, submitted a job and explained a race leaves no trace of having decided any of it.

The cost that struck me is the demo. Someone asks why it reached for the assembly tool instead of the kernel one, and the answer has already scrolled away. Same when a turn goes wrong — the selector's reasoning is the first thing you would want to re-read, and it is exactly what is gone. And a reviewer of #424 can see the design but not the behaviour: that the selector ranks on the evidence a tool returns rather than on symptom keywords is the interesting claim in that PR, and the only support for it is the five integration tests plus whatever the person who ran it remembers.

So I wrote a patch for it, on the chance it is useful. Take it, adapt it, or ignore it — it is your PR, there is no hurry on my side, and ignoring it is a perfectly fine outcome. I have not pushed anything or opened anything. It is a `git format-patch` against `ab50862`, which was the tip a couple of hours ago; you are moving this branch several times a day, so `git am -3` for the drift. Happy to open it as a PR against `feat/cia-reporting` if you would rather read it as a diff, or to send it over some other way, or to drop it.

What it does: one JSON line per event, appended to a file per session, holding what the assistant decided. The route, the tools the selector ranked and its own sentence for them, which tools ran in which order, the CIA job ids they produced and the category and confidence each of those jobs reached, and whether the critic accepted the answer. It hangs off `invoke_agent` so both front doors get it from one place, and it reads the state the graph returns rather than the progress stream — which is what keeps the CLI's no-callback path awaited, and recording the same thing as the browser.

The confidentiality design is really your rule rather than mine. `_announce_tool` in this PR carries the tool's name and not its arguments, because for `triage_kernel_source` those arguments are the user's whole kernel; I took that as the policy for the whole record. Nothing the user pasted is written down — the question, the plan, the tool arguments and the tool output are each reduced to a length, a line count, a fence count and a digest. The one exception is the selector's reason, because "why this tool" is the point of recording a decision at all, and that goes through the same path and address scrubbers `aorta bundle` uses, ungated so `--no-redact` cannot widen it. Files are `0600` under `$XDG_STATE_HOME` in a `0700` directory, and nothing reads them back or sends them anywhere. There is a `full` mode that keeps everything verbatim, for when you know what is being pasted; it warns once per session naming the file.

Three things to make it cheap to judge. It is off unless someone sets `AORTA_CHAT_SESSION_LOG=1`, so taking it changes nothing until that happens. It adds 39 tests, which is 39 added passes against `ab50862` with no existing outcome changed — 1682 to 1721, the same ten pre-existing failures either way, all of them from optional extras my environment is missing rather than from the code. And I ran a real turn with a kernel body, a customer filesystem path and a host address planted in it, then grepped the resulting file for all three: none appear, and the path that had reached the selector's reason came back out as `<PATH:0>`.

The limit, rather than leaving you to find it. Autopsy explains a cause; nothing applies a mitigation and nothing re-runs a workload to check that a fix worked, so `next_probes` is a suggestion no code executes. This therefore records decisions with no ground truth about whether anything was actually resolved. There is a `resolution` field in the schema that is null on every record the patch can write, with a stable `(session_id, turn)` key and the job ids beside it so an outcome can be attached later from an archived probe matrix or from a person. It is a decision log, and something else has to supply the other half.

Last, and it is our reason rather than yours: the RL post-training work needs real debugging trajectories, and a turn's decision sequence is the policy-step data it has no source for today. That falls out of the logging rather than shaping it — I did not add a field for it, and if you cut something for your own reasons I would rather have your version than mine.

src/aorta/chat/ui/app.py
```diff
@@ -36,14 +41,166 @@ def _unavailable_message(reason: str) -> str:
...
+
+    async def handle(self, delta: dict) -> None:
+        """Open a step on a tool's announcement, close it on its completion."""
+        name = delta.get("tool")
+        # Keyed by call rather than name: a turn runs the same tool more than
+        # once with different arguments, and closing by name would end the
+        # wrong one.
+        call = str(delta.get("id") or name)
+        if delta.get("done"):
+            await self._finish(call, name, delta.get("seconds"))
```
amd-vivekag: **[Medium]** Cancellation status is discarded, so a still-running job renders as finished.

The producer now sends `cancelled="stopped"` or `"still running"`, but this call drops it and `_finish` always writes `finished`.

Fix: pass the cancellation status into `_finish`, render stopped vs. cancellation-requested/still-running distinctly, and cover both payloads in the UI lifecycle tests.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

This revision resolves every prior High finding: prompt placement, LangGraph floor compatibility, per-invocation staging, sanitizer-required semantics, and the shipped/default CORS policy. It also fixes callback isolation, no-state handling, packaged Chainlit config, streamed arguments, Markdown fencing, normal tool-step lifecycle, attachment routing, CIA test partitioning, and quick-tool starvation.

**Findings:** 0 High · 2 Medium · 0 Low

**Prior reconciliation:** all prior findings are fixed correctly except cancellation lifecycle, which is partial: cancellation now reaches triage/Slurm and the producer distinguishes `stopped` from `still running`, but the UI discards that status and always renders `finished`. All five Copilot findings are fixed correctly; the remaining lifecycle issue is the cancellation edge beyond Copilot's original normal-completion report.

**Architecture:** the callback seam, separate bounded pools, per-call cancellation token, per-invocation staging, and conditional prompt/tool registration now integrate coherently. The leaf is mergeable and its own stack is green; #423 remains behind main, so stack sequencing/rebase still belongs at the parent.

**CI:** pre-commit, CPU 3.10–3.14, chat 3.11–3.14 including `[cia]`, the LangGraph 0.5.3 floor, MI350 GPU tests, and workload regression all pass at this head.

src/aorta/chat/graph/nodes.py
```diff
@@ -1287,6 +1395,20 @@ async def _act_text(state: AgentState) -> dict[str, Any]:
...
+    """The user's own words this turn, for a node that needs to check them.
+
+    Truncated, because a pasted disassembly can be megabytes and the critic is
+    checking claims rather than reading the listing. The head is what carries a
+    learning rate, a launch config or a missing barrier.
+    """
+    for message in reversed(state.get("messages") or []):
+        if isinstance(message, HumanMessage) and isinstance(message.content, str):
+            text = message.content
+            return text if len(text) <= 4000 else text[:4000] + "\n... (truncated)"
```
amd-vivekag: **[Medium]** The critic cannot verify grounded code after the first 4,000 characters.

A valid cause or fix later in a pasted kernel/workload is removed here, so the critic can reject it as invented even though the answer read it from the user's code.

Fix: carry the exact cited excerpt/line as structured evidence, verify it against the full user text, and send that bounded relevant window to the critic instead of always keeping only the head.

src/aorta/chat/decision_log.py
```diff
@@ -0,0 +1,397 @@
...
+        {
+            **_base(session_id, turn, "route", mode),
+            "route": state.get("route"),
+        },
+        {
+            **_base(session_id, turn, "selection", mode),
+            "ranked_tools": list(state.get("candidate_tools") or []),
+            # This sentence is the point of the decision log, but filesystem
+            # paths and addresses are never part of that point.
+            "reason": _scrub_reason(state.get("selection_rationale")),
```
amd-vivekag: **[Medium]** Summary mode can persist user-derived secrets through the selector rationale.

`why` is free-form model output produced after seeing the full recent user turns, but this stores it verbatim after scrubbing only paths and IPs; a quoted kernel line, customer identifier, or credential therefore bypasses the summary-mode content boundary.

Fix: store structured selector fields or a digest in summary mode and retain verbatim rationale only in `full`; add a test whose rationale repeats `secret_kernel` and a credential-like token.

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The two findings from the last review are fixed correctly: cancellation state now reaches the Chainlit step, and the critic receives exact, re-verified source windows from deep in a paste. Every earlier High/Medium and all five Copilot findings remain fixed.

**Blocking findings:** 0 High · 1 Medium

The blocker is the new summary decision-log privacy boundary: the selector's free-form rationale is stored verbatim after only path/IP scrubbing, even though that model saw the user's full recent turns and can echo source, customer identifiers, or credentials. Fix this in summary mode, or remove the independently deliverable decision-log feature from #425 and track it separately.

**Safely separable follow-ups**

1. **Issue: Bound cancellation latency for non-cooperative chat tools**
   - Scope: define behavior for shell/plugin tools that ignore the cancellation token; keep cooperative CIA cancellation unchanged.
   - Acceptance: a cancelled turn returns within a configured grace period, reports `still running` honestly, and never emits a false successful completion.
2. **Issue: Rotate and offload opt-in chat decision logs**
   - Scope: move JSONL serialization/writes off the request loop and add configurable per-session size/age retention.
   - Acceptance: a multi-megabyte full-mode turn does not stall an event-loop heartbeat; disk growth is bounded; rotated files remain valid JSONL with 0700/0600 permissions.

These are follow-ups, not merge blockers for the progress/cancellation feature.

**Architecture:** the leaf's streaming, cancellation, staging, prompt-evidence, and CLI/UI convergence contracts are now coherent. #423 and #424 are both clean with green checks; #425 is independently mergeable, subject to the decision-log privacy fix.

**CI:** all 23 current checks pass, including CPU/CIA matrices, chat 3.11–3.14, LangGraph floor, MI350, and workload regression.

docs/chat/configuration.md
```diff
@@ -71,6 +71,35 @@ If you would rather not store the key at all, leave it out of the file and
...
+`$XDG_STATE_HOME/aorta/chat/sessions/` (default
+`~/.local/state/aorta/chat/sessions/`). The directory is mode `0700` and each
+session JSONL file is mode `0600`.
+
+Both the browser and CLI record from the state returned by `invoke_agent`, so
+they capture the same decisions without changing the CLI to a streaming path:
+
+- route and selector-ranked tools;
+- the selector's own reason, with filesystem paths and IP addresses always
+  scrubbed regardless of `--no-redact`;
```
amd-vivekag: **[Medium]** The documented summary-log contract still describes the pre-fix behavior.

Summary mode now stores only the rationale's shape/digest; path/IP scrubbing and verbatim rationale apply to `full` mode, so operators following this section cannot know what their logs contain.

Fix: state that summary mode digests selector rationale alongside other model content, and that full mode retains it verbatim (including paths/IPs).

amd-vivekag (commented):
## Re-review summary

**Verdict: fix-before-ship**

The last privacy blocker is fixed correctly in code and tests: summary mode now digests selector rationale, while full mode retains it explicitly. Every earlier human High/Medium and all five Copilot findings remain fixed.

**Blocking findings:** 0 High · 1 Medium

The public decision-log contract was not updated with that fix and still promises a scrubbed rationale in summary mode. This is the operator-facing acceptance contract for the new logging feature, so it should match before merge.

**Safely separable follow-ups**

1. **Issue: Bound cancellation latency for non-cooperative chat tools**
   - Scope: shell/plugin tools that ignore the cancellation token; keep cooperative CIA cancellation unchanged.
   - Acceptance: cancelled turns return within a configured grace period, report `still running` honestly, and never emit false success.
2. **Issue: Rotate and offload opt-in chat decision logs**
   - Scope: off-loop JSONL persistence plus configurable size/age retention.
   - Acceptance: large full-mode turns do not stall an event-loop heartbeat; disk growth is bounded; rotated files remain valid private JSONL.

These are non-blocking follow-ups and should not expand this PR.

**Architecture:** streaming, cancellation, staging, prompt evidence, decision-log privacy, and CLI/UI convergence are coherent. #423 and #424 are clean and green; #425 is independently mergeable after the documentation contract is corrected.

**CI:** all 23 current checks pass, including CPU/CIA matrices, chat 3.11–3.14, LangGraph floor, MI350, and workload regression.

amd-vivekag (approved):
No further blocking comments — approving; the two previously specified cancellation-latency and decision-log retention follow-ups remain appropriately out of scope.

PR #452, opened 2026-09-03
ci: stage TokenSpeed serving for nightly perf gating, with an opt-in docker socket

vivekkhandelwal1:
Adds a `tokenspeed_serve_smoke` entry to the nightly evaluation matrix, plus the
Docker client and opt-in socket plumbing needed to support it, and the rollout
plan for turning perf gating on for it.

## The socket sign-off has been given, for the nightly lane only

**Updated 2026-09-03.** An earlier revision of this description said the socket
decision was still open and that no lane enabled it. That was stale, and both
reviewers correctly flagged the contradiction against `nightly-eval.yml`'s
`docker_socket: true`. The decision has been given: the nightly lane gets the
host daemon socket, and nothing else does. The code was right; this description
and the rollout doc were what needed fixing.

The reasoning is recorded in `docs/tokenspeed-gating-rollout.md` under
*Security: the marginal risk, and what actually bounds it*, rather than only in
a review thread — a future reader deciding whether a **second** lane may have
the socket needs the argument, not the verdict. In short:

- **The marginal risk is close to zero, because the privilege is already there.**
  `eval-reusable.yml` and `refresh-baselines.yml` both open with
  `docker run -v "$ws":/ws …` and `sudo -n` as **plain host steps**, before any
  container exists (`eval-reusable.yml:58-69`, ahead of the container setup at
  line 86). The runner account therefore has direct daemon access *and*
  passwordless sudo today. Any job step that can execute on this runner is
  already host-root capable, with or without the socket. The mount relocates an
  existing capability into `aorta-ci-gpu`; it does not create one.
- **Reachability is bounded by write access, and every part of that is greppable.**
  There is no `pull_request_target` anywhere in `.github/`. Both PR-reachable
  routes to the runner carry a fork guard on
  `head.repo.full_name == github.repository` (`gpu-tests.yml`, twice, and
  `bump-validate.yml`). `gpu-tests.yml` does not go through `eval-reusable.yml`
  and passes no socket argument at all. `bump-validate.yml` pins
  `docker_socket: false` explicitly rather than inheriting a default a later
  edit could flip.

So the population that can reach the daemon through this mount is exactly the
population that can push a branch to this repository — which is the population
that can already run `sudo` on the runner through any workflow step.

**The "single-tenant runner" framing is withdrawn.** It was doing work the
argument does not need and cannot currently support: `aorta` and
`aorta-internal` hold *separate repo-level runner registrations*, and the
internal one's July jobs reported machine `sharkmi300x-3` rather than the MI350
host, so shared tenancy is unconfirmed. It does not need to be confirmed —
tenancy bears on *measurement* (a neighbour's load moving the numbers), which
the ten-night window and the two-cell control handle, not on this grant.

## The socket is opt-in per lane

`rocm-ci-setup` already had a `docker-socket` input defaulting to `false`, but no
caller could reach it. This adds a matching `docker_socket` `workflow_call` input
on `eval-reusable.yml`, also defaulting to `false`, forwarded to the setup step.

It is deliberately **not** set inside `eval-reusable.yml`. That workflow is
shared by the nightly and by `bump-validate.yml`, so a value chosen centrally
would hand the socket to a PR-triggered lane as a side effect of enabling the
nightly.

| Lane | Trigger | `docker_socket` |
|---|---|---|
| `nightly-eval.yml` | `workflow_run` (Nightly wheels) + dispatch | `true` |
| `bump-validate.yml` | `pull_request` | `false`, set explicitly |
| `refresh-baselines.yml` | dispatch only | n/a — calls `rocm-ci-setup` directly, passes no socket argument |
| `sanitizers-nightly.yml` | schedule + dispatch | n/a — does not use `eval-reusable.yml`; calls `rocm-ci-setup` directly and passes no socket argument |

The grant is per lane and should stay that way. Any *other* lane wanting the
socket is a fresh decision, and the argument above does not transfer to a lane
whose trigger is reachable from a fork or an unreviewed branch.

## Review round: what changed

Eleven threads from `Copilot` and `amd-vivekag`. All addressed; the substantive
fixes are below.

**A scoped baseline refresh no longer disarms the entries it was protecting.**
`--perf-gate-entry` exists so that blessing one workload leaves the other
sixteen cells alone, but the baseline file is rewritten *whole* and an
out-of-scope cell was rebuilt as `{"passed": true}` plus correctness metrics —
so a scoped bless of workload B silently ungated workload A, the exact inverse
of the flag's purpose. Existing `step_time_ms` and non-equal-policy metric specs
are now carried over verbatim (deep-copied). The carry-over is keyed on "this
run is not deriving bounds for this cell" rather than on "out of scope", because
the same hole was reachable through the default correctness-only mode — which is
this workflow's routine dispatch. Omitting `--perf-gate` means "do not arm new
gates", never "disarm the ones already blessed".

**`needs_docker_daemon` is honoured on the refresh path, not just the nightly.**
`build_baselines` still called `run_entry` for every GPU-eligible entry, and the
refresh lane mounts no socket — so the serving row produced no `matrix.json`,
routed into `incomplete`, and aborted **every** baseline refresh as partial, for
every unrelated workload. It now gets the same contract `min_gpus` gives the
8-GPU rows: skip, carry over existing baselines, record the reason. Naming such
an entry in `--perf-gate-entry` is instead a hard error — carrying it silently
would emit a correctness-only file that reads in the PR diff exactly like a
successful bless of a cell that never ran.

**A `perf_gate_entry` that parses to no names is rejected.** `,` or whitespace
passed the raw non-empty check, contributed no flags, and left a bare
`--perf-gate` — a *global* perf refresh arming ceilings on all sixteen cells,
reached by an operator who thought they had scoped it, and without even the
unscoped warning. Parsed before validation now, with a second refusal inside the
container at the point the degradation would happen.

**The socket guard no longer has a hole in it.** It compared against the literal
string `true`, so a PR-triggered caller written `${{ true }}` would get the
socket and be absent from `enabled` — a test passing over precisely the case it
exists to prevent. Replaced with a fail-closed classifier: only an omitted input
or an explicit literal `false` counts as off. Mutation-tested the way the
existing six guards were.

**The serving pre-warm step actually pre-warms.** `docker pull` warms the
engine-image cache, and the exported `HF_HOME` was not the path the workload
reads — `tokenspeed_serve.py:730-731` derives it from `workload_config.hf_home`
or `<work_dir>/u<uid>/hf`. Split into an image pull and a real
`hf download Qwen/Qwen3-0.6B` under the resolved path, as the running uid.

**A re-bless no longer deletes bounds `--perf-gate` cannot re-derive.** Found in
the second review round, in the carry-over fix itself: keying the carry on "not
deriving" was right for auto-gateable bounds but wrong for ones the refresher is
structurally incapable of producing. `median_itl_ms` is allowlisted yet
`_NO_AUTO_GATE` — measured at ~0, where a relative margin is not a bound — so a
hand-written ceiling on it was silently dropped by a scoped re-bless of its own
entry. The same hole existed for `step_time_ms` when a cell reported no
`mean_step_time_ms`. Both carried forward now, mutation-tested. The invariant,
now written down: a refresh may decline to arm a gate; it never disarms one.

**The work-root command repairs rather than merely creates.** `mkdir -p` as root
over a leftover user-owned `/tmp/ts-work-serve` returns 0 and changes neither
owner nor mode, so the next run failed the same ownership check with the remedy
apparently already applied. Now `mkdir -p`, `chown 0:0`, `chmod 1777` —
idempotent either way — plus a `stat` read-back.

## Perf-gate scoping

`refresh_baselines.py` has a repeatable `--perf-gate-entry NAME`, validated
against the matrix, but `refresh-baselines.yml` is dispatch-only and had no
matching input — so the scoped bless was not runnable from the Actions UI and
the only reachable perf refresh was the unscoped one, which rewrites the whole
baseline file and arms step-time ceilings across all sixteen cells (including
`gpu_smoke`, `inference_offline`, `training_ddp`, `training_fsdp`, `race` and
`llm_determinism` — other people's workloads, each from a single observation
with no variance data).

`perf_gate_entry` defaults to empty, so a dispatch that omits it behaves exactly
as before. A comma- or space-separated list maps to one flag per name. A scope
without `perf_gate` is a fast error; a scope that names nothing is now also an
error; an unscoped perf refresh warns but is not blocked, because it is a
legitimate end state once every workload has variance data.

Dispatch inputs reach the container through `env:`/`-e` rather than being
interpolated into the shell body — `perf_gate_entry` is free text and `${{ }}`
substitution happens before bash sees the script. This also removes three
pre-existing interpolations.

## `warmup_steps: 1` → `2`

A step-0 Triton compile excursion reaches the metrics at `warmup_steps: 1`,
measured at 1 cell-run in 13: a ~10× first-token spike (465 ms against a
43–47 ms clean range) and a 2825 ms step against a 1108–1178 ms range. It lands
on the first *measured* step, after the single warmup step has already been
discarded, and breaches three of the four originally-proposed gates on a run
that is not a regression. `num_warmups` cannot substitute — it warms requests
within one bench invocation, not the compile cache across invocations.

Consequence, handled in the docs: the variance data in
`docs/tokenspeed-gating-rollout.md` was all taken at `warmup_steps: 1`, so it is
no longer a predictor of what the record-only window will show. The table is
kept — there is no measurement at the new setting yet — but explicitly reframed
as the rationale for the change rather than the baseline to bless against, and
the ten-night window is pinned to `warmup_steps: 2`.

## Rollout

See `docs/tokenspeed-gating-rollout.md` for the record-only window and the
two-metric first bless: **`median_tpot_ms`** and **`p99_itl_ms`**. Those two are
the metrics whose definitions exclude the excursion (1.01× and 1.00× on the
excursion run), which is why they are the ones that survive. The three
duration-derived metrics stay record-only until the window shows the excursion
gone.

Two operational preconditions before the entry first runs, both on the runner:
`/tmp/ts-work-serve` must exist root-owned and `1777` (sequence in the doc), and
both pinned digests — the base image in `docker/Dockerfile.ci-gpu` and the
engine image in the recipe — must hold for the whole window, since a mid-window
bump invalidates it rather than being absorbed by it.

## Tests

`tests/ci/`: **221 passed** (189 on `main`, 32 added across the original change
and the two review rounds). `tests/docker/`: 56 passed. The workflow-shape tests are
verified by mutation — enabling the socket on `bump-validate.yml` via eight
different truthy expressions, hardcoding it into the shared reusable workflow
three ways, deleting bump-validate's explicit `false`, flipping the reusable
default, reverting `warmup_steps`, dropping the `perf_gate_entry` input, and
reintroducing shell interpolation each fail.

### DIFF
--- a/.github/workflows/nightly-eval.yml
+++ b/.github/workflows/nightly-eval.yml
@@ -35,6 +35,15 @@ jobs:
       ref: ${{ github.event.workflow_run.head_sha || github.sha }}
       wheel_run_id: ${{ github.event.workflow_run.id || '' }}
       do_alert: true
+      # The nightly is the ONLY lane that gets the host docker socket: the
+      # tokenspeed_serve_smoke entry launches the TokenSpeed container itself, so
+      # it needs a route to the daemon. This lane is not PR-triggered -- it runs
+      # on a published nightly wheel or an explicit dispatch -- which is what
+      # makes the grant acceptable. Do not copy this line into bump-validate.yml.
+      # Until the socket is signed off, leaving this false is also safe: the
+      # entry declares `needs_docker_daemon: true` and reports `skip` with the
+      # reason recorded rather than failing the nightly.
+      docker_socket: true
     secrets: inherit
 
   # Append today's results to the ci-results data branch. The dashboard itself is
--- a/.github/workflows/nightly-eval.yml
+++ b/.github/workflows/nightly-eval.yml
@@ -37,12 +37,26 @@ jobs:
       do_alert: true
       # The nightly is the ONLY lane that gets the host docker socket: the
       # tokenspeed_serve_smoke entry launches the TokenSpeed container itself, so
-      # it needs a route to the daemon. This lane is not PR-triggered -- it runs
-      # on a published nightly wheel or an explicit dispatch -- which is what
-      # makes the grant acceptable. Do not copy this line into bump-validate.yml.
-      # Until the socket is signed off, leaving this false is also safe: the
-      # entry declares `needs_docker_daemon: true` and reports `skip` with the
-      # reason recorded rather than failing the nightly.
+      # it needs a route to the daemon. Signed off 2026-09-03, for this lane
+      # only. Do not copy this line into bump-validate.yml or any other lane.
+      #
+      # What the sign-off rests on, in one line each -- the full argument is in
+      # docs/tokenspeed-gating-rollout.md under Security:
+      #   * The marginal risk is close to zero. eval-reusable.yml's first step
+      #     already runs `docker run -v "$ws":/ws` and `sudo -n` as PLAIN HOST
+      #     steps, before any container exists, so the runner account has daemon
+      #     access and passwordless sudo today. The socket moves that inside
+      #     aorta-ci-gpu; it does not create it.
+      #   * Reachability is bounded by write access, and that is greppable: no
+      #     `pull_request_target` anywhere in .github/, and both PR-reachable
+      #     routes to this runner (gpu-tests.yml, bump-validate.yml) gate on
+      #     head.repo.full_name == github.repository, so fork code never lands
+      #     here at all.
+      #   * Neither PR lane can obtain it regardless: gpu-tests.yml does not use
+      #     eval-reusable.yml and passes no socket argument, and
+      #     bump-validate.yml pins `docker_socket: false` explicitly rather than
+      #     inheriting a default a later edit could flip.
+      # Not part of the argument: runner tenancy. See the doc.
       docker_socket: true
     secrets: inherit
 
--- a/.github/workflows/refresh-baselines.yml
+++ b/.github/workflows/refresh-baselines.yml
@@ -21,6 +21,20 @@ on:
         type: boolean
         required: false
         default: false
+      # The scoping guard for a perf refresh. An unscoped `--perf-gate` rewrites
+      # the whole baseline file and arms step-time ceilings on every matrix
+      # entry -- gpu_smoke, inference_offline, training_ddp, training_fsdp, race
+      # and llm_determinism included -- each from a single observation with no
+      # variance data behind it. Naming the entry under test keeps the refresh
+      # correctness-only everywhere else, so perf gating can be rolled out one
+      # workload at a time. Unknown names are rejected by refresh_baselines.py
+      # rather than silently scoping the gate to nothing.
+      perf_gate_entry:
+        description: >-
+          Restrict perf_gate to these matrix entries (comma- or space-separated,
+          e.g. "tokenspeed_serve_smoke"). Empty gates EVERY entry.
+        required: false
+        default: ""
 
 permissions:
   contents: write
@@ -58,14 +72,51 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v5
 
+      # refresh_baselines.py rejects both of these too, but only after pip has
+      # installed the wheel inside the container -- minutes of GPU runner time to
+      # learn about a typo in a dispatch form. Same rules, checked in seconds.
+      - name: Validate perf-gate inputs
+        env:
+          PERF_GATE: ${{ inputs.perf_gate }}
+          PERF_GATE_ENTRY: ${{ inputs.perf_gate_entry }}
+        run: |
+          set -euo pipefail
+          if [ -n "${PERF_GATE_ENTRY}" ] && [ "${PERF_GATE}" != "true" ]; then
+            echo "::error::perf_gate_entry has no effect without perf_gate: true"
+            exit 1
+          fi
+          # Deliberately a warning, not an error: an unscoped perf refresh is a
+          # legitimate end-state operation (bless everything once every workload
+          # has variance data), and it is the behaviour every existing dispatch
+          # of this workflow has had. Loud, not blocked.
+          if [ "${PERF_GATE}" = "true" ] && [ -z "${PERF_GATE_ENTRY}" ]; then
+            echo "::warning::Unscoped perf refresh: this arms step-time ceilings on EVERY matrix entry from a single observation each. Set perf_gate_entry to scope it to the workload under test."
+          fi
+
       - name: Set up ROCm CI container
         uses: ./.github/actions/rocm-ci-setup
         with:
           rocm-shared-key: ${{ secrets.ROCM_SHARED_KEY }}
 
+      # Dispatch inputs reach the container as environment variables rather than
+      # being interpolated into this shell string. `perf_gate_entry` is free
+      # text, and `${{ }}` substitution happens before bash sees the script, so
+      # interpolating it would let a dispatch value close the quote and run
+      # arbitrary commands on the runner.
       - name: Install nightly wheel and regenerate baselines
+        env:
+          STEP_TIME_MARGIN: ${{ inputs.step_time_margin }}
+          THROUGHPUT_MARGIN: ${{ inputs.throughput_margin }}
+          PERF_GATE: ${{ inputs.perf_gate }}
+          PERF_GATE_ENTRY: ${{ inputs.perf_gate_entry }}
         run: |
-          bash scripts/ci/docker_cmd.sh exec "${{ env.CONTAINER_NAME }}" bash -lc '
+          bash scripts/ci/docker_cmd.sh exec \
+            -e STEP_TIME_MARGIN="${STEP_TIME_MARGIN}" \
+            -e THROUGHPUT_MARGIN="${THROUGHPUT_MARGIN}" \
+            -e PERF_GATE="${PERF_GATE}" \
+            -e PERF_GATE_ENTRY="${PERF_GATE_ENTRY}" \
+            -e DEV_WHEELS_INDEX="${{ env.DEV_WHEELS_INDEX }}" \
+            "${{ env.CONTAINER_NAME }}" bash -lc '
             set -euo pipefail
             cd /workspace/aorta
             python -m pip install --upgrade pip
@@ -73,11 +124,29 @@ jobs:
             # so baselines are generated on the exact dependency set they gate.
             constraint=""
             [ -f config/ci/ci-constraints.txt ] && constraint="-c config/ci/ci-constraints.txt"
-            pip install --pre "amd-aorta[hw-queue]" -f "${{ env.DEV_WHEELS_INDEX }}" $constraint
+            pip install --pre "amd-aorta[hw-queue]" -f "$DEV_WHEELS_INDEX" $constraint
+            # NOTE: this whole script is inside a single-quoted bash -lc, so it
+            # must contain no apostrophes.
+            args=()
+            if [ "${PERF_GATE}" = "true" ]; then
+              args+=(--perf-gate)
+            fi
+            # Comma- or space-separated list -> one --perf-gate-entry per name,
+            # which is what the repeatable flag expects. Empty fields (from a
+            # trailing comma) are dropped rather than passed as an empty string,
+            # which the known-entry check in refresh_baselines.py would reject.
+            if [ -n "${PERF_GATE_ENTRY}" ]; then
+              IFS=", " read -r -a _entries <<< "${PERF_GATE_ENTRY}"
+              for _e in "${_entries[@]}"; do
+                if [ -n "${_e}" ]; then
+                  args+=(--perf-gate-entry "${_e}")
+                fi
+              done
+            fi
             python scripts/ci/refresh_baselines.py \
-              --step-time-margin "${{ github.event.inputs.step_time_margin || 0.25 }}" \
-              --throughput-margin "${{ github.event.inputs.throughput_margin || 0.15 }}" \
-              ${{ github.event.inputs.perf_gate == 'true' && '--perf-gate' || '' }}
+              --step-time-margin "${STEP_TIME_MARGIN:-0.25}" \
+              --throughput-margin "${THROUGHPUT_MARGIN:-0.15}" \
+              ${args[@]+"${args[@]}"}
           '
 
       # The container wrote regression_baselines.yaml into the mounted workspace;
--- a/.github/workflows/refresh-baselines.yml
+++ b/.github/workflows/refresh-baselines.yml
@@ -81,7 +81,25 @@ jobs:
           PERF_GATE_ENTRY: ${{ inputs.perf_gate_entry }}
         run: |
           set -euo pipefail
-          if [ -n "${PERF_GATE_ENTRY}" ] && [ "${PERF_GATE}" != "true" ]; then
+          # Noglob before the word-splitting below: an unquoted expansion is
+          # also a glob, and a scope of `*` would otherwise become a list of
+          # filenames from the workspace root.
+          set -f
+          # Parse the scope FIRST, then validate what parsing produced. A raw
+          # non-empty test is not the same question: `,`, `, ,` or a lone tab
+          # are all non-empty and name no entries, so they would pass the check
+          # below, contribute no --perf-gate-entry flags, and leave a bare
+          # --perf-gate -- degrading a scoped bless into the unscoped global
+          # refresh this input exists to prevent, and without even the unscoped
+          # warning, since the raw value is not empty. Same delimiters as the
+          # container-side split, which is what makes this a preview of it.
+          names=( ${PERF_GATE_ENTRY//,/ } )
+
+          if [ -n "${PERF_GATE_ENTRY}" ] && [ "${#names[@]}" -eq 0 ]; then
+            echo "::error::perf_gate_entry (${PERF_GATE_ENTRY}) is delimiters only and names no matrix entry. Leave it empty for a deliberate unscoped refresh, or name the workload under test."
+            exit 1
+          fi
+          if [ "${#names[@]}" -gt 0 ] && [ "${PERF_GATE}" != "true" ]; then
             echo "::error::perf_gate_entry has no effect without perf_gate: true"
             exit 1
           fi
@@ -89,8 +107,13 @@ jobs:
           # legitimate end-state operation (bless everything once every workload
           # has variance data), and it is the behaviour every existing dispatch
           # of this workflow has had. Loud, not blocked.
-          if [ "${PERF_GATE}" = "true" ] && [ -z "${PERF_GATE_ENTRY}" ]; then
+          if [ "${PERF_GATE}" = "true" ] && [ "${#names[@]}" -eq 0 ]; then
             echo "::warning::Unscoped perf refresh: this arms step-time ceilings on EVERY matrix entry from a single observation each. Set perf_gate_entry to scope it to the workload under test."
+          fi
+          if [ "${#names[@]}" -gt 0 ]; then
+            echo "perf-gate scope: ${names[*]}"
+          else
+            echo "perf-gate scope: none (unscoped)"
           fi
 
       - name: Set up ROCm CI container
@@ -136,12 +159,23 @@ jobs:
             # trailing comma) are dropped rather than passed as an empty string,
             # which the known-entry check in refresh_baselines.py would reject.
             if [ -n "${PERF_GATE_ENTRY}" ]; then
+              _named=0
               IFS=", " read -r -a _entries <<< "${PERF_GATE_ENTRY}"
               for _e in "${_entries[@]}"; do
                 if [ -n "${_e}" ]; then
                   args+=(--perf-gate-entry "${_e}")
+                  _named=$((_named + 1))
                 fi
               done
+              # Second layer under the validation step, at the point where the
+              # degradation would actually happen. A non-empty scope that
+              # yields no names must not fall through to a bare --perf-gate:
+              # that is a GLOBAL perf refresh arming ceilings on every entry,
+              # which is the opposite of what the operator asked for.
+              if [ "${_named}" -eq 0 ]; then
+                echo "perf_gate_entry was set but named no entries; refusing to fall back to a global perf refresh" >&2
+                exit 1
+              fi
             fi
             python scripts/ci/refresh_baselines.py \
               --step-time-margin "${STEP_TIME_MARGIN:-0.25}" \
--- a/config/ci/nightly_eval_matrix.yaml
+++ b/config/ci/nightly_eval_matrix.yaml
@@ -12,6 +12,12 @@
 #               defaults to nproc when nproc is set)
 #   timeout_sec - optional; per-entry wall-clock budget (default 1800). Exceeding
 #               it kills the whole process group and FAILS the entry.
+#   needs_docker_daemon - optional; skip when no Docker daemon is reachable from
+#               wherever nightly_eval.py is running. For workloads that start
+#               their engine in a sibling container. Like min_gpus this states a
+#               runner CAPABILITY, so its absence is a skip, not a failure --
+#               and only entries that set it consult the probe, so a broken
+#               daemon cannot turn the rest of the nightly into skips.
 #
 # Record-only is NOT a field here: a cell is record-only exactly while
 # config/ci/regression_baselines.yaml has no key for it. Adding an entry to
@@ -88,17 +94,6 @@ entries:
     nproc: 8
     min_gpus: 8
 
-# Entries that are ready to run but blocked on a runner prerequisite. This key
-# is not read by anything: nightly_eval.py and refresh_baselines.py both iterate
-# `entries` only, so nothing below executes. It exists because the alternative
-# for a staged-but-blocked entry is a commented-out block (the convention the
-# two NOTEs above use), and a comment cannot be checked -- tests/ci/
-# test_nightly_eval.py validates these the same way it validates `entries`, so
-# the recipe path, the GPU requirement and the metric names are known-good on
-# the day the prerequisite lands rather than discovered by a red nightly.
-#
-# Promote by moving the entry (minus `blocked_on`) into `entries` above.
-pending_entries:
   # TokenSpeed online serving. Chosen over the other four serving recipes as the
   # cheapest cell that still answers something: Qwen3-0.6B is ~1.2 GB of weights
   # against ~40 GB for the gpt-oss recipes, and two cells against four (models)
@@ -109,40 +104,50 @@ pending_entries:
   # differ in speed.
   #
   # 3600s, not the 1800s default: two bring-ups at the observed startup spread
-  # (189-379 s, see docs/tokenspeed.md) plus the teardown VRAM drain is ~15 min
-  # before the container image pull and any cold-cache weight download, and a
-  # timeout is an unconditional entry FAILURE rather than a slow record.
+  # (180-415 s measured over 13 cell-runs) plus the teardown VRAM drain is ~15
+  # min before the image pull and any cold-cache weight download, and a timeout
+  # is an unconditional entry FAILURE rather than a slow record.
+  #
+  # `needs_docker_daemon` is what makes this row safe to have here today. The
+  # nightly runs nightly_eval.py INSIDE aorta-ci-gpu, and `tokenspeed_serve`
+  # starts the TokenSpeed engine in its own container, so it needs a docker
+  # client in the image (present, and proven: the built image carries 29.7.2 and
+  # no daemon) plus a route to a daemon (a per-lane opt-in that no lane sets,
+  # because it grants effective root on the runner and that is the CI owner's
+  # call). Until a lane sets `docker-socket: true` this entry SKIPS, with the
+  # reason in the results, rather than failing on "cannot connect to the Docker
+  # daemon" -- the same contract min_gpus gives the 8-GPU rows.
+  #
+  # Proven on gfx950 on 2026-09-02: two sweeps driven from inside the CI
+  # container over a mounted socket, four cell-runs, all passed, twelve clean
+  # steps (1108-1149 ms), metrics within 5% of the host-side envelope.
+  #
+  # The recipe's `work_dir: /tmp/ts-work-serve` must be mounted at the SAME
+  # string on both sides and be root-owned -- see the work_dir section of
+  # docs/tokenspeed-gating-rollout.md, which the demonstration made a good deal
+  # more specific.
   - name: tokenspeed_serve_smoke
     recipe: recipes/tokenspeed/tokenspeed-serve-bench-smoke.yaml
     min_gpus: 1
     timeout_sec: 3600
-    # The nightly runs `nightly_eval.py` INSIDE the aorta-ci-gpu container
-    # (eval-reusable.yml -> docker_cmd.sh exec), and `tokenspeed_serve` runs the
-    # TokenSpeed engine in its own container, so the entry needs a docker client
-    # and a route to a daemon in there. Both now EXIST but neither is switched
-    # on, and nothing has been observed working:
-    #
-    #   * Dockerfile.ci-gpu installs the client (docker/install_docker_cli.py).
-    #     The installer was run for real and the binary it produces drives a
-    #     live daemon, but the CI IMAGE has never been built -- the ROCm 10 base
-    #     is ~52 GB and no machine with that much free disk has run the build.
-    #   * The socket is a per-lane opt-in (`docker-socket: true` on the
-    #     rocm-ci-setup action, mounting docker-compose.docker-socket.yaml). No
-    #     lane sets it, deliberately: it grants effective root on the runner and
-    #     that is the CI owner's call, not this branch's.
-    #   * work_dir must then be set explicitly to the path that override mounts
-    #     at the SAME string on both sides. The default is not sufficient by
-    #     being equal to it -- see the work_dir section in the rollout doc.
-    #
-    # So the failure mode has moved from "'docker' not on PATH" to "cannot
-    # connect to the Docker daemon", which reddens the nightly exactly as hard.
-    # Promotion needs an observed container launch, not these three edits.
-    # See docs/tokenspeed-gating-rollout.md for the remaining steps.
-    blocked_on: >-
-      docker client and socket override exist but are unproven and unswitched:
-      the CI image has not been built, no lane sets docker-socket: true (a
-      root-equivalent grant needing CI-owner sign-off), and no container launch
-      from inside aorta-ci-gpu has been demonstrated
+    needs_docker_daemon: true
+
+# Entries that are ready to run but blocked on a runner prerequisite. This key
+# is not read by anything: nightly_eval.py and refresh_baselines.py both iterate
+# `entries` only, so nothing below executes. It exists because the alternative
+# for a staged-but-blocked entry is a commented-out block (the convention the
+# two NOTEs above use), and a comment cannot be checked -- tests/ci/
+# test_nightly_eval.py validates these the same way it validates `entries`, so
+# the recipe path, the GPU requirement and the metric names are known-good on
+# the day the prerequisite lands rather than discovered by a red nightly.
+#
+# Promote by moving the entry (minus `blocked_on`) into `entries` above.
+# Empty as of 2026-09-02: tokenspeed_serve_smoke, the only entry that was ever
+# staged here, was promoted into `entries` once a cell was demonstrated running
+# from inside the CI container. Keep the key and this comment -- the convention
+# is worth more than the one row was, and the next blocked entry should land
+# here rather than in a comment block.
+pending_entries: []
 
 # dtype axis (fp32/fp16/bf16) is a follow-up: the smoke recipes pin one dtype
 # each, so per-dtype coverage needs dtype recipe variants (or workload_config
--- a/config/ci/nightly_eval_matrix.yaml
+++ b/config/ci/nightly_eval_matrix.yaml
@@ -10,6 +10,15 @@
 #   nproc     - optional; launch under `torchrun --standalone --nproc_per_node=<n>`
 #   min_gpus  - optional; skip when torch.cuda.device_count() is lower (default 1;
 #               defaults to nproc when nproc is set)
+#   timeout_sec - optional; per-entry wall-clock budget (default 1800). Exceeding
+#               it kills the whole process group and FAILS the entry.
+#
+# Record-only is NOT a field here: a cell is record-only exactly while
+# config/ci/regression_baselines.yaml has no key for it. Adding an entry to
+# `entries` therefore starts it record-only by construction. Note the limit of
+# that -- record-only defers PERFORMANCE gating only. A cell that errors or
+# fails is a `fail` with or without a baseline (nightly_eval.py is fail-closed),
+# so an entry may only be added here once it can actually pass on the runner.
 #
 # Phase 1 wires one config per registered workload (the smoke recipes). Per-workload
 # matrices (dtypes, larger GPU counts) are Phase 3 -- add rows here, no code change.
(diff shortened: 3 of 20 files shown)


.github/workflows/nightly-eval.yml
```diff
@@ -35,6 +35,15 @@ jobs:
...
       do_alert: true
+      # The nightly is the ONLY lane that gets the host docker socket: the
+      # tokenspeed_serve_smoke entry launches the TokenSpeed container itself, so
+      # it needs a route to the daemon. This lane is not PR-triggered -- it runs
+      # on a published nightly wheel or an explicit dispatch -- which is what
+      # makes the grant acceptable. Do not copy this line into bump-validate.yml.
+      # Until the socket is signed off, leaving this false is also safe: the
+      # entry declares `needs_docker_daemon: true` and reports `skip` with the
+      # reason recorded rather than failing the nightly.
+      docker_socket: true
```
amd-vivekag, line 60: **[High]** This enables host-root access before the required sign-off.

The rollout still says no lane enables the socket and the PR says single-tenant confirmation is pending.

Fix: Keep `docker_socket` false or omitted until named approval, then enable it in a separate reviewed change.

config/ci/nightly_eval_matrix.yaml
```diff
@@ -79,6 +94,61 @@ entries:
...
+  #
+  # The recipe's `work_dir: /tmp/ts-work-serve` must be mounted at the SAME
+  # string on both sides and be root-owned -- see the work_dir section of
+  # docs/tokenspeed-gating-rollout.md, which the demonstration made a good deal
+  # more specific.
+  - name: tokenspeed_serve_smoke
+    recipe: recipes/tokenspeed/tokenspeed-serve-bench-smoke.yaml
+    min_gpus: 1
+    timeout_sec: 3600
+    needs_docker_daemon: true
```
amd-vivekag, line 137: **[High]** This live entry makes every baseline refresh fail.

`build_baselines()` runs every GPU-eligible entry, but the refresh workflow neither mounts the socket nor handles `needs_docker_daemon`, so this produces no matrix and triggers the atomic abort.

Fix: Preserve unavailable daemon-dependent entries during ordinary refreshes, and provide an approved socket-enabled path when this entry is explicitly selected.

scripts/ci/refresh_baselines.py
```diff
@@ -103,12 +104,20 @@ def build_baselines(
...
-            if perf_gate:
+            #
+            # --perf-gate is also scopable to named entries, because this file is
+            # rewritten WHOLE on every refresh: without a scope, arming one
+            # workload's perf gates arms every other entry's at the same time,
+            # off whatever single run happened to be under way. Those entries
+            # have no variance evidence behind them, and a bound derived from one
+            # observation is the flaky-gate failure this whole exercise exists to
+            # avoid -- so rolling gating out per workload has to be expressible.
+            if perf_gate and (not perf_gate_entries or name in perf_gate_entries):
```
amd-vivekag: **[High]** A scoped refresh silently removes existing performance gates.

Out-of-scope specs are rebuilt without their old `step_time_ms` or performance metrics, then the whole baseline file is replaced; blessing workload B therefore ungates workload A.

Fix: Merge refreshed correctness fields into each out-of-scope existing spec while preserving its performance bounds.

tests/ci/test_dashboard_and_alert.py
```diff
@@ -1851,3 +1852,418 @@ def test_hidden_path_uploads_declare_include_hidden_files():
...
+                if match:
+                    forwarding[f"./.github/workflows/{path.name}"] = match.group(1)
+
+    enabled = []
+    for path in workflows:
+        for name, job in (_load_workflow(path.name).get("jobs") or {}).items():
+            for step in job.get("steps") or []:
+                if "rocm-ci-setup" not in (step.get("uses") or ""):
+                    continue
+                if str((step.get("with") or {}).get("docker-socket", "")).lower() == "true":
```
amd-vivekag: **[Medium]** This security guard recognizes only the literal string `true`.

A PR caller using `${{ true }}` or another truthy expression gets the socket at runtime but is omitted from `enabled`, so this test still passes.

Fix: For PR-triggered routes, reject every socket value except explicit literal `false`, and add an expression-valued mutation case.

.github/workflows/refresh-baselines.yml
```diff
@@ -58,26 +72,81 @@ jobs:
...
+            # must contain no apostrophes.
+            args=()
+            if [ "${PERF_GATE}" = "true" ]; then
+              args+=(--perf-gate)
+            fi
+            # Comma- or space-separated list -> one --perf-gate-entry per name,
+            # which is what the repeatable flag expects. Empty fields (from a
+            # trailing comma) are dropped rather than passed as an empty string,
+            # which the known-entry check in refresh_baselines.py would reject.
+            if [ -n "${PERF_GATE_ENTRY}" ]; then
```
amd-vivekag, line 161: **[Medium]** A delimiter-only scope silently becomes a global perf refresh.

Inputs such as `,` or whitespace pass the raw non-empty check, produce no names after splitting, and leave only `--perf-gate`, which arms every matrix entry without the unscoped warning.

Fix: Parse before validation and reject a non-empty raw value when no names remain; add comma-only and whitespace-only tests.

scripts/ci/dashboard_metadata.py
```diff
@@ -293,6 +298,73 @@ def _workload_repro(
...
+                    "Both cells pass with completed_total == num_prompts * steps and "
+                    "failed_total == 0. median_ttft_ms / median_tpot_ms / "
+                    "output_throughput appear in perf.md. Serving cells are "
+                    "record-only until a baseline is blessed — see "
+                    "docs/tokenspeed-gating-rollout.md for which metrics get bounds "
+                    "and why bring-up time never does (189-379s on one node with "
+                    "nothing changed)."
+                ),
+                setup_extra=[
+                    {
```
amd-vivekag, line 356: **[Low]** This step does not pre-warm the model cache.

`docker pull` only warms the engine-image cache, and this exported `HF_HOME` is not the cache path the workload resolves.

Fix: Rename the step to pull the engine image, or actually download `Qwen/Qwen3-0.6B` into the workload's mounted cache as the running UID.

docs/tokenspeed-gating-rollout.md
```diff
@@ -0,0 +1,1067 @@
...
+
+To reproduce:
+
+```bash
+cd docker
+bash ../scripts/ci/docker_compose.sh --env-file .env.ci -f docker-compose.build.yaml build
+docker run --rm aorta:ci-gpu docker --version
+
+# work root must be root-owned; creating it through the daemon is the easy way
+docker run --rm -v /tmp:/mnt busybox:1.37 mkdir -p /mnt/ts-work-serve
```
amd-vivekag: **[Low]** This command does not repair the documented ownership failure.

If `/tmp/ts-work-serve` already exists under a user UID, `mkdir -p` as root leaves its owner and mode unchanged, so the next command still fails.

Fix: Require a fresh path, or explicitly verify and correct ownership and permissions before starting Compose.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

I reviewed the full current diff across the workflow/Docker boundary, runtime gating, tests, and rollout docs. CI is green, but the baseline rollout path is still broken and the host-root socket is enabled before its stated sign-off; all four prior Copilot findings remain open at this head. I also found three new guard/input/documentation gaps.

**Findings:** 3 High · 2 Medium · 2 Low

**Still open from last review**
- **High:** The nightly mounts the root-equivalent Docker socket while single-tenant/sign-off confirmation is still pending.
- **High:** The new daemon-dependent matrix entry makes every baseline refresh fail, including the documented serving bless.
- **High:** A scoped refresh silently deletes performance gates previously blessed for other entries.

No prior concern is resolved yet; no commits landed after those comments.

.github/workflows/nightly-eval.yml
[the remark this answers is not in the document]
vivekkhandelwal1: The sign-off has since been given — for this lane and nothing else — so the flag is staying. What you caught is real, though: the PR description and the rollout doc both still said the decision was open, so the diff and the prose disagreed and a reader had no way to tell which was current. That contradiction was the actual defect, and it is fixed in ed7f65f.

The argument now lives in `docs/tokenspeed-gating-rollout.md` under *Security: the marginal risk, and what actually bounds it*, rather than in a review thread, because the next person deciding whether a **second** lane may have the socket needs the reasoning and not the verdict:

- **The privilege is already there.** `eval-reusable.yml:58-69` runs `docker run -v "$ws":/ws …` and `sudo -n` as *plain host steps*, before any container exists — the container setup is line 86. `refresh-baselines.yml` opens the same way. The runner account therefore already has direct daemon access *and* passwordless sudo, so any job step that can execute there is host-root capable with or without this mount. The socket relocates an existing capability into `aorta-ci-gpu`; it does not create one.
- **Reachability is bounded by write access, and each part is greppable.** No `pull_request_target` anywhere in `.github/`. Both PR-reachable routes to the runner carry `head.repo.full_name == github.repository` (`gpu-tests.yml`, twice, and `bump-validate.yml`). `gpu-tests.yml` does not go through `eval-reusable.yml` and passes no socket argument at all. `bump-validate.yml` pins `docker_socket: false` explicitly rather than inheriting a default a later edit could flip.

One part I'd push back on: *"It also makes the new skip-safe staging behavior irrelevant on the nightly lane."* The skip path is what keeps the grant **scoped**. `refresh-baselines.yml` and `bump-validate.yml` run the same matrix without the socket, and as of a4ad506 the baseline refresher honours `needs_docker_daemon` as well — which it previously did not, and that was the bug you flagged one file over. So the capability declaration is load-bearing on every lane *except* the one that now has the socket. Without it, this row aborts every baseline refresh in the repo.

Also withdrawn in the same commit: the "single-tenant runner" framing this PR previously leaned on. `aorta` and `aorta-internal` hold separate repo-level runner registrations, and the internal one's July jobs reported machine `sharkmi300x-3` rather than the MI350 host, so shared tenancy is unconfirmed. It does not need to be — tenancy bears on *measurement*, which the ten-night window and the two-cell control handle, not on this grant.

config/ci/nightly_eval_matrix.yaml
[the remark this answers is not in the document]
vivekkhandelwal1: Correct, and this was the worst of the set operationally — the refresh is exactly the step that blesses metrics at the end of the record-only window, so this row would have broken blessing for `gpu_smoke`, `inference_offline` and every other workload that has nothing to do with serving.

Fixed in a4ad506. `refresh_baselines.build_baselines` now honours `needs_docker_daemon` before it calls `run_entry` (`scripts/ci/refresh_baselines.py:78`), giving the entry the same contract `min_gpus` already gives the 8-GPU rows: skip, carry over any existing baselines, and record the reason in the refresh log. Both consumers of the matrix honour the flag now; previously only `nightly_eval.evaluate` did, which is precisely the asymmetry you identified.

Tests in `tests/ci/test_nightly_eval.py`:

- `test_a_refresh_skips_a_daemon_dependent_entry_rather_than_failing` — asserts `ran == ["gpu_smoke"]`, i.e. the unrelated refresh completes and the serving row never reaches `run_entry`.
- `test_a_daemon_dependent_entry_keeps_its_baseline_on_a_lane_with_no_socket` — the carry-over, so a refresh that never measured the cell cannot revert it to record-only.
- `test_a_daemon_dependent_entry_is_refreshed_when_a_daemon_is_reachable` — the other half, so the flag cannot make the entry permanently unblessable.
- `test_the_probe_is_only_consulted_by_entries_that_declare_it` — a broken daemon must not quarantine the rest of the refresh.

On the second half — *"Add a trusted, default-off socket input to the refresh workflow and/or carry daemon-dependent entries when unavailable, while failing if the explicitly selected target cannot run."* I've done the second and third and **deliberately not the first**, so flagging that as a decline rather than an oversight.

Adding a socket input to `refresh-baselines.yml` would widen a grant that was signed off for the nightly lane specifically. It would also buy very little: rollout step 6 replaces every number the refresher derives with one computed from the ten-night window and deletes most of the keys it writes, so what the refresher contributes for this entry is a skeleton. Naming a daemon-dependent entry in `--perf-gate-entry` on a lane that cannot run it is therefore a **hard error** with a message pointing at the two real options, rather than a silent skip — covered by `test_blessing_a_daemon_dependent_entry_that_cannot_run_is_refused`. Carrying it silently would emit a correctness-only file that reads in the PR diff exactly like a successful bless of a cell that never ran, which is the same failure mode the unknown-entry-name check already refuses. The rollout doc's step 5 now carries an `[!IMPORTANT]` note saying this entry is hand-written rather than refresher-derived, and why.

If the refresh lane should get the socket later, that's a separate decision with a much easier case (dispatch-only, so the fork-guard half of the argument is trivial) — but it shouldn't ride along here.

scripts/ci/refresh_baselines.py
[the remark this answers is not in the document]
vivekkhandelwal1: Agreed, and this was the worst defect in the set: it is the exact inverse of what `--perf-gate-entry` was added to do. The scope exists to **protect** the other sixteen cells, and instead it disarmed them.

Fixed in a4ad506. `scripts/ci/refresh_baselines.py:149` now computes `derive_perf` once, and when a run is *not* deriving bounds for a cell, that cell's existing `step_time_ms` and its non-equal-policy metric specs are carried over verbatim.

Two details worth calling out, both beyond what the comment strictly asked for:

1. **The carry-over is keyed on "not deriving" rather than on "out of scope."** The same hole was reachable through the default correctness-only mode — which is the *routine* dispatch of `refresh-baselines.yml` — so scoping the fix to `--perf-gate` runs would have left the more common path still disarming everything. Omitting `--perf-gate` now means "do not arm new gates", never "disarm the ones already blessed".
2. **A spec that names no policy is carried too.** `compare_to_baseline` falls back to the allowlist when a spec omits `policy`, so a hand-written `{"value": 2.0}` is a live gate and must survive. Correctness (`equal`-policy) metrics are the one deliberate exclusion, since re-deriving those is the point of a refresh; they are applied after the carry-over so this run's observation wins.

Tests in `tests/ci/test_nightly_eval.py`, including the one you asked for:

- `test_a_scoped_refresh_leaves_other_entries_bounds_untouched` — fails if an out-of-scope entry's `step_time_ms` ceiling or `tokens_per_sec` floor changes, while asserting its correctness data *is* refreshed and the in-scope entry *does* get new bounds.
- `test_a_correctness_only_refresh_does_not_disarm_existing_gates` — the same hole through the default mode.
- `test_an_in_scope_entry_has_its_stale_bounds_replaced` — the carry-over must not shadow the derivation it's protecting.
- `test_a_carried_bound_is_not_mutated_by_the_refresh` — carried by value (`copy.deepcopy`), so the written document cannot alias the input.
- `test_a_hand_written_bound_without_a_policy_is_also_carried`.

`docs/ci-nightly-eval.md` and the rollout doc's "whole-file caveat" are both updated: the only thing that now re-arms a deliberately demoted cell is an *unscoped* `--perf-gate` refresh, which re-derives every cell by definition.

scripts/ci/dashboard_metadata.py
```diff
@@ -293,6 +298,73 @@ def _workload_repro(
...
+                        "title": "Serving-specific: pre-warm the model cache as the running uid",
+                        "commands": [
+                            "export HF_HOME=/tmp/ts-work-serve/u$(id -u)/hf",
+                            "# run_as_current_user defaults to true: a cache populated by a",
+                            "# root container leaves the trial failing with PermissionError",
+                            "docker pull lightseekorg/tokenspeed-amd@sha256:60c12e37c01496891053b9c30c4204e5d1cf9b4b641859d3aadcbd95bccc7c78",
+                        ],
+                    },
+                ],
```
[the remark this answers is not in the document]
vivekkhandelwal1: Correct on both counts, and fixed in 7837e91 by doing both of the things you offered as alternatives rather than picking one — a step that looks like it warms a cache and doesn't is worse than no step, because the next reader budgets for it.

The image pull is now its own step, honestly titled *"Serving-specific: pull the pinned engine image"*, so neither title claims the other's work. The pre-warm step (`scripts/ci/dashboard_metadata.py:363`) actually populates the cache the workload resolves:

```bash
mkdir -p /tmp/ts-work-serve/u$(id -u)/hf
HF_HOME=/tmp/ts-work-serve/u$(id -u)/hf hf download Qwen/Qwen3-0.6B
```

That is the path `tokenspeed_serve.py:730-731` derives from `workload_config.hf_home` falling back to `<work_dir>/u<uid>/hf`, and it matches the form `recipes/tokenspeed/tokenspeed-serve-gptoss.yaml:19` already documents. Run as the invoking uid, since `run_as_current_user` defaults to true and a cache populated by a root container leaves the trial failing with `PermissionError` — that comment survived the rewrite because it is still the trap.

`test_the_serving_prewarm_step_populates_the_cache_the_workload_resolves` resolves the expected path from the recipe rather than repeating the literal, asserts the download targets it, and fails if the recipe ever sets `hf_home` — at which point that becomes the resolved path and the command has to follow it there. It also asserts `docker pull` is *not* in the pre-warm step, which is how the original claim came to be wrong in the first place.

.github/workflows/nightly-eval.yml
vivekkhandelwal1: **[High]** Addressed — but by fixing the documentation rather than the code, because the approval you are asking for does exist. It had simply never been written down anywhere a reviewer could see it, so the diff and the prose contradicted each other and you were right to read that as the code getting ahead of the decision.

Recorded now in the PR description and in `docs/tokenspeed-gating-rollout.md` under *Security: the marginal risk, and what actually bounds it* (ed7f65f), with the reasoning rather than just the verdict — a future reader deciding whether a **second** lane may have the socket needs the argument:

- **The privilege is already on this runner.** `eval-reusable.yml:58-69` runs `docker run -v "$ws":/ws …` and `sudo -n` as *plain host steps*, before any container exists (setup is at line 86), and `refresh-baselines.yml` opens identically. The runner account has direct daemon access and passwordless sudo today, so any job step that can execute there is already host-root capable. The mount moves that capability inside `aorta-ci-gpu`; it does not introduce it.
- **Reachability equals write access, and every part is checkable by grep.** No `pull_request_target` anywhere in `.github/`; both PR-reachable routes to the runner gate on `head.repo.full_name == github.repository` (`gpu-tests.yml` twice, `bump-validate.yml`); `gpu-tests.yml` passes no socket argument; `bump-validate.yml` pins `docker_socket: false` explicitly.

So the population that can reach the daemon through this mount is exactly the population that can push a branch here — which is already the population that can `sudo` on the runner through any workflow step.

The **"single-tenant"** premise the PR previously rested on is withdrawn rather than confirmed. `aorta` and `aorta-internal` hold separate repo-level runner registrations, and the internal one's July jobs reported machine `sharkmi300x-3` rather than the MI350 host, so co-tenancy is genuinely unconfirmed. The argument above never depended on it; tenancy matters for *measurement* (a neighbour's load moving the numbers), which the ten-night window and the two-cell control handle.

On *"then enable it in a separate reviewed change"* — I'd argue against splitting it, and this is the one point where I'm declining rather than deferring. A follow-up PR whose whole diff is one boolean attracts less scrutiny than this one, not more: everything a reviewer needs to judge the grant (the fork guards, the per-lane pinning, the skip contract, the guard tests) is in *this* diff. What was missing was the written record, not the review. If you'd still prefer it split after reading the security section, say so and I'll pull the one line out.

config/ci/nightly_eval_matrix.yaml
vivekkhandelwal1: **[High]** Fixed in a4ad506. Your diagnosis was exact: `build_baselines()` ran every GPU-eligible entry, the refresh lane mounts no socket, so the row produced no `matrix.json`, routed into `incomplete`, and tripped the atomic abort — taking down every unrelated workload's refresh with it.

`scripts/ci/refresh_baselines.py:78` now checks `needs_docker_daemon` before calling `run_entry`, so the entry gets the same contract `min_gpus` gives the 8-GPU rows: skip, preserve existing baselines, record the reason. Both consumers of the matrix honour the flag now, not just `nightly_eval.evaluate`.

Covered in `tests/ci/test_nightly_eval.py` by `test_a_refresh_skips_a_daemon_dependent_entry_rather_than_failing` (the unrelated refresh completes, and the serving row never reaches `run_entry`), `test_a_daemon_dependent_entry_keeps_its_baseline_on_a_lane_with_no_socket` (preservation), `test_a_daemon_dependent_entry_is_refreshed_when_a_daemon_is_reachable` (the skip disappears on a lane that has the socket, with no further edit), and `test_the_probe_is_only_consulted_by_entries_that_declare_it`.

On *"and provide an approved socket-enabled path when this entry is explicitly selected"* — **declining that half**, deliberately, so it doesn't look like an oversight.

Giving `refresh-baselines.yml` a socket input would widen a grant that was signed off for the nightly lane specifically, and it would buy very little. Rollout step 6 replaces every number the refresher derives with one computed from the ten-night window and deletes most of the keys it writes — including `step_time_ms.max`, which is the one that arms the gate the observed compile excursion breaches hardest. What the refresher would contribute for this entry is a skeleton.

So instead, selecting a daemon-dependent entry on a lane that cannot run it is a **hard error** naming the two real options (dispatch from a lane with the socket, or hand-write the bounds), rather than a skip. Silently carrying it would produce a correctness-only file that reads in the PR diff exactly like a successful bless of a cell that never ran — the same failure the unknown-name check already refuses. See `test_blessing_a_daemon_dependent_entry_that_cannot_run_is_refused`, and the new `[!IMPORTANT]` note in the rollout doc's step 5.

Extending the sign-off to the refresh lane later is a much easier case to make (it's dispatch-only, so the fork-guard half is trivially satisfied), but it should be its own decision rather than riding along in this PR.

scripts/ci/refresh_baselines.py
vivekkhandelwal1: **[High]** Fixed in a4ad506, and I agree with the severity — this is the inverse of the flag's purpose. `--perf-gate-entry` was added so that blessing one workload leaves the other sixteen cells alone; instead a scoped bless of B ungated A.

Implemented as you suggested: refreshed correctness fields are merged into each out-of-scope existing spec while its performance bounds are preserved. `scripts/ci/refresh_baselines.py:149` computes `derive_perf` once, and when the run is not deriving bounds for a cell, that cell's `step_time_ms` and non-equal-policy metric specs are carried over verbatim, deep-copied so the written document cannot alias the input.

Two things I extended beyond the comment:

1. **Keyed on "not deriving", not on "out of scope."** The same hole was reachable through the default correctness-only mode — the routine dispatch of this workflow — so fixing only the `--perf-gate` path would have left the more common one still disarming every gated cell. Omitting `--perf-gate` now means "do not arm new gates", never "disarm the ones already blessed".
2. **Specs with no explicit `policy` are carried too**, since `compare_to_baseline` falls back to the allowlist and a hand-written `{"value": 2.0}` is therefore a live gate.

Tests in `tests/ci/test_nightly_eval.py`: `test_a_scoped_refresh_leaves_other_entries_bounds_untouched`, `test_a_correctness_only_refresh_does_not_disarm_existing_gates`, `test_an_in_scope_entry_has_its_stale_bounds_replaced` (the carry-over must not shadow the derivation it protects), `test_a_carried_bound_is_not_mutated_by_the_refresh`, and `test_a_hand_written_bound_without_a_policy_is_also_carried`.

`docs/ci-nightly-eval.md` and the rollout doc's whole-file caveat are updated to match: a hand-reverted cell now stays reverted through a scoped refresh of another entry *and* through a default correctness-only refresh, and only an unscoped `--perf-gate` re-arms it.

tests/ci/test_dashboard_and_alert.py
vivekkhandelwal1: **[Medium]** Fixed in 99c239c. Agreed on the framing, and it's the reason I treated this as higher priority than "Medium" suggests: this test exists to make the socket scoping unfalsifiable, so a hole in it is worse than no test — it is also a claim.

The literal `true` comparison is replaced with a fail-closed classifier, `_socket_is_statically_off` (`tests/ci/test_dashboard_and_alert.py:2021`). Exactly two values are provably off by reading the file: the input not being passed at all (both the action and the reusable workflow default to `false`), and an explicit literal false. Everything else — `${{ true }}`, `${{ !false }}`, `${{ inputs.x }}`, `${{ github.event_name != 'pull_request' }}`, `1`, `yes`, `on` — is a value whose runtime truth cannot be decided statically, and GitHub's truthiness rules are not this file's to reimplement.

The forwarding step needs an exemption or every caller of `eval-reusable.yml` looks armed, but it is deliberately narrow: only a value matching the input that workflow was *discovered* to forward. So `${{ inputs.docker_socket || true }}` sitting in that position is not waved through as plumbing.

Mutation-tested the way the existing six guards were, against the real workflow tree with one value replaced and run through the same `_pr_socket_offenders` the guard itself uses — so it is that guard being mutated, not a copy of it:

- `test_a_pr_lane_cannot_smuggle_the_socket_through_an_expression` — eight mutants on `bump-validate.yml`, including the `${{ true }}` case you named, plus `${{ !false }}`, `1`, `yes` and `${{ inputs.docker_socket }}`.
- The same test also deletes bump-validate's explicit `false` and asserts a **new** pinning check fires: a PR-triggered lane must say `false` out loud rather than inherit a default, since a later change to that default would otherwise arm it silently.
- `test_a_hardcoded_socket_inside_the_reusable_workflow_is_caught` — three mutants in the exempted forwarding position, which would hand the socket to every caller at once.
- `test_the_socket_guard_treats_anything_but_an_explicit_false_as_enabling` — the classifier itself, tabulated over 6 off-values and 13 on-values.

.github/workflows/refresh-baselines.yml
vivekkhandelwal1: **[Medium]** Fixed in 27086f0, and I'd rate this alongside the gate-stripping one: same failure class, and arguably nastier because the operator has positive evidence they scoped it.

Parsing now happens **before** validation, using the same delimiters as the container-side split so the host check is a genuine preview of it rather than a second opinion. A raw value that is non-empty but yields no names is rejected (`.github/workflows/refresh-baselines.yml:99`), and the unscoped warning is keyed on the *parsed* result, so `,` no longer suppresses it by being technically non-empty.

Three things beyond the literal ask:

1. **`set -f` before the split.** An unquoted expansion is also a glob, so a scope of `*` would otherwise have become a list of filenames from the workspace root — turning a typo into a long list of `--perf-gate-entry` flags. Covered by `test_the_scope_is_not_glob_expanded_against_the_workspace`.
2. **A second refusal inside the container**, at the point where the degradation would actually happen. The flags are built container-side after the wheel install, so that is the guard that matters for correctness; the host-side one exists to fail in seconds instead. Covered by `test_the_container_side_split_also_refuses_to_degrade_to_a_global_refresh`.
3. **The scope-without-`perf_gate` check now runs on parsed names too**, so `perf_gate: false` with a scope of `,` is rejected rather than read as "no scope, nothing to complain about".

Tests are in `tests/ci/test_dashboard_and_alert.py` and **execute the workflow's real validation body** rather than asserting on its shape — it takes its inputs through `env:` rather than `${{ }}`, so it is ordinary bash and can be run directly:

- `test_a_delimiter_only_scope_is_rejected_rather_than_silently_global` — the comma-only and whitespace-only cases you asked for: `,`, `, ,`, `,,,`, `" "`, `"\t"`, `" , \t "`.
- `test_a_real_scope_and_an_empty_one_both_still_pass_validation` — the fix must not cost the two legitimate invocations, and a scoped refresh must not warn.
- `test_a_scope_without_perf_gate_is_still_rejected`.

scripts/ci/dashboard_metadata.py
vivekkhandelwal1: **[Low]** Fixed in 7837e91, doing both of the things you offered rather than choosing: the step is renamed *and* the download is real.

*"Rename the step to pull the engine image"* — done, that is now its own step. *"Or actually download `Qwen/Qwen3-0.6B` into the workload's mounted cache as the running UID"* — also done, as a separate step (`scripts/ci/dashboard_metadata.py:363`):

```bash
mkdir -p /tmp/ts-work-serve/u$(id -u)/hf
HF_HOME=/tmp/ts-work-serve/u$(id -u)/hf hf download Qwen/Qwen3-0.6B
```

which is the path `tokenspeed_serve.py:730-731` resolves from `workload_config.hf_home` → `<work_dir>/u<uid>/hf`, and the same form `recipes/tokenspeed/tokenspeed-serve-gptoss.yaml:19` documents. As the running uid, per your note and because `run_as_current_user` defaults to true.

`test_the_serving_prewarm_step_populates_the_cache_the_workload_resolves` derives the expected path from the recipe rather than hardcoding it, so it stays correct if `work_dir` moves, and hard-fails if the recipe ever gains an `hf_home` (which would become the resolved path). It also asserts the image pull is *not* inside the model-cache step — conflating the two is exactly how the original claim came to be wrong.

docs/tokenspeed-gating-rollout.md
vivekkhandelwal1: **[Low]** Fixed in ed7f65f — and low severity undersells it slightly, because this is the command that is about to be run on the runner before the entry first executes, and the leftover-directory case is the *likely* one rather than the exotic one: a host-side sweep taken by hand as an ordinary user creates exactly that state.

You're right that `mkdir -p` is the whole problem. As root over an existing user-owned `/tmp/ts-work-serve` it returns 0 and changes neither owner nor mode, so the next run fails the same ownership check with the remedy apparently already applied — the worst shape for a documented fix.

Of the two options you offered I took the second, since requiring a fresh path pushes the problem onto whoever hits it. Both occurrences in `docs/tokenspeed-gating-rollout.md` (the `work_dir` discussion and the reproduction recipe) now state the end state explicitly and are idempotent whether or not the directory exists:

```bash
docker run --rm -v /tmp:/mnt busybox:1.37 sh -c \
  "mkdir -p /mnt/ts-work-serve && chown 0:0 /mnt/ts-work-serve && chmod 1777 /mnt/ts-work-serve"
docker run --rm -v /tmp:/mnt busybox:1.37 stat -c "%n %u:%g %a" /mnt/ts-work-serve
```

Two notes that go with it in the doc:

- **`chown` on the root only, never `-R`.** The per-uid scratch directories beneath it belong to whichever uid created them, and `-R` would take those from a co-tenant.
- **The `stat` line is worth running.** The check the workload makes is on the parent root's owner — neither 0 nor the caller's uid is a hard fail — so `0:0 1777` is the entire acceptance criterion and reading it back is cheaper than diagnosing the failure later.

This is also listed as one of two operational preconditions in *What is still missing*, so it isn't only discoverable from the reproduction recipe.

.github/workflows/nightly-eval.yml
amd-vivekag: Resolved — thanks. Sign-off is now in the workflow comment and the rollout doc; nightly-only remains the right grant.

config/ci/nightly_eval_matrix.yaml
amd-vivekag: Resolved — thanks. `build_baselines` now skips daemon-dependent entries the same way `min_gpus` does, and naming one in `--perf-gate-entry` is a hard error.

scripts/ci/refresh_baselines.py
amd-vivekag: Resolved — thanks. Out-of-scope and correctness-only refreshes now carry existing perf specs instead of rebuilding them as correctness-only.

tests/ci/test_dashboard_and_alert.py
amd-vivekag: Resolved — thanks. The classifier is fail-closed now, and the mutations cover the `${{ true }}` hole.

.github/workflows/refresh-baselines.yml
amd-vivekag: Resolved — thanks. Delimiter-only scopes are rejected on the host and again inside the container.

scripts/ci/dashboard_metadata.py
amd-vivekag: Resolved — thanks. Image pull and `hf download` into the workload-resolved cache are now separate, real steps.

docs/tokenspeed-gating-rollout.md
amd-vivekag: Resolved — thanks. The work-root recipe now chowns/chmods and reads back owner+mode.

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

Re-review of the current head (`ed7f65f`). Last round's seven comments are all Fixed (correct): the socket grant is documented as nightly-only, daemon-dependent rows skip on refresh, scoped/correctness-only refreshes carry existing perf specs, the socket guard is fail-closed, delimiter-only scopes are refused, the serving pre-warm actually downloads weights, and the work-root command repairs owner+mode. The new commits also add the right tests.

One gap remains in the same family as the last High. Copilot flagged it too: a `--perf-gate` re-bless of the *same* cell still rebuilds `metric_specs` from auto-gateable observations only, so a hand-written `_NO_AUTO_GATE` bound (`median_itl_ms`) is dropped. That contradicts the contract this PR added.

**Prior comments:** 7/7 Fixed (correct). Nothing left open from last review.

scripts/ci/refresh_baselines.py
```diff
@@ -93,22 +135,63 @@ def build_baselines(
...
+            # observation wins for the metrics it did re-derive.
             for mname, value in summary.items():
                 if value is not None and eval_lib.is_correctness_metric(mname):
                     metric_specs[mname] = {"policy": "equal", "value": value}
 
-            # Performance thresholds (min/max) are opt-in via --perf-gate. Only
-            # allowlisted metrics are gated; unknown metrics (step_time_p99,
-            # final_loss, ...) are NEVER auto-gated as min.
-            if perf_gate:
+            if derive_perf:
```
amd-vivekag, line 217: **[Medium]** A `--perf-gate` re-bless of this cell still drops hand-written `_NO_AUTO_GATE` bounds.

Carry-over only runs when `derive_perf` is false, so `median_itl_ms` (gateable by hand, never auto-armed) is rewritten away on the next scoped bless of the same entry. Copilot flagged this too.

Fix: before deriving auto-gateable specs, copy existing non-auto-gateable / non-equal metric specs forward. Add a test that a hand-written `median_itl_ms` bound survives `--perf-gate --perf-gate-entry` on that same entry.

```suggestion
            if derive_perf:
                # Hand-written bounds for metrics --perf-gate must never arm
                # (see _NO_AUTO_GATE) cannot be re-derived, so they have to
                # survive a re-bless of this same cell.
                existing_spec = existing_baselines.get(key) or {}
                for mname, mspec in (existing_spec.get("metrics") or {}).items():
                    if eval_lib.is_auto_gateable(mname):
                        continue
                    effective = (mspec or {}).get("policy") or eval_lib.metric_policy(mname)
                    if effective == "equal":
                        continue
                    metric_specs[mname] = copy.deepcopy(mspec)
                st = metrics.get("mean_step_time_ms")
                if st is not None:
                    spec["step_time_ms"] = {"max": round(st * (1.0 + step_time_margin), 4)}
                for mname, value in summary.items():
                    if value is None or not eval_lib.is_auto_gateable(mname):
                        continue
                    policy = eval_lib.metric_policy(mname)  # "min" or "max"
                    if policy == "min":
                        bound = round(value * (1.0 - throughput_margin), 4)
                    else:  # "max"
                        bound = round(value * (1.0 + step_time_margin), 4)
                    metric_specs[mname] = {"policy": policy, "value": bound}
```

vivekkhandelwal1: **[Medium]** Good catch, and you're right — this was a real hole in the fix I just made, not a pre-existing one. Copilot flagged the same thing in its 06:58 pass (suppressed comment on `refresh_baselines.py:201`), so two independent reads landed on it. Fixed in 73c9f22.

The bug was that I keyed the carry-over on `not derive_perf`, which is the correct condition for *auto-gateable* bounds — those get re-derived, so preserving them would shadow the derivation — but wrong for bounds the refresher is structurally incapable of producing. `median_itl_ms` is allowlisted yet in `_NO_AUTO_GATE` precisely because it is measured at ~0 and a relative margin around zero is not a bound (an observation of `0.0` would bless a ceiling of `0.0`, which no run can pass). So `--perf-gate` has nothing to write in its place, and dropping it was a silent disarm of exactly the kind the original comment was about. The invariant I should have written down the first time: **a refresh may decline to arm a gate; it must never disarm one.**

I took your suggestion, with two changes:

1. **Factored into a shared helper** rather than a second inline loop. `_carry_metric_bounds(existing_spec, metric_specs, only_underivable=...)` now serves both paths — `only_underivable=False` for the out-of-scope/correctness-only carry, `True` for the re-bless carry. The two rules differ by exactly one predicate, and having them as separate copies is how they would drift.
2. **Fixed the sibling hole one level up**, which neither review mentioned. `mean_step_time_ms` is read with `.get`, so a cell can report none; with no observation there was no new ceiling to write and the existing `step_time_ms` was dropped the same way. It is kept now. That makes the next comparison fail loudly on `mean_step_time_ms missing (expected <= …)` — which `compare_to_baseline` already handles at `eval_lib.py:220-226` — rather than passing quietly against a gate that had been removed. Flagging it explicitly since it's beyond what you asked for; if you'd rather a re-bless with no step-time observation drop the ceiling instead, say so and I'll invert it.

Tests in `tests/ci/test_nightly_eval.py`, including the one you asked for:

- `test_a_re_bless_keeps_a_hand_written_no_auto_gate_bound` — a hand-written `median_itl_ms` ceiling survives `--perf-gate --perf-gate-entry` on its own entry, while its auto-gateable neighbour `median_tpot_ms` is still re-derived. It also asserts the surviving value isn't `0.0`, which is what auto-arming would have produced.
- `test_a_re_bless_keeps_a_bound_on_a_metric_off_the_allowlist` — same argument one step out, for a spec the allowlist doesn't know.
- `test_a_re_bless_without_a_step_time_observation_keeps_the_old_ceiling` — the sibling hole.
- `test_a_re_bless_does_not_resurrect_a_correctness_metric` — re-pins the deliberate exclusion on the derive path, so the carry can't start preserving stale checksums.

Mutation-tested both halves: removing the carry-forward fails the first two, removing the step-time fallback fails the third.

`docs/ci-nightly-eval.md` and the rollout doc's whole-file caveat both said an unscoped `--perf-gate` "re-derives every cell", which is now imprecise — corrected to "every cell it can", with the exception spelled out. That matters for this rollout specifically: if the ten-night window later justifies gating `median_itl_ms` by hand, a subsequent refresh of the same entry will no longer quietly undo it.

amd-vivekag: Resolved — thanks, and the fix went further than what I asked. `_carry_metric_bounds(only_underivable=True)` covers off-allowlist specs as well as `_NO_AUTO_GATE` ones, and catching the `mean_step_time_ms is None` case in the same pass was a good spot. I checked the rationale rather than taking it on trust: `eval_lib.compare_to_baseline` does treat a live `step_time_ms.max` with a missing observation as a failure, so the carried ceiling fails loudly instead of sitting there inert. Factoring both call sites through one helper is nicer than the inline duplication I suggested.

amd-vivekag (approved):

All 8 prior findings are fixed and verified at `73c9f227`: the seven from the first round, plus last round's Medium on the `--perf-gate` re-bless deleting bounds it cannot re-derive. Gap sweep over the new commit found nothing further, and CI is green.

Issue #450, opened 2026-09-03
sanitizers: daily-consan-tiny and daily-consan-gemm pair a load-only driver with consan_policy: strict, so they can never return a verdict


vivekkhandelwal1:
## Summary

`daily-consan-tiny.yaml` and `daily-consan-gemm.yaml` drive ConSan through `consan_load.hip`, which loads and instruments a code object but deliberately never dispatches it. Both recipes also set `consan_policy: strict`, and strict sets `RJ_CONSAN_MOI_REQUIRE_RECORDS`, which requires visible dynamic records. A driver that never dispatches cannot produce records, so these two recipes fail closed every time. They cannot return a verdict under their current configuration.

ConSan is behaving correctly here and says so in as many words:

```
[rocjitsu-dbi-hooks] RJ_CONSAN_MOI_REQUIRE_RECORDS requested, but 0 auto MOI report
buffer(s) contained zero visible records and no kernel dispatch packet was observed
```

The recipes surface that as `overall_verdict=error`, `execution_status=error`, reason `combined_hook_exit_86`, with zero findings. This is a recipe-configuration defect, not a rocjitsu bug.

## Why it happens

- [`consan.py:662`](https://github.com/ROCm/aorta/blob/main/src/aorta/instrumentation/rocjitsu_sanitizers/consan.py#L662) sets `RJ_CONSAN_POLICY=strict` from `consan_policy: strict`.
- [`consan_load.hip:4`](https://github.com/ROCm/aorta/blob/main/recipes/sanitizers/fixtures/kernels/consan_load.hip#L4) states its own contract: "No kernel dispatch (production StreamK GEMM kernels need hipBLASLt to launch)". It calls `hipModuleLoad` and `hipModuleUnload` and nothing else.
- No dispatch means no records, and `RJ_CONSAN_MOI_REQUIRE_RECORDS` then rejects the run.

The contrast with the third caller-supplied-object recipe is what localises this. `daily-consan-lds-dispatch.yaml` also supplies its own code object and also runs `consan_policy: strict`, but its driver [`lds_dispatch.hip:54`](https://github.com/ROCm/aorta/blob/main/recipes/sanitizers/fixtures/kernels/lds_dispatch.hip#L54) calls `hipModuleLaunchKernel`, so it produces records and returns a real `pass`. The discriminator is therefore whether the driver dispatches, not whether the object is caller-supplied.

[...]

Issue #453, opened 2026-09-03
sanitizers: nightly red since 2026-09-02 — rj_waitcheck CFG dataflow does not converge on ROCm 10's gemm_NT_M256_N4096_K1024, so the case yields error against an expected warn


vivekkhandelwal1:
`sanitizers-nightly.yml` has failed on its schedule since 2026-09-02. The gate job
`gfx950 sanitizer regression` fails because `rj_waitcheck` cannot complete its dataflow
analysis on one of the two f32 GEMM code objects the ROCm 10 base image ships, so that
kernel produces no verdict at all and the run aggregates to `error` against a baseline
that expects `warn`.

This is **not** the #450 mechanism: different sanitizer, different job, and #450's ConSan
cases pass here (see the control below).

## Timeline

| When (UTC) | Run | Trigger | Gate job |
|---|---|---|---|
| 2026-09-01 12:24 | [33507543444](https://github.com/ROCm/aorta/actions/runs/33507543444) | schedule | **success** — last green |
| 2026-09-01 13:08 | `30dcc055` ([#411](https://github.com/ROCm/aorta/pull/411)) | merge | CI base image moved to ROCm 10 |
| 2026-09-02 12:26 | [33629901482](https://github.com/ROCm/aorta/actions/runs/33629901482) | schedule | **failure** — first run on the new image |
| 2026-09-03 10:22 | [33743980591](https://github.com/ROCm/aorta/actions/runs/33743980591) | workflow_dispatch | **failure** — byte-identical signature |

The last green ran **44 minutes before** the merge, which is as tight a bracket as a
nightly can give: nothing else landed in between.

> [!NOTE]
> The run list is misleading if you read only the conclusion column. Run
> 33743980591 shows a run-level conclusion of `cancelled`, because the
> non-gating `gfx950 workload survey` job was cancelled. The gate job
> `gfx950 sanitizer regression` concluded `failure` in **both** 33629901482 and
> 33743980591.

#411 swapped the `Dockerfile.ci-gpu` base:

```
- rocm/pytorch:rocm7.2.4_ubuntu24.04_py3.12_pytorch_release_2.10.0@sha256:4449f856…
+ rocm/pytorch:rocm10.0_ubuntu26.04_py3.14_pytorch_release_2.13.0@sha256:3174cb70…
```

## Mechanism

`scripts/sanitizers/prepare_gemm_isa.py` extracts the f32 GEMM code objects from the
[...]

Issue #370, opened 2026-08-12
sanitizers-nightly: confirm and raise the effective self-hosted-runner/org job-timeout cap (~20 min observed)


amd-vivekag:
## Context

Split out from #366 (implemented in #369).

#366's original Task 3 asked to confirm and raise the **effective** job-timeout for the `sanitizers-gpu` nightly job. This cannot be done from the workflow file: `.github/workflows/sanitizers-nightly.yml` already declares `timeout-minutes: 120`, but a tighter **effective ~20-min cap has been observed** on the `[self-hosted, gpu]` (MI350x) runner — pointing to a self-hosted-runner or org-level Actions policy, not the workflow. GitHub's `timeout-minutes` can only *lower* the ceiling; it cannot override a shorter runner/org limit.

Because the fix lives in runner/org configuration (outside this repo), it was removed from #366's scope — Tasks 1 & 2 (progress logging + parallelized waitcheck) shipped in #369 — and is tracked here.

## Task

- Confirm the **effective** timeout on the `[self-hosted, gpu]` (MI350x) runner: inspect the runner service/config and any org/repo Actions policy for a binding cap shorter than the workflow's `timeout-minutes: 120`.
- Raise/remove that binding cap so a normal nightly run (build + daily 3/3 gate + the ~5-min informational gemm timeout + overhead) cannot be killed abruptly.
- Keep the workflow's explicit `timeout-minutes` as the documented ceiling once the effective limit is confirmed.

## Acceptance criteria

- [ ] The effective runner/org timeout for `sanitizers-gpu` is confirmed and documented.
- [ ] The binding cap is raised so a normal run cannot be killed mid-scan; the workflow retains an explicit, generous `timeout-minutes`.

## References

- #366 (parent), #369 (Tasks 1 & 2 landed).
- Slow run: https://github.com/ROCm/aorta/actions/runs/31511121124

Issue #359, opened 2026-08-11
Revert recipe-driven ConSan from lenient back to strict once rocjitsu MOI-inventory perf (rocm-systems#9964) is fixed


amd-vivekag:
## Context

Recipe-driven ConSan (`mode: sanitizer`, `source.consan_command`; #347) over large
code objects currently hits a rocjitsu performance bug: the ConSan **MOI inventory**
phase does not terminate for some large f32 GEMM code objects on gfx950, so a
`consan_policy: strict` run fails closed with `combined_hook_timeout` and yields no
verdict. Upstream bug: ROCm/rocm-systems#9964.

## Temporary workaround

Recipes driving ConSan over such heavy objects use `consan_policy: lenient` instead
of `strict` while ROCm/rocm-systems#9964 is open. (Note: this is a mitigation for
the missing verdict; it does not itself make heavy objects complete — that requires
the upstream fix.)

## Action (once ROCm/rocm-systems#9964 is fixed)

- Flip `consan_policy: lenient` → `strict` in the affected ConSan recipes.
- Re-run and confirm ConSan completes and returns a real `pass`/`fail` under strict.

## Acceptance criteria

- [ ] ROCm/rocm-systems#9964 resolved (MOI inventory terminates for the affected objects).
- [ ] ConSan recipes over large objects are back on `consan_policy: strict`.
- [ ] A strict run returns a real verdict (not `combined_hook_timeout`).

Blocked by ROCm/rocm-systems#9964.

amd-vivekag:
Reverted to `consan_policy: strict` now. The lenient workaround was verified **ineffective** for the timeout case: ConSan's MOI inventory runs *before* the strict/lenient policy check, so the top-5 f32 GEMM object still timed out under lenient (`combined_hook_timeout` at 3602s, identical to strict). There is therefore no lenient workaround to keep and nothing to revert later. The only remaining blocker for a real verdict on the heavy objects is the upstream rocjitsu fix (ROCm/rocm-systems#9964).

PR #454, opened 2026-09-03
fix(sanitizers): stop daily-consan-gemm discarding its own coverage, and reject vacuous sanitizer outcomes

vivekkhandelwal1:
Refs #450. **Not** `Fixes` — see "Does this close the issue?" below; the review turned up a reason it should not auto-close.

## What was wrong

`daily-consan-gemm` drove ConSan through `consan_load.hip`, a load-only driver that calls
`hipModuleLoad`/`hipModuleUnload` and never dispatches, while setting `consan_policy: strict`.
strict sets `RJ_CONSAN_MOI_REQUIRE_RECORDS`, which demands visible dynamic records, so a driver
that never dispatches failed closed with `combined_hook_exit_86` and zero findings. ConSan was
behaving correctly; the recipe was asking it for something the driver could not produce.

## Task 1 — `lenient` is a real fix on the small object, and it is genuinely not enough on the big one

On a 16,265,200-byte ROCm 7.0.2.2 object, `lenient` is not a cosmetic de-escalation. Same node,
same rocjitsu bundle, policy the only variable:

| | `consan` verdict | execution | findings | coverage |
|---|---|---|---|---|
| **strict** | `error` | `error`, `combined_hook_exit_86` | 0 | none reported |
| **lenient** | `pass` | `complete` | 0 consan + **32 waitcheck** | **access=68894/68894** |

The coverage was always there and strict discarded it: the pre-fix log shows the object
transforming cleanly (`outcome=modified-valid`, 75978 patches over 68894 sites) and only *then*
being rejected on require-records. strict was also suppressing 32 real `missing s_waitcnt`
findings on that object.

**@amd-vivekag was right that this does not generalise to CI, and the correction is the main
change in this revision.** CI runs the digest-pinned ROCm 10 image, whose object is ~10.5x
larger. Re-measured *inside that image* on gfx950, against the byte-identical object the nightly
extracts — sha256 `57c5d8ef…`, 169,996,560 bytes, matching run
[34032862308](https://github.com/ROCm/aorta/actions/runs/34032862308):

| `daily-consan-gemm`, ROCm 10 | verdict | exec | findings | reason | coverage |
|---|---|---|---|---|---|
| **strict** | `error` | `error` | 0 | `consan_strict_load_rejection` (exit 92) | — |
| **lenient** | `error` | `error` | 0 | `consan_coverage_incomplete` | **access=0/719550** |

Both fail for the same reason, and it is **not** the record requirement. ConSan needs
**1,721,958,400 bytes** of patched image against a **419,430,400-byte** ceiling — 4.11x over — so
the first-light probe rejects the transform and `patches=0`. The object is never instrumented.

Two things worth stating precisely, because they change what this PR can claim:

1. **The `timeout_seconds: 6000` ceiling is not what bites.** The run ends in 595s under strict
   and 859s under lenient. The growth ceiling is the binding constraint; the runtime question is
   simply never reached.
2. **On this object, `lenient` moves the fail-closed out of the sanitizer and into AORTA.** Under
   strict the hook terminates itself (`fail_closed=true`, `action=terminate`, exit 92). Under
   lenient it declines to (`fail_closed=false`), exits **0**, and returns an object with none of
   its 719,550 access sites patched — so only AORTA's coverage cross-check turns that into an
   error. Isolated by forcing the ceiling below what the small LDS fixture needs:
   `RJ_CONSAN_POLICY=strict` exits 92 with a `load rejection … action=terminate` line;
   `RJ_CONSAN_POLICY=default` exits 0 with the same growth rejection and **no load-rejection line
   at all**. That is the "quiet pass" hazard #450 worried about, and it is real on this object —
   the cross-check is what prevents it.

`lenient` is **kept**, because it is the right target state and a precondition: the moment the
ceiling admits the object, strict would fail it again on require-records, which is the #450
defect. It is now described as necessary rather than sufficient, and the recipe records what was
measured where.

## Task 1b — `daily-consan-tiny` is deliberately NOT changed

#450 proposed flipping both. Measured on ROCm 7.0.2.2 *and* re-measured on ROCm 10, and it does
not hold for `tiny` on either: `lenient` leaves it `error`/`error` with zero findings and only
relabels the reason to `consan_coverage_incomplete` at `access=0/0`. `tiny_vecadd`'s only memory
operations are ordinary global loads/stores, which ConSan does not admit as MOI sites, so the run
is rejected by the **coverage cross-check**, not the record requirement — and removing
`RJ_CONSAN_MOI_REQUIRE_RECORDS` cannot help something with nothing to be complete about. Its exit
86 is an intended negative control, root-caused in #409.

So `tiny` stays strict. What was wrong is that this intent lived only in a YAML comment where no
check could read it. It is now **declared** in `verdict_baselines.json` as an `expected_error`,
verified against all four ROCm 10 nightly survey trees and a direct re-run in the CI image.

## Task 2 — the check that would have caught this

`compare_verdict_baselines.py` only inspected the three **gated** cases, and these two are
informational, so an `error` verdict with zero findings was nobody's problem. #359's acceptance
criterion for the strict flip was that a run "returns a real verdict, not
`combined_hook_timeout`", which exit 86 satisfies literally.

The comparator now also **sweeps every report under the results root**, gated or not, and rejects
any that terminated `error` or `not_checked` with zero findings unless the baselines declare that
outcome. Zero findings alone is never the trigger — `daily-consan-lds-dispatch` passes clean with
none and must not trip it.

### `expected_error` is now an exact declaration, not a bypass

Addressing [@Copilot's finding](https://github.com/ROCm/aorta/pull/454#discussion_r3924815213),
which was correct on all four counts. The declaration names `execution_status`,
`overall_verdict` and the **complete set** of check reasons, and all three are matched exactly.
So a substring like `combined_hook_exit_860` no longer passes, a second undeclared reason riding
along no longer passes, and a `not_checked` run is no longer accepted where an `error` was
reviewed.

**A malformed declaration is a failure, not a bypass.** Every `expected_error` is validated when
the baselines file is loaded, whether or not that case ran — a mis-spelled `reason` for `reasons`
used to be silently ignored, which is exactly how a bypass gets in. A declared status or verdict
that could never match is rejected too, since dead configuration that reads as protection is
worse than none.

One extra hole the exactness work surfaced: `waitcheck_gemm` is **both** a gated baseline key and
what the survey directory `informational/waitcheck-gemm` resolves to — two different recipes. A
single `expected_error` there would have excused both. A gated key may no longer carry one at all.

### Absent reports no longer fail open

Addressing [@amd-vivekag's finding](https://github.com/ROCm/aorta/pull/454#discussion_r3925899487).
All six survey invocations run under `|| true`, so a recipe that dies *before* writing a report
leaves nothing to judge — the same fail-open shape the sweep exists to close, one level up.
`--require <case-dir>` makes the absence of an invoked case an error.

It is scoped to cases the caller names rather than blanket-failing, because an empty tree and a
tree from a run that was never asked to produce anything are indistinguishable from a directory
listing. A **job-level** kill needs no tolerance and never did: the runner cap in #370, a
cancellation or a step timeout takes the whole payload down, so the comparator does not run and
has no verdict to give. Only a *recipe-level* failure reaches it, and that is precisely the case
that should fail.

### The survey sweep is armed, with a self-retiring suppression

Replayed over the **real** survey trees of the last four nightlies, the sweep flags two rows:

| row | outcome | why |
|---|---|---|
| `consan_gemm` | `error`/`error`, 0 findings, `consan_strict_load_rejection` | the growth ceiling above |
| `waitcheck_gemm` | `error`/`error`, 0 findings, `worklist_not_fully_checked` | the `rj_waitcheck` non-convergence of #453, on that same object |

Neither is fixed here, so arming the sweep unqualified would make the survey job red every night
— and a check that always fails cannot distinguish a new break from the standing two. Arming it
was still the right call, so the two rows are named at the call site:

```
--known-vacuous informational/consan-gemm
--known-vacuous informational/waitcheck-gemm
```

This is the **opposite** of an `expected_error`, deliberately:

- an `expected_error` says *this outcome is correct*; `--known-vacuous` says *this outcome is
  wrong, it is tracked elsewhere, and it must not mask the other rows*;
- it lives in the workflow rather than the baselines, so it carries no dashboard meaning and
  shows up in a workflow diff;
- **it is self-retiring.** The comparator *fails* if a listed case starts producing signal
  (`Remove informational/consan-gemm from the caller's --known-vacuous list`) or stops reporting
  at all, so a fixed row cannot leave a stale suppression behind;
- a case may not be both — passing `--known-vacuous` for a row declared as an `expected_error`
  is a usage error, since the two make opposite claims;
- the suppressed message deliberately does **not** repeat the sweep's "declare it as an
  `expected_error`" advice, which would be exactly the anti-pattern #453 warns about. There is a
  test asserting that string never appears for a known-vacuous row.

**The other four rows are checked for real from the next scheduled run**, which is the #450
regression path this whole change exists to close. Verified by replaying the exact wired
invocation over runs 33755195031, 33872660689, 33965800908 and 34032862308 — exit 0, both blocked
rows reported, six reports judged:

```
consan_gemm: KNOWN VACUOUS, not failing -- execution_status='error', ... reason
  'consan_strict_load_rejection', and zero findings. This row is on the caller's
  --known-vacuous list: the outcome is wrong and tracked elsewhere, not accepted.
waitcheck_gemm: KNOWN VACUOUS, not failing -- ... reason 'worklist_not_fully_checked' ...
vacuity sweep: ok (6 report(s) carry signal or are declared) (2 known-vacuous)
```

**The gate half is unaffected**: replayed over runs
[33507543444](https://github.com/ROCm/aorta/actions/runs/33507543444) (last green) and
[34032862308](https://github.com/ROCm/aorta/actions/runs/34032862308), the sweep adds no failure
there.

26 new tests in `tests/sanitizers/test_compare_baselines.py`, including the false-positive guards
(clean pass with zero findings; an errored run that still reported findings) and the
malformed-declaration, wrong-reason, extra-reason, gated-key-collision and
known-vacuous-suppression cases (including the one that fails when a suppressed row recovers).

## Does this close the issue?

**No, and it should not.** The recipe misconfiguration #450 reported is fixed and the mechanism
that let it hide for weeks is closed. But the row #450 was filed about still produces no signal
on the image CI actually runs, for a capacity reason #450 did not know about. Auto-closing would
assert coverage that does not exist — the same anti-pattern #453 argues against. Downgraded to
`Refs`.

## A finding that belongs to #453

The `waitcheck_preflight` check in the ROCm 10 `gemm-lenient` run reported:

```
waitcheck CFG dataflow did not converge at .text+0x24d7070
(in:loadcnt,uncertain-order;out:loadcnt,uncertain-order)
```

The **same offset** #453 records, reproduced on a Slurm dev node rather than the CI runner. That
independently supports #453's diagnosis that the failure is content-driven and not
runner-specific, and confirms it reaches the survey lane as well as the gate.

## Hardware verification

Slurm `meta64` node `cv350-rck-g03-e15-18`, gfx950 / MI350X (`0x75a0`), inside
`rocm/pytorch:rocm10.0_ubuntu26.04_py3.14_pytorch_release_2.13.0@sha256:3174cb70…` — the exact
digest `docker/Dockerfile.ci-gpu` pins, with the same devel-symlink layer that Dockerfile adds,
and the nightly's own fixture-build steps. rocjitsu bundle `bed84edd`. Node-local scratch;
deliberately **not** the self-hosted CI runner `smci350-rck-g03-f16-12`, which is reserved for a
performance baseline window.

```
case                  verdict  exec      find  reason                        access
--------------------  -------  --------  ----  ----------------------------  ---------
control-lds-dispatch  pass     complete  0     -                             5/5
gemm-lenient          error    error     0     consan_coverage_incomplete    0/719550
gemm-strict           error    error     0     consan_strict_load_rejection  -
tiny-lenient          error    error     0     consan_coverage_incomplete    0/0
tiny-strict  (kept)   error    error     0     combined_hook_exit_86         -
```

`gemm-strict` reproduces the nightly's reason exactly, which is what establishes that this
environment matches CI rather than approximating it. The control `daily-consan-lds-dispatch` still
passes at `access=5/5`.

## Tests

395 in `tests/sanitizers` (26 new), 135 in `tests/instrumentation/rocjitsu_sanitizers`,
`pre-commit run --all-files` clean. Rebased onto current `main` (`4b4553e`); no file overlap with
the four commits that landed since.

### DIFF
--- a/.github/workflows/sanitizers-nightly.yml
+++ b/.github/workflows/sanitizers-nightly.yml
@@ -552,7 +552,12 @@ jobs:
               "${KROOT}/consan_load.hip" -o "${FIXTURES}/bin/consan_tiny_load"
             hipcc --offload-arch=gfx950 -DOBJECT="\"$(pwd)/${FIXTURES}/isa/consan_gemm_f32.hsaco\"" \
               "${KROOT}/consan_load.hip" -o "${FIXTURES}/bin/consan_gemm_load"
-            for c in consan-gemm consan-lds-dispatch consan-tiny; do
+            # Named once, because the vacuity check at the end of this payload
+            # has to know which cases were invoked in order to tell "produced
+            # nothing" from "was never asked to". Two lists, one per sanitizer,
+            # so each stays next to the invocations it describes.
+            CONSAN_CASES="consan-gemm consan-lds-dispatch consan-tiny"
+            for c in ${CONSAN_CASES}; do
               mkdir -p "${SANITIZER_OUT}/informational/${c}"
             done
             aorta sweep run --recipe recipes/sanitizers/daily-consan-gemm.yaml \
@@ -568,7 +573,8 @@ jobs:
             # waitcheck-gemm under one "gemm" kernel heading. Each survey waitcheck
             # scan is a dedicated best-effort run publishing its OWN report -- the
             # gated Tab-1 waitcheck report is never reused here. Non-gating.
-            for c in waitcheck-gemm waitcheck-lds-dispatch waitcheck-tiny; do
+            WAITCHECK_CASES="waitcheck-gemm waitcheck-lds-dispatch waitcheck-tiny"
+            for c in ${WAITCHECK_CASES}; do
               mkdir -p "${SANITIZER_OUT}/informational/${c}"
             done
             # daily-waitcheck-gemm-object scans the same consan_gemm_f32.hsaco object
@@ -580,6 +586,63 @@ jobs:
               --output-dir "${SANITIZER_OUT}/informational/waitcheck-lds-dispatch" || true
             aorta sweep run --recipe recipes/sanitizers/daily-waitcheck-tiny.yaml \
               --output-dir "${SANITIZER_OUT}/informational/waitcheck-tiny" || true
+
+            # Every survey recipe above runs under `|| true`, because a non-gating
+            # row must not fail the job merely for reporting something. That is
+            # what let ROCm/aorta#450 sit unnoticed: rows ended `error` with zero
+            # findings on every run, and the baseline comparison in the gate job
+            # only covers the three GATED cases, so nothing ever looked at these
+            # reports at all.
+            #
+            # So look at them. --vacuous-only asserts the one thing that is always
+            # wrong regardless of what a row is expected to observe: a run that
+            # terminated without producing any signal. Cases where that is the
+            # intended outcome declare it in verdict_baselines.json (consan_tiny
+            # does). This CAN turn the survey job red, which is the point -- the
+            # artifact upload below is `if: always()`, so Tab 2 still publishes.
+            #
+            # --require closes the same hole one level up. `|| true` also swallows
+            # a recipe that dies BEFORE writing a report -- a bad input, a missing
+            # fixture, a startup failure -- and a report that does not exist cannot
+            # be judged vacuous. Naming the six invoked cases makes their absence
+            # an error, so an entirely empty survey fails here instead of passing
+            # silently. A job-level kill (cancellation, the runner cap in #370, the
+            # timeout-minutes ceiling) does not need excusing: it takes this whole
+            # payload down, so the line below never runs at all.
+            #
+            # --known-vacuous names the two rows that are ALREADY vacuous on the
+            # ROCm 10 objects this lane scans, so the check can be armed today
+            # without being red for reasons nobody can act on here. Measured on
+            # the real survey trees of runs 33755195031, 33872660689, 33965800908
+            # and 34032862308, both on the same 169,996,560-byte Tensile object:
+            #
+            #   consan-gemm     consan_strict_load_rejection, 0 findings. ConSan
+            #                   needs 1,721,958,400 B of patched image against a
+            #                   419,430,400 B ceiling (4.11x) and never instruments
+            #                   it. consan_policy does not change that -- see
+            #                   daily-consan-gemm.yaml and
+            #                   docs/sanitizers/consan-gemm-patched-image-growth-cap.md.
+            #   waitcheck-gemm  worklist_not_fully_checked, 0 findings. The
+            #                   rj_waitcheck CFG dataflow non-convergence of #453.
+            #
+            # This is NOT a blessing and is deliberately not spelled as an
+            # expected_error: it says these outcomes are wrong and being worked,
+            # not that they are acceptable. It is self-retiring -- the comparator
+            # FAILS if a listed case starts producing signal, so a fixed row cannot
+            # leave a stale suppression behind, and it fails if a listed case has no
+            # report at all. The other four rows are checked for real from tonight,
+            # which is the #450 regression path this whole change exists to close.
+            echo "[$(date +%T)] survey: rejecting vacuous outcomes …"
+            REQUIRE=""
+            for c in ${CONSAN_CASES} ${WAITCHECK_CASES}; do
+              REQUIRE="${REQUIRE} --require informational/${c}"
+            done
+            # shellcheck disable=SC2086  # REQUIRE is a deliberate argument list
+            python scripts/sanitizers/compare_verdict_baselines.py \
+              --vacuous-only ${REQUIRE} \
+              --known-vacuous informational/consan-gemm \
+              --known-vacuous informational/waitcheck-gemm \
+              "${SANITIZER_OUT}"
           '
 
       - name: Reclaim results ownership
--- a/recipes/sanitizers/daily-consan-gemm.yaml
+++ b/recipes/sanitizers/daily-consan-gemm.yaml
@@ -5,11 +5,24 @@ description: >
   Informational (non-gating) gfx950 ConSan run over a real, heavy f32 Type_SS
   hipBLASLt Tensile code object (generic public content extracted in CI from the
   synthetic gemm_shapes_unique.csv), driven via the source.consan_command path
-  added in #347. This exercises the guardrail's behavior on large production code
-  objects: whatever ConSan concludes, strict policy must reach a real verdict or
-  fail closed, never a false pass. The observed outcome is whatever the run
-  reports -- known rocjitsu limitations hit by this object are tracked in
-  docs/sanitizers/consan-4112-overlapping-anchor-patches.md. Rendered on the
+  added in #347. consan_gemm_load only LOADS that object -- it never dispatches --
+  so under consan_policy: lenient what this case asserts is static instrumentation
+  coverage on a real production object: that ConSan reads it, patches it, and
+  reports how many of its discovered access sites it actually covered. It asserts
+  nothing about dynamic race evidence; daily-consan-lds-dispatch.yaml is the
+  end-to-end control for that.
+
+  A pass here therefore means "instrumented, and the coverage cross-check agrees",
+  never merely "ran without complaining" -- under lenient ConSan itself no longer
+  fails closed, so the coverage cross-check is what stands between an
+  uninstrumented object and a false pass.
+
+  STATUS ON THE CURRENT CI IMAGE: this case does NOT reach that assertion today.
+  On the digest-pinned ROCm 10 base the object is 169,996,560 bytes and ConSan
+  rejects it on the patched-image growth ceiling before instrumenting anything, so
+  the row is error with zero findings every night. That is a capacity limit, not
+  something consan_policy can change -- see the policy comment below and
+  docs/sanitizers/consan-gemm-patched-image-growth-cap.md. Rendered on the
   dashboard's Workload survey tab (Tab 2) as an observed-only case; it does NOT
   gate the nightly.
 sanitizer_plan:
@@ -30,7 +43,72 @@ sanitizer_plan:
   sanitizers:
     - consan
   policy:
-    consan_policy: strict
+    # lenient, not strict, because consan_gemm_load is a LOAD-ONLY driver: it calls
+    # hipModuleLoad/hipModuleUnload and never dispatches (consan_load.hip states the
+    # contract -- production StreamK GEMM kernels need hipBLASLt to launch). strict
+    # sets RJ_CONSAN_MOI_REQUIRE_RECORDS, which demands visible dynamic records, so a
+    # driver that never dispatches produces no dispatch packet and no records and
+    # strict fails closed with combined_hook_exit_86 no matter how healthy the run
+    # was. That is what #450 observed: error/error, zero findings, every single time.
+    #
+    # This is the same rule harvest_code_objects.py already applies to its own
+    # load-mode recipes, for the same reason. Under lenient this case verifies what a
+    # load-only run can actually verify -- static instrumentation coverage, i.e. that
+    # ConSan reads, patches and analyzes a real production Tensile object.
+    #
+    # WHAT THAT IS MEASURED ON, AND WHAT IT IS NOT. Two different objects are in
+    # play and the policy argument only closes on one of them:
+    #
+    #   * ROCm 7.0.2.2, 16,265,200-byte object, gfx950 Slurm node. lenient is a
+    #     real fix: the object transforms clean (outcome=modified-valid, 75978
+    #     patches over 68894 discovered access sites), consan=pass at
+    #     access=68894/68894, plus 32 genuine waitcheck "missing s_waitcnt"
+    #     findings. Under strict the identical run was thrown away on
+    #     require-records at exit 86. So the coverage signal was there all along
+    #     and only the policy discarded it.
+    #   * ROCm 10.0.0, 169,996,560-byte object -- the one CI actually scans.
+    #     lenient does NOT fix this case, and it was wrong of an earlier revision
+    #     of this comment to imply the 7.0.2 result generalised. Measured inside
+    #     the digest-pinned CI base image
+    #     (rocm/pytorch:rocm10.0_ubuntu26.04_py3.14_pytorch_release_2.13.0@sha256:3174cb70…)
+    #     on gfx950, against the byte-identical object the nightly extracts
+    #     (sha256 57c5d8ef…, matching run 34032862308):
+    #
+    #       strict   error/error, 0 findings, consan_strict_load_rejection (exit 92)
+    #       lenient  error/error, 0 findings, consan_coverage_incomplete,
+    #                access=0/719550, barriers 0/109497
+    #
+    #     Both fail for the SAME underlying reason, which is not the record
+    #     requirement: ConSan needs 1,721,958,400 bytes of patched image against a
+    #     419,430,400-byte ceiling (4.11x over), so the first-light probe rejects
+    #     the transform and patches=0. The 6000s ceiling below is not what bites --
+    #     the run ends in 859s.
+    #
+    #     What the policy DOES change there is who fails it. Under strict the hook
+    #     terminates itself (fail_closed=true, action=terminate, exit 92). Under
+    #     lenient it declines to (fail_closed=false), exits 0, and hands back an
+    #     object with none of its 719,550 access sites patched; only aorta's
+    #     coverage cross-check turns that into an error. Confirmed in isolation by
+    #     forcing the ceiling below what the small LDS fixture needs, where
+    #     RJ_CONSAN_POLICY=strict exits 92 with a load-rejection line and
+    #     RJ_CONSAN_POLICY=default exits 0 with the same growth rejection and no
+    #     load-rejection line at all. So lenient is necessary for a load-only
+    #     driver but not sufficient for this object, and on this object it also
+    #     makes the coverage cross-check load-bearing rather than a second opinion.
+    #
+    # lenient is kept because it is the right target state and a precondition for
+    # this case ever passing: the moment the growth ceiling admits the object,
+    # strict would fail it again on require-records -- the #450 defect -- whereas
+    # lenient would report the coverage it measured. Fixing the capacity limit is
+    # tracked below and in docs/sanitizers/consan-gemm-patched-image-growth-cap.md;
+    # note that document measures a 13.6 MiB per-chip sibling object that DOES
+    # clear the ceiling on ROCm >= 7.14, which the move to ROCm 10 has now made
+    # reachable, at the cost of representing an MI355X-class device.
+    #
+    # Dynamic race evidence over a dispatching GEMM is genuinely out of scope here and
+    # is tracked separately; daily-consan-lds-dispatch.yaml is the end-to-end control
+    # that proves the record/replay path works on a caller-supplied object.
+    consan_policy: lenient
     on_missing_backend: fail
     # Raised from the 300s cap of #364, which was chosen only because ROCm/rocm-systems#9964
     # made MOI inventory non-terminating: every ceiling produced the same
@@ -122,20 +200,31 @@ sanitizer_plan:
     # ROCm/rocm-systems#10686. If that lands, this ceiling can drop to roughly a
     # quarter of its current value.
     #
-    # !! STALE AS A BUDGET, 2026-08-27 !! Everything above was measured against a
-    # 16,265,200-byte (15.5 MB) object on ROCm 7.0.2.2. The object CI extracts on
-    # the 7.2.4 base is 191,935,808 bytes (~183 MiB) with 637,823 access ranges --
-    # 11.8x the bytes and 9.3x the sites -- so these figures no longer describe this
-    # case. See prepare_gemm_isa.py's docstring: the size is a per-release property
-    # of the shipped Tensile libraries, not something this repo pins.
+    # !! STALE AS A BUDGET, re-confirmed against ROCm 10 on 2026-09-07 !! Everything
+    # above was measured against a 16,265,200-byte (15.5 MB) object on ROCm 7.0.2.2.
+    # The object CI extracts on the current ROCm 10 base is 169,996,560 bytes
+    # (~162 MiB) with 719,550 access sites and 109,497 barrier sites -- 10.5x the
+    # bytes and 10.4x the sites -- so these figures do not describe this case. See
+    # prepare_gemm_isa.py's docstring: the size is a per-release property of the
+    # shipped Tensile libraries, not something this repo pins. (The 7.2.4 base this
+    # block used to cite was 191,935,808 bytes with 637,823 access ranges; ROCm 10
+    # is smaller on both axes but still far over the ceiling.)
     #
     # The ceiling is currently NOT what limits this case. ConSan rejects the object
-    # before instrumenting it, on the patched-image growth policy (needs ~1.39 GiB
-    # against a 400 MiB default), so the run ends at ~730s with exit 92
-    # status=4112. Raising RJ_CONSAN_MAX_PATCHED_IMAGE_GROWTH_PERCENT past ~778
-    # would let the transform proceed -- and 6000s would then very likely be too
-    # small, because 9.3x the sites is work no measurement here covers. Re-measure
-    # before assuming this number means anything.
+    # before instrumenting it, on the patched-image growth policy: it needs
+    # 1,721,958,400 bytes against the 419,430,400-byte default, i.e. 4.11x over, a
+    # 10.1x expansion of the input. patches=0, so nothing is instrumented. Measured
+    # in the digest-pinned CI image, where the run ends in 595s under strict (exit
+    # 92) and 859s under lenient (exit 0, caught by the coverage cross-check) --
+    # both comfortably inside this 6000s ceiling, which is why raising it fixes
+    # nothing.
+    #
+    # Raising RJ_CONSAN_MAX_PATCHED_IMAGE_GROWTH_PERCENT past ~1013 would let the
+    # transform proceed -- and 6000s would then very likely be too small, because
+    # 10.4x the sites is work no measurement here covers, and the survey job itself
+    # caps at timeout-minutes: 120. So raising the ceiling alone most likely trades
+    # this rejection for combined_hook_timeout. Re-measure before assuming either
+    # number means anything.
     # See docs/sanitizers/consan-gemm-patched-image-growth-cap.md.
     timeout_seconds: 6000
   output:
--- a/scripts/sanitizers/compare_verdict_baselines.py
+++ b/scripts/sanitizers/compare_verdict_baselines.py
@@ -7,15 +7,87 @@ single top-level field. Each report is *strictly* reloaded through
 ``overall_verdict``/``execution_status`` actually agree with the checks), and
 then compared against the committed baseline's ``overall_verdict``,
 ``execution_status`` (when declared), per-check verdicts, and finding shape.
+
+On top of that per-case comparison the comparator sweeps *every* report under
+the results root for vacuous outcomes -- a run that terminated ``error`` or
+``not_checked`` and produced no findings at all. Such a run is a broken
+configuration, not a result: the sanitizer never got far enough to have an
+opinion, so any coverage claim derived from it is empty rather than positive.
+
+That sweep exists because of ROCm/aorta#450. ``daily-consan-tiny`` and
+``daily-consan-gemm`` paired a load-only driver with ``consan_policy: strict``,
+which sets ``RJ_CONSAN_MOI_REQUIRE_RECORDS`` and so demands dynamic records a
+non-dispatching driver can never produce. Both failed closed with
+``combined_hook_exit_86`` and zero findings on every run for weeks, and nothing
+noticed: the baseline comparison above only covers the three *gated* cases, and
+these two are informational, so an ``error`` verdict there was nobody's problem.
+The sweep is deliberately independent of whether a case is gated.
+
+A case that is *legitimately* expected to end without findings declares it in
+the baselines file as an ``expected_error`` entry, e.g. ``daily-consan-tiny``,
+whose exit 86 is an intended negative control. The declaration is an exact
+description of the one outcome that was signed off -- execution status, overall
+verdict and the complete set of check reasons -- and the run must match it on
+every field:
+
+.. code-block:: json
+
+    "consan_tiny": {
+      "expected_error": {
+        "execution_status": "error",
+        "overall_verdict": "error",
+        "reasons": ["combined_hook_exit_86"],
+        "why": "..."
+      }
+    }
+
+A declaration that is not well-formed is a **failure, not a bypass**. Every
+``expected_error`` in the baselines file is validated when the file is loaded,
+whether or not that case ran, because a mis-spelled or half-written declaration
+would otherwise silently widen an exemption that exists to cover exactly one
+case: the entry is meant to *narrow* what a row is allowed to do, so anything
+about it that cannot be checked has to be rejected rather than skipped.
+
+Absence is judged too, but only for cases the caller says it invoked. Pass
+``--require <case-dir>`` for each one; a required case with no report is an
+error. See ``_sweep_vacuous`` for why a job-level kill does not need tolerating
+here.
+
+A row that is vacuous today for a reason tracked *elsewhere* is spelled
+``--known-vacuous <case-dir>`` at the call site, and it is the opposite of an
+``expected_error``: the declaration says "this outcome is correct", whereas this
+says "this outcome is wrong, it is being worked, and it must not mask the other
+rows". Consequently it **fails when the case stops being vacuous**, so a fixed
+row cannot leave a stale suppression behind, and a case may not be both.
+Spelling it as an argument rather than a baselines entry is deliberate: it
+belongs to one lane's current state, shows up in a workflow diff, and carries no
+dashboard meaning.
+
+Usage::
+
+    compare_verdict_baselines.py <results-root>                  # gate + sweep
+    compare_verdict_baselines.py --vacuous-only <results-root>   # sweep only
+    compare_verdict_baselines.py --vacuous-only \
+        --require informational/consan-gemm \
+        --known-vacuous informational/consan-gemm <results-root>
+
+``--vacuous-only`` is for result trees that do not contain the gated cases at
+all -- notably the non-gating survey job, which runs in its own job with its own
+output root and is where the two #450 recipes actually live.
 """
 
 from __future__ import annotations
 
 import json
 import sys
+from collections.abc import Sequence
 from pathlib import Path
 
-from aorta.instrumentation.rocjitsu_sanitizers.models import SanitizerReport
+from aorta.instrumentation.rocjitsu_sanitizers.models import (
+    ExecutionSummary,
+    SanitizerReport,
+    Verdict,
+)
 from aorta.instrumentation.rocjitsu_sanitizers.report import read_report
 
 _BASELINES = Path("recipes/sanitizers/fixtures/expected/verdict_baselines.json")
@@ -24,12 +96,107 @@ _CASES = {
     "consan_clean": Path("consan-clean") / "sanitizer_report.json",
     "consan_racy": Path("consan-racy") / "sanitizer_report.json",
 }
+_GATED_BY_PATH = {relative: name for name, relative in _CASES.items()}
+
+# Execution summaries that carry no signal. ERROR is the #450 shape (the run died
+# before concluding anything); NOT_CHECKED means the sanitizer never ran, which is
+# equally uninformative when nothing declared it acceptable. COMPLETE and PARTIAL
+# both mean the sanitizer reached a conclusion, so a finding-free run under either
+# is a real clean result -- daily-consan-lds-dispatch passes with zero findings and
+# must not trip this.
+_VACUOUS_STATUSES = frozenset({ExecutionSummary.ERROR.value, ExecutionSummary.NOT_CHECKED.value})
+
+# The verdicts those statuses can carry. Spelled out for the same reason as the
+# statuses: a declaration is only protection if it could actually match, so
+# `overall_verdict: "pass"` is a typo to reject, not a rule to file away.
+_VACUOUS_VERDICTS = frozenset({Verdict.ERROR.value, Verdict.NOT_CHECKED.value})
+
+# Every field an `expected_error` must carry, and no others. Spelled as an exact
+# set rather than a minimum so that a typo -- `reason` for `reasons`, say -- is a
+# loud failure instead of a field nobody compares. An exemption that is not
+# checkable is indistinguishable from no exemption at all.
+_EXPECTED_ERROR_FIELDS = frozenset({"execution_status", "overall_verdict", "reasons", "why"})
+
+
+def _expected_error_problems(name: str, declared: object) -> list[str]:
+    """Reject an ``expected_error`` that does not describe exactly one outcome.
+
+    Checked at load time for every declared case, run or not: a malformed
+    declaration is a defect in the committed baselines, and waiting for the case
+    to run before saying so would leave the exemption unverified on every run
+    where the recipe failed to produce a report at all.
+    """
+    where = f"{name}.expected_error"
+    if not isinstance(declared, dict):
+        return [f"{where} must be an object, got {type(declared).__name__}"]
+
+    problems = []
+    missing = sorted(_EXPECTED_ERROR_FIELDS - declared.keys())
+    unknown = sorted(declared.keys() - _EXPECTED_ERROR_FIELDS)
+    if missing:
+        problems.append(f"{where} is missing {', '.join(missing)}")
+    if unknown:
+        problems.append(f"{where} has unknown field(s) {', '.join(unknown)}")
+
+    status = declared.get("execution_status")
+    if "execution_status" in declared and status not in _VACUOUS_STATUSES:
+        # Only a vacuous status can ever reach this exemption, so declaring any
+        # other one is dead configuration that reads as protection.
+        problems.append(
+            f"{where}.execution_status={status!r} is not one of "
+            f"{sorted(_VACUOUS_STATUSES)}, so it could never match"
+        )
+    verdict = declared.get("overall_verdict")
+    if "overall_verdict" in declared and verdict not in _VACUOUS_VERDICTS:
+        problems.append(
+            f"{where}.overall_verdict={verdict!r} is not one of "
+            f"{sorted(_VACUOUS_VERDICTS)}, so it could never match"
+        )
+    why = declared.get("why")
+    if "why" in declared and (not isinstance(why, str) or not why.strip()):
+        problems.append(f"{where}.why must be a non-empty string explaining the sign-off")
+
+    reasons = declared.get("reasons")
+    if "reasons" in declared:
+        if (
+            not isinstance(reasons, list)
+            or not reasons
+            or not all(isinstance(reason, str) and reason.strip() for reason in reasons)
+        ):
+            problems.append(
+                f"{where}.reasons must be a non-empty list of non-empty strings, got {reasons!r}"
+            )
+        elif len(set(reasons)) != len(reasons):
+            problems.append(f"{where}.reasons contains duplicates: {reasons!r}")
+    return problems
 
 
 def _load_baselines(path: Path) -> dict[str, object]:
     data = json.loads(path.read_text(encoding="utf-8"))
     if not isinstance(data, dict) or not data:
         raise ValueError(f"baselines file is empty or malformed: {path}")
+    problems = [
+        problem
+        for name, expected in data.items()
+        if isinstance(expected, dict) and "expected_error" in expected
+        for problem in _expected_error_problems(name, expected["expected_error"])
+    ]
+    # A gated key may not also carry an exemption, because the two mechanisms
+    # address different reports under the same name. `_case_key` maps the survey
+    # directory `informational/waitcheck-gemm` to `waitcheck_gemm`, which is also
+    # the gated key for `waitcheck/sanitizer_report.json` -- two different
+    # recipes. An `expected_error` added for either one would silently excuse the
+    # other, which is the cross-case bypass this exactness work exists to
+    # prevent. Declaring one of them means renaming so the two stop colliding.
+    problems.extend(
+        f"{name}.expected_error is not allowed on a gated case: {name} is also the "
+        f"baseline key for a survey case directory, so the exemption could not be "
+        f"attributed to one report or the other"
+        for name in sorted(data.keys() & _CASES.keys())
+        if isinstance(data[name], dict) and "expected_error" in data[name]
+    )
+    if problems:
+        raise ValueError("; ".join(problems))
     return data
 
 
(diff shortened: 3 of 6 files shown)


amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

The policy change is proven only on the much smaller ROCm 7.0.2 object, while the nightly is pinned to ROCm 10 and this recipe already documents that its actual object exceeds the current growth/runtime bounds. The new sweep also fails open when recipe startup failures leave reports absent. I independently confirmed both existing Copilot findings.

**Findings:** 1 High · 2 Medium · 1 Low

**Must fix before merge**
- **High:** Validate a non-vacuous GEMM result in the digest-pinned ROCm 10 nightly environment; the checked-in recipe currently says that input cannot complete within its bounds.

**Still open from Copilot review**
- **Medium:** [`expected_error` is a permissive bypass](https://github.com/ROCm/aorta/pull/454#discussion_r3924815213), accepting malformed declarations, substring matches, and `not_checked` outcomes.
- **Low:** [The dashboard description still claims strict-policy behavior](https://github.com/ROCm/aorta/pull/454#discussion_r3924815284) after the recipe switches to lenient.

recipes/sanitizers/daily-consan-gemm.yaml
```diff
@@ -30,7 +30,27 @@ sanitizer_plan:
...
+    # ConSan reads, patches and analyzes a real production Tensile object. Measured on
+    # gfx950 / ROCm 7.0.2.2: the object transforms clean (outcome=modified-valid,
+    # 75978 patches over 68894 discovered access sites) and only then did strict
+    # reject it, so the coverage signal was there all along and only the policy
+    # discarded it.
+    #
+    # Dynamic race evidence over a dispatching GEMM is genuinely out of scope here and
+    # is tracked separately; daily-consan-lds-dispatch.yaml is the end-to-end control
+    # that proves the record/replay path works on a caller-supplied object.
+    consan_policy: lenient
```
amd-vivekag, line 111: **[High]** This is not validated against the environment that runs the nightly.

The same recipe documents that ROCm 10's object has about 9.6x more sites, currently exceeds the patched-image growth limit, and cannot plausibly complete inside 6000s; the successful lenient result used ROCm 7.0.2's much smaller object.

Fix: Run this against the digest-pinned ROCm 10 image and use a bounded object or resolve the growth/runtime limits so it produces a non-vacuous report before enabling the new sweep.

scripts/sanitizers/compare_verdict_baselines.py
```diff
@@ -70,17 +109,112 @@ def _compare_case(name: str, report: SanitizerReport, expected: dict[str, object
...
+
+    Judges only the reports that are present. Absent reports are deliberately not
+    an error here: the survey job is non-gating and may legitimately be killed by
+    the runner cap tracked in #370, and the gated cases already have their own
+    presence check in the comparison above.
+    """
+    failed = False
+    reports = sorted(root.rglob("sanitizer_report.json"))
+    if not reports:
+        print(f"vacuity sweep: no reports found under {root}")
```
amd-vivekag, line 421: **[Medium]** Missing reports fail open, including an entirely empty survey.

All six workflow invocations use `|| true`, so a startup, recipe, or input failure that writes no report is swallowed; a runner-level kill does not justify this path because it prevents the comparator from running at all.

Fix: Require the six invoked survey report paths and fail when any are absent; job cancellation can remain tolerated by never reaching this comparator.

recipes/sanitizers/daily-consan-gemm.yaml
vivekkhandelwal1: You were right, and the measurement changed the shape of this PR. Thank you — this was not a hypothetical.

## Not hypothetical, and not speculation about a future break

[#453](https://github.com/ROCm/aorta/issues/453) exists precisely because the #411 image flip changed these GEMM code objects and broke a *different* sanitizer expectation on this same lane. That lane has now failed on **five consecutive scheduled nights**: [33629901482](https://github.com/ROCm/aorta/actions/runs/33629901482) (Sep 2), [33755195031](https://github.com/ROCm/aorta/actions/runs/33755195031) (Sep 3), [33872660689](https://github.com/ROCm/aorta/actions/runs/33872660689) (Sep 4), [33965800908](https://github.com/ROCm/aorta/actions/runs/33965800908) (Sep 5), [34032862308](https://github.com/ROCm/aorta/actions/runs/34032862308) (Sep 6), against a last-green of [33507543444](https://github.com/ROCm/aorta/actions/runs/33507543444) on Sep 1. So you were generalising from a failure that is currently ongoing, not predicting one.

## Validated on ROCm 10, and the answer is that `lenient` does not fix this case

Run inside the exact digest `docker/Dockerfile.ci-gpu` pins — `rocm/pytorch:rocm10.0_ubuntu26.04_py3.14_pytorch_release_2.13.0@sha256:3174cb70…` — on Slurm `meta64` node `cv350-rck-g03-e15-18`, gfx950 / MI350X (`0x75a0`), with the same devel-symlink layer that Dockerfile adds and the nightly's own fixture-build steps. Node-local scratch. Deliberately **not** `smci350-rck-g03-f16-12`, which is reserved for the performance baseline window.

The object came out byte-identical to the one the nightly extracts, which is what makes this the CI environment rather than an approximation of it:

```
consan_gemm_f32.hsaco  169996560 bytes
sha256 57c5d8efa448315ddc5f49170757ebeec1eb397468676ff158d19c71572b1c55
   (run 34032862308 recorded the same sha256, and input_image=169996560)
```

| `daily-consan-gemm`, ROCm 10 | verdict | exec | findings | reason | coverage |
|---|---|---|---|---|---|
| **strict** (before) | `error` | `error` | 0 | `consan_strict_load_rejection` (exit 92) | — |
| **lenient** (this PR) | `error` | `error` | 0 | `consan_coverage_incomplete` | **access=0/719550** |

`gemm-strict` reproduces the nightly's reason exactly, which is the control that establishes the environment matches.

**So `lenient` changes the failure mode and not the outcome.** Both fail on the same thing, and it is not the record requirement:

```
ConSan transform admission request input_image=169996560 growth_policy=absolute-bytes growth_value=419430400
ConSan patch end   visited=true modified=false outcome=invalid errors=1 patches=0
ConSan MOI first-light probe rejected patched-image file growth:
  required total 1721958400 bytes, limit 419430400 bytes (policy absolute-bytes=419430400)
```

1,721,958,400 B against 419,430,400 B — **4.11x over**, `patches=0`, the object never instrumented.

## Two corrections to your comment, both narrowing rather than disputing it

**The growth limit: confirmed.** Your reading was right, and the recipe's own note was right about the mechanism.

**The runtime: not what bites, and I could not confirm it.** The run ends in **595s under strict and 859s under lenient**, both far inside `timeout_seconds: 6000`. The growth ceiling fires first, so the run never does the work that a runtime estimate would be about. "Cannot plausibly complete inside 6000s" is a reasonable *prediction* for the post-ceiling world — the recipe now says raising the ceiling most likely trades this rejection for `combined_hook_timeout`, since the survey job itself caps at `timeout-minutes: 120` — but it is not the current failure and I have not measured it.

**The sites are ~10.4x, not 9.6x.** The 9.6x in the recipe was derived from memory-instruction counts read off the ELF. Measured live, the ROCm 10 object yields **719,550 access sites and 109,497 barrier sites** against 68,894 access sites on the 15.5 MB object. Recipe corrected.

## A finding I did not expect, and it makes `lenient` more delicate here, not less

On this object `lenient` moves the fail-closed out of the sanitizer and into AORTA.

- Under **strict** the hook terminates itself: `fail_closed=true`, `action=terminate`, `exit_code=92`.
- Under **lenient** it declines to — `fail_closed=false` — **exits 0**, and hands back an object with none of its 719,550 access sites patched. Only AORTA's coverage cross-check turns that into an `error`.

Isolated cleanly by forcing the ceiling below what the small LDS fixture needs, so the object size is not a variable:

```
RJ_CONSAN_POLICY=strict   -> exit 92, "load rejection ... action=terminate exit_code=92"
RJ_CONSAN_POLICY=default  -> exit 0,  same growth rejection, NO load-rejection line at all
```

That is the "quiet pass" hazard #450 worried about, and on this object it is real. The coverage cross-check is the only thing preventing it. Worth knowing before anyone raises the ceiling.

## What I changed as a result

**`lenient` is kept, but described as necessary rather than sufficient.** It is still the correct policy for a load-only driver and a precondition for this case ever passing: the moment the ceiling admits the object, strict fails it again on require-records, which is the #450 defect. What was wrong was the *claim*, so the recipe now states which object each result was measured on, that the ROCm 10 object is blocked earlier and by what, and that the coverage cross-check is load-bearing under lenient. The pre-ROCm-10 figures in the timeout block are corrected too.

**The survey wiring is dropped from this PR** — your "before enabling the new sweep", taken at face value. Replayed over the real survey trees of all four ROCm 10 nightlies, the sweep flags **two** rows, not one:

| row | reason | owner |
|---|---|---|
| `consan_gemm` | `consan_strict_load_rejection`, 0 findings | the growth ceiling above |
| `waitcheck_gemm` | `worklist_not_fully_checked`, 0 findings | #453, on that same object |

A check that fails every night cannot distinguish a new break from the standing two. So the invocation is left out with a comment recording exactly what has to be true to turn it on. **The gate half stays live**: replayed over 33507543444 (last green) and 34032862308 it adds no failure there, so it is protection without noise.

**`Fixes #450` → `Refs #450`.** The misconfiguration #450 reported is fixed and the mechanism that hid it is closed, but the row it was filed about still produces no signal on the image CI runs. Auto-closing would assert coverage that does not exist.

## On "use a bounded object"

I did not do this, and I want to be explicit that it is a live option rather than one I dismissed. `docs/sanitizers/consan-gemm-patched-image-growth-cap.md` measured a 13.6 MiB per-chip sibling (`_ID75a3-75a2`, same family and NT layout, plain f32) that **does** clear the ceiling and instruments cleanly in 87s — and it is available only on ROCm ≥ 7.14, so **the move to ROCm 10 has just unblocked it**. Reaching it needs no code change, only `prepare_gemm_isa.py --chip-id 0x75a3`.

I left it out because it changes what this case represents — an MI355X-class object scanned on an MI350X gate — and that is a call about the fixture's meaning that deserves its own review and its own re-bless, not a paragraph in a policy fix. Happy to do it here instead if you would rather; it is the one change that would make this row non-vacuous today.

## Also worth surfacing rather than burying

The same ROCm 10 run reported, on the `waitcheck_preflight` check:

```
waitcheck CFG dataflow did not converge at .text+0x24d7070
(in:loadcnt,uncertain-order;out:loadcnt,uncertain-order)
```

The **same offset** #453 records, reproduced on a Slurm dev node rather than the CI runner. That independently supports #453's diagnosis that this is content-driven rather than runner-specific, and shows it reaches the survey lane as well as the gate. Noted in the PR body; I have not edited #453 itself.

New head `6320959`.

scripts/sanitizers/compare_verdict_baselines.py
```diff
@@ -70,17 +109,112 @@ def _compare_case(name: str, report: SanitizerReport, expected: dict[str, object
...
+            f"declare it as an 'expected_error' entry in {_BASELINES}."
+        ]
+
+    want = declared.get("reason")
+    if isinstance(want, str) and not any(want in reason for reason in reasons):
+        return [
+            f"{shape} -- declared expected_error reason {want!r} is absent, so this is "
+            f"a different failure than the one that was signed off."
+        ]
+    return []
```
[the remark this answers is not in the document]
vivekkhandelwal1: Correct on all four counts, and the gap between what the docstring claimed and what the code did was the real problem. Fixed by making the declaration exact and by failing on a malformed one rather than skipping it.

**The match is now exact on three fields, not a substring on one.** `_vacuous_problems` compares `execution_status`, `overall_verdict` and the **complete set** of check reasons:

```python
if reasons != sorted(declared["reasons"]):
    mismatches.append(f"reasons={reasons!r}, declared {sorted(declared['reasons'])!r}")
```

Taking your four cases in turn:

| your case | now |
|---|---|
| missing / non-string `reason` skips the condition | `reasons` is required and must be a non-empty list of non-empty strings; absence is a load-time failure |
| `combined_hook_exit_860` passes on substring | set equality — rejected, with a test naming that exact string |
| additional unrelated reasons ignored | the full reason set must match, so a second failure riding along is rejected |
| `not_checked` accepted if the reason matches | `execution_status` and `overall_verdict` are compared too |

**A malformed declaration is a failure, not a bypass.** This is the part I think mattered most in your comment, because it is the one that fails *open*. Every `expected_error` is validated when the baselines file is loaded, whether or not that case ran:

```python
_EXPECTED_ERROR_FIELDS = frozenset({"execution_status", "overall_verdict", "reasons", "why"})
```

Spelled as an exact set rather than a minimum, so `reason` for `reasons` — precisely the old spelling — is now a loud error listing both the unknown field and the missing one, instead of a field nobody compares. Validating at load time rather than at match time is deliberate: waiting for the case to run would leave the exemption unverified on exactly the runs where the recipe produced no report at all.

A declaration that could never match is also rejected, since dead configuration that reads as protection is worse than none: `execution_status` must be one of the vacuous statuses the sweep actually looks at, and `overall_verdict` likewise. Declaring `complete` or `pass` there is a typo, not a rule.

**One further hole this surfaced, which is your point one level up.** `waitcheck_gemm` is **both** a gated baseline key *and* what the survey directory `informational/waitcheck-gemm` resolves to through `_case_key` — two different recipes (`daily-waitcheck-gemm.yaml` and `daily-waitcheck-gemm-object.yaml`). A single `expected_error` under that name would have excused both reports. That is the same class of defect you described — a mechanism meant to name one specific expected failure quietly covering a class of them — so a gated key may no longer carry an `expected_error` at all, and declaring one of those rows now requires renaming so the two stop colliding.

**Why the sharpness matters here specifically.** This exemption exists to protect one row: `daily-consan-tiny`'s intended negative control. A loose version defeats its own purpose, since the thing it is protecting is an assertion that the guardrail fails *closed*. Its declaration is now verified against the environment that runs the nightly rather than a dev node — the real ROCm 10 survey trees of runs 33755195031, 33872660689, 33965800908 and 34032862308 all report exactly `error`/`error` with `combined_hook_exit_86` on both checks, and a direct re-run inside the digest-pinned ROCm 10 CI image reproduced it. Under `lenient` that same case reports `consan_coverage_incomplete` at `access=0/0` instead, and the declaration deliberately does **not** admit that.

11 tests cover this directly, including the old `reason` spelling, a non-object declaration, unusable reason lists (bare string, empty, `[""]`, `[86]`, duplicates), the never-matchable status and verdict, an empty `why`, `combined_hook_exit_860`, an extra undeclared reason, the declared reason on a `not_checked` run, the gated-key collision, and the real shape where one reason legitimately repeats across both checks. Plus `test_committed_baselines_declarations_are_well_formed`, since the validation only bites if the shipped file satisfies it.

New head `6320959`.

scripts/sanitizers/compare_verdict_baselines.py
vivekkhandelwal1: Agreed, and fixed — but not by requiring the six paths unconditionally, because the distinction you drew in your own "Fix" is the one that decides the shape. Reasoning below, since you asked me to decide carefully.

**Your diagnosis is right.** All six invocations run under `|| true`, so a recipe that dies *before* writing a report leaves nothing behind, and a report that does not exist cannot be judged vacuous. That is the same fail-open shape the sweep exists to close, one level up.

**Your parenthetical is also right, and it is what resolves the #370 objection**: "job cancellation can remain tolerated by never reaching this comparator." That is exactly it. A job-level kill — the runner cap in #370, a cancellation, the step's `timeout-minutes` ceiling — takes the whole `bash -lc` payload down with it. The comparator is the last line in that payload, so it never runs and has no verdict to give. There was never anything to tolerate; the earlier "absent reports are not this check's business" note conflated a *job-level* kill with a *recipe-level* failure, and only the second one can actually reach the comparator. Nothing has to distinguish "capped" from "produced nothing", because the capped case cannot get here.

**What I did differently from a blanket requirement**, and why:

```
--require informational/consan-gemm --require informational/consan-tiny ...
```

The caller names what it invoked; a required case with no report is an error. I did not make the sweep fail on *any* empty tree, because an empty tree and a tree from a run that was never asked to produce anything are indistinguishable from a directory listing — the sweep genuinely cannot infer the six. Making absence an error only for cases the caller declares keeps the check honest in both directions, and it means the same code path serves the gate job (which has its own three cases and its own presence check) without special-casing.

The practical difference from hardcoding the six paths into the comparator is that the list lives next to the invocations it describes, so adding a seventh survey recipe cannot silently escape the presence check by being added in one place and not the other.

**One thing to flag: the survey invocation is not in this PR.** `--require` and the sweep are implemented and tested, but wiring them into `sanitizers-nightly.yml` is left out, for the reason you gave on the other thread — on today's ROCm 10 objects the sweep would be red on arrival on two rows (`consan_gemm` on the patched-image growth ceiling, `waitcheck_gemm` on #453), neither fixed here. So the fail-open you identified is *closed in the mechanism* but not yet armed in that lane. I would rather say that plainly than claim the hole is shut. The comment at the end of the survey payload records what has to be true to arm it, and the gate job's sweep — which is clean on both the last green run and the current one — stays live.

If you would rather have the survey invocation land now and accept a red non-gating job as the accurate report, say so and I will put it back; it is one line plus the six arguments. My read is that a check which fails every night stops being read, but that is a judgement call and it is yours as much as mine.

5 tests cover this: a required case with no report (asserting the *other* required case is not also reported), an entirely empty survey, all-required-present, `--require` in gate mode, and the argument-parsing errors (`--require` with no value, `--require` swallowing the positional).

New head `6320959`.

recipes/sanitizers/daily-consan-gemm.yaml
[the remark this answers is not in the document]
vivekkhandelwal1: Right, and the dashboard point is what makes it more than a comment nit — that description is rendered on Tab 2, so it was actively telling readers the case asserts something it no longer does. Rewritten.

The sentence you flagged said strict policy "must reach a real verdict or fail closed, never a false pass". Under `lenient` that is no longer the claim, and the honest replacement had to say what the case *does* assert:

> `consan_gemm_load` only LOADS that object — it never dispatches — so under `consan_policy: lenient` what this case asserts is static instrumentation coverage on a real production object: that ConSan reads it, patches it, and reports how many of its discovered access sites it actually covered. It asserts nothing about dynamic race evidence; `daily-consan-lds-dispatch.yaml` is the end-to-end control for that.
>
> A pass here therefore means "instrumented, and the coverage cross-check agrees", never merely "ran without complaining" — under lenient ConSan itself no longer fails closed, so the coverage cross-check is what stands between an uninstrumented object and a false pass.

That second paragraph is not padding. Measuring this on ROCm 10 for @amd-vivekag's review turned up that `lenient` sets `fail_closed=false`, so on an object ConSan declines to instrument the hook now **exits 0** rather than terminating — and only AORTA's coverage cross-check turns that into an error. The old description's "never a false pass" was carried by strict; a reader of the new one needs to know what carries it now.

I also added a status paragraph, because a description that says what the case asserts while the row is red every night is its own kind of stale:

> **STATUS ON THE CURRENT CI IMAGE:** this case does NOT reach that assertion today. On the digest-pinned ROCm 10 base the object is 169,996,560 bytes and ConSan rejects it on the patched-image growth ceiling before instrumenting anything, so the row is `error` with zero findings every night. That is a capacity limit, not something `consan_policy` can change.

The old `known rocjitsu limitations ... consan-4112-overlapping-anchor-patches.md` sentence is gone, replaced by a pointer to `consan-gemm-patched-image-growth-cap.md` — the document that actually describes the limit this case hits. The anchor-overlap defect it used to cite is fixed; both surface as `status=4112`, which is the single most misleading thing about this failure.

While in the file I also corrected the timeout block, which still reasoned from the ROCm 7.2.4 base: 169,996,560 bytes and 719,550 access sites rather than 191,935,808 and 637,823, the measured 1,721,958,400-byte growth requirement, and the fact that the 6000s ceiling is not what limits this case (the run ends in 595s under strict, 859s under lenient).

New head `6320959`.

.github/workflows/sanitizers-nightly.yml
```diff
@@ -580,6 +580,44 @@ jobs:
               --output-dir "${SANITIZER_OUT}/informational/waitcheck-lds-dispatch" || true
             aorta sweep run --recipe recipes/sanitizers/daily-waitcheck-tiny.yaml \
               --output-dir "${SANITIZER_OUT}/informational/waitcheck-tiny" || true
+
+            # NOT YET WIRED, deliberately: the vacuity sweep this branch adds to
+            # scripts/sanitizers/compare_verdict_baselines.py is NOT invoked here.
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fair, and you changed my mind. Wired, in `33f8b88`.

You were right about the substance: with the invocation left out, the exact #450 regression would still pass every nightly, and the gate job's separate three-case root does not cover the six informational rows. That was the whole point of the change, so leaving it unarmed made it a mechanism rather than a fix.

(On the contradiction you noted — the PR body had already been rewritten to say the wiring was excluded, so the two were consistent, just consistently timid. The substantive criticism stands regardless.)

**Where I had got stuck, and how your "or explicitly modeling" resolved it.** Replayed over the real ROCm 10 survey trees of the last four nightlies, the sweep flags **two** rows, neither fixable here:

| row | reason | owner |
|---|---|---|
| `consan-gemm` | `consan_strict_load_rejection`, 0 findings | ConSan needs 1,721,958,400 B of patched image against a 419,430,400 B ceiling — 4.11x over — and never instruments the object |
| `waitcheck-gemm` | `worklist_not_fully_checked`, 0 findings | the `rj_waitcheck` CFG dataflow non-convergence of #453, on that same object |

Arming it unqualified meant red every night, which is a check that cannot tell a new break from the standing two. My reading was that this made it worse than not arming it. The modelling option is the way out, so:

```
python scripts/sanitizers/compare_verdict_baselines.py \
  --vacuous-only ${REQUIRE} \
  --known-vacuous informational/consan-gemm \
  --known-vacuous informational/waitcheck-gemm \
  "${SANITIZER_OUT}"
```

`--known-vacuous` is deliberately the **opposite** of an `expected_error`, because the difference is the whole reason this is safe:

- an `expected_error` asserts *this outcome is correct*. `--known-vacuous` asserts *this outcome is wrong, it is tracked elsewhere, and it must not mask the other rows*;
- it lives at the **call site**, not in `verdict_baselines.json`, so it carries no dashboard meaning and shows up in a workflow diff rather than in a fixture;
- **it is self-retiring.** The comparator *fails* if a listed case starts producing signal — `Remove informational/consan-gemm from the caller's --known-vacuous list -- a suppression that is no longer true hides the next regression` — and fails if a listed case has no report at all. So the list cannot outlive the break it describes, which is the failure mode any suppression list normally has;
- a case may not be both. Passing `--known-vacuous` for a row declared as an `expected_error` is a usage error (exit 2), since the two make opposite claims about the same row;
- the suppressed message deliberately does **not** repeat the sweep's "declare it as an `expected_error`" advice, which would be precisely the anti-pattern #453 warns against. There is a test asserting that string never appears for a known-vacuous row.

**So the four healthy rows are checked for real from the next scheduled run**, which is the regression path you were pointing at. Verified by replaying the exact wired invocation over the four real survey trees — exit 0, both blocked rows named, six reports judged:

```
consan_gemm: KNOWN VACUOUS, not failing -- execution_status='error', overall_verdict='error',
  reason 'consan_strict_load_rejection', and zero findings. This row is on the caller's
  --known-vacuous list: the outcome is wrong and tracked elsewhere, not accepted.
waitcheck_gemm: KNOWN VACUOUS, not failing -- ... reason 'worklist_not_fully_checked' ...
vacuity sweep: ok (6 report(s) carry signal or are declared) (2 known-vacuous)
```

And the gate root is unaffected: over runs 33507543444 (last green) and 34032862308 the sweep adds no failure there.

One note on scope, since #453 independently asks for a "visibly not coverage" state in the baselines for cases like these. `--known-vacuous` is intentionally narrower than that — one lane's current state, at the call site, no dashboard semantics — so it closes this path now without pre-empting that design. If #453 lands the richer state, these two arguments are what it replaces.

5 new tests: suppressed-and-not-failing, suppression not masking a second regression, failing once a suppressed row recovers, failing when a suppressed row has no report, and the expected_error conflict.

recipes/sanitizers/daily-consan-gemm.yaml
vivekkhandelwal1: **Correction to my reply above: the survey sweep IS now wired, as of `33f8b88`.**

I had left the invocation out on the reading of your "before enabling the new sweep". [@Copilot pushed back](https://github.com/ROCm/aorta/pull/454#discussion_r3947532191) that this leaves the exact #450 regression passing every nightly — the gate job's comparator only covers its own three-case root — and suggested arming it "after resolving or explicitly modeling the two standing vacuous outcomes". That second option is the one I had not thought of, and it is better than either of the choices I offered you.

So the two blocked rows are now modelled at the call site rather than the sweep being left off:

```
--known-vacuous informational/consan-gemm
--known-vacuous informational/waitcheck-gemm
```

This is the opposite of an `expected_error`, and the distinction is what makes it safe: an `expected_error` says the outcome is *correct*, whereas `--known-vacuous` says it is *wrong and tracked elsewhere*. It sits in the workflow rather than in `verdict_baselines.json`, so it carries no dashboard meaning and shows up in a workflow diff. Crucially it is **self-retiring** — the comparator *fails* if a listed row starts producing signal, or stops reporting at all — so it cannot outlive the break it describes. And a row may not be both known-vacuous and declared; that is a usage error.

The practical effect is the one you asked for: all six survey reports are required to exist, the four healthy rows are judged for real from the next scheduled run, and the two blocked ones are named loudly in the log without failing the job. Replaying the exact wired invocation over the four real ROCm 10 survey trees gives exit 0 with both rows reported.

Nothing else in my reply changes — the growth-ceiling measurement, the `fail_closed` finding, and `Refs #450` all stand. I have left this thread unresolved because the bounded-object question at the end of my reply is still yours to call.

scripts/sanitizers/compare_verdict_baselines.py
vivekkhandelwal1: **Answering my own question from the reply above: the survey invocation is now wired, as of `33f8b88`** — so the fail-open you identified is closed in the lane as well as in the mechanism.

I had asked whether you would rather have it armed and accept a red non-gating job. [@Copilot's push-back](https://github.com/ROCm/aorta/pull/454#discussion_r3947532191) supplied a third answer that is better than either: arm it, and *explicitly model* the two rows that are already vacuous, rather than choosing between a red job and an unarmed check.

```
python scripts/sanitizers/compare_verdict_baselines.py \
  --vacuous-only ${REQUIRE} \
  --known-vacuous informational/consan-gemm \
  --known-vacuous informational/waitcheck-gemm \
  "${SANITIZER_OUT}"
```

`${REQUIRE}` is the six `--require informational/<case>` arguments built from the same two case lists the invocations iterate, so a seventh survey recipe cannot escape the presence check by being added in one place and not the other — which was the point of your original comment.

`--known-vacuous` is deliberately not an `expected_error`: it asserts the outcome is *wrong and tracked elsewhere*, not that it is acceptable. It is **self-retiring** — the comparator fails if a listed row starts producing signal or stops reporting — so the list has to shrink to nothing rather than quietly persisting, and a row may not be both known-vacuous and declared.

Verified by replaying the exact wired invocation over the four real ROCm 10 survey trees: exit 0, all six reports required and judged, both blocked rows named in the log. So an entirely empty survey now fails here, a recipe that dies before writing a report now fails here, and the four healthy rows are checked for real from the next scheduled run.

vivekkhandelwal1:
All four review comments are answered in-thread, and the head is now `33f8b88`. Summary of what changed, because the High finding was correct and it moved the PR's central claim.

**@amd-vivekag's High finding was right, and the answer is that `lenient` does not fix this case on the image CI runs.** I ran `daily-consan-gemm` inside the digest-pinned ROCm 10 base (`sha256:3174cb70…`) on a gfx950 / MI350X Slurm node, against the byte-identical object the nightly extracts (sha256 `57c5d8ef…`, 169,996,560 bytes, matching run [34032862308](https://github.com/ROCm/aorta/actions/runs/34032862308)):

| | verdict | exec | findings | reason | coverage |
|---|---|---|---|---|---|
| strict | `error` | `error` | 0 | `consan_strict_load_rejection` (exit 92) | — |
| lenient | `error` | `error` | 0 | `consan_coverage_incomplete` | access=**0/719550** |

Both fail on the patched-image growth ceiling — 1,721,958,400 B needed against 419,430,400 B, 4.11x over, `patches=0` — not on the record requirement. Two corrections that narrow the finding rather than dispute it: the **6000 s ceiling is not what bites** (595 s strict, 859 s lenient), and the site count is **10.4x**, not 9.6x.

**An unexpected finding worth reading:** on this object `lenient` sets `fail_closed=false`, so ConSan **exits 0** and returns an object with none of its 719,550 access sites patched. Only AORTA's coverage cross-check turns that into an error. Isolated by forcing the ceiling below what the small LDS fixture needs — `RJ_CONSAN_POLICY=strict` exits 92 with a load-rejection line, `default` exits 0 with none. That is the quiet-pass hazard #450 worried about, and it is real here.

`lenient` is kept as a **precondition** rather than a fix, and the recipe now says which object each result came from. **`Fixes #450` → `Refs #450`**: the misconfiguration is fixed, but the row is still vacuous on the image CI runs, so auto-closing would assert coverage that does not exist.

**@Copilot's two findings were both correct and are fixed**: `expected_error` is now an exact match on execution status, verdict and the complete reason set, with malformed declarations failing at load time rather than being skipped; and the dashboard-facing description no longer claims a strict-policy assertion. A third hole surfaced while making it exact — `waitcheck_gemm` is both a gated key and what `informational/waitcheck-gemm` resolves to, so one exemption there would have covered two recipes. Gated keys may no longer carry one.

**@Copilot's follow-up on this push was also right, and it produced a better answer than the one I had.** I had left the survey invocation out, reading @amd-vivekag's "before enabling the new sweep" literally. That leaves the exact #450 regression passing every nightly. The suggestion to arm it *after explicitly modelling* the standing vacuous outcomes is the option I had missed, so the sweep is now wired with `--known-vacuous informational/consan-gemm` and `--known-vacuous informational/waitcheck-gemm`. That flag is the deliberate opposite of an `expected_error` — it asserts the outcome is wrong and tracked elsewhere, lives at the call site with no dashboard meaning, and **fails when a listed row starts producing signal or stops reporting**, so it cannot outlive the break. The other four rows are checked for real from the next scheduled run.

**One thread left open on purpose.** [The High thread](https://github.com/ROCm/aorta/pull/454#discussion_r3925899483) is unresolved because its underlying condition is not fixed: `consan-gemm` still produces no signal in CI, and there is a real choice for @amd-vivekag to make. `docs/sanitizers/consan-gemm-patched-image-growth-cap.md` measured a 13.6 MiB per-chip sibling object (`_ID75a3-75a2`, same family and NT layout, plain f32) that **does** clear the ceiling and instruments cleanly in 87 s — available only on ROCm ≥ 7.14, so **the move to ROCm 10 has just unblocked it**, reachable with `prepare_gemm_isa.py --chip-id 0x75a3` and no code change. I left it out because it changes what this case represents (an MI355X-class object on an MI350X gate), which deserves its own review and re-bless. Happy to do it here instead if you would prefer; it is the one change that would make this row non-vacuous today.

**A finding that belongs to #453, surfaced rather than buried.** The same ROCm 10 run reported `waitcheck CFG dataflow did not converge at .text+0x24d7070` — the **same offset** #453 records — reproduced on a Slurm dev node rather than the CI runner, and in the survey lane as well as the gate. That supports #453's diagnosis that the failure is content-driven, not runner-specific. I have not edited #453.

Also worth stating plainly for the record: this lane has now failed on **five consecutive scheduled nights**, Sep 2 through Sep 6, against a last green of Sep 1 — so the review was generalising from an ongoing failure, not a hypothetical.

Rebased onto current `main` (`4b4553e`); no file overlap with the four commits that landed since. 395 tests in `tests/sanitizers` (26 new), 135 in `tests/instrumentation/rocjitsu_sanitizers`, `pre-commit run --all-files` clean. Hardware work was on Slurm node `cv350-rck-g03-e15-18` with node-local scratch, deliberately **not** the CI runner `smci350-rck-g03-f16-12`.

scripts/sanitizers/compare_verdict_baselines.py
```diff
@@ -106,6 +503,9 @@ def main(argv: list[str]) -> int:
                 print(f"{name}: {problem}")
         else:
             print(f"{name}: ok ({report.overall_verdict.value})")
+
+    if _sweep_vacuous(root, baselines, required):
```
amd-vivekag: **[Medium]** `--known-vacuous` is parsed, then dropped on the gate+sweep path.

Conflict-checking runs for every invocation, but this call only forwards `required`, so `compare_verdict_baselines.py --known-vacuous informational/consan-gemm <root>` still fails that row as an ordinary vacuous report. Copilot flagged this too. The nightly is fine today because it uses `--vacuous-only`.

Fix: forward the list (and add a gate-mode test next to `test_require_also_applies_to_the_gate_mode`).

```suggestion
    if _sweep_vacuous(root, baselines, required, known_vacuous):
```

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

Re-review of `33f8b88`. The ROCm 10 measurement, exact `expected_error` matching, `--require`, and the wired survey sweep with self-retiring `--known-vacuous` all land, and the PR no longer claims that `lenient` restores CI coverage. One hole remains in the new CLI: the gate+sweep path still drops `--known-vacuous`.

**Prior comments**
- **High** (ROCm 10 validation): resolved as a claim/honesty issue. CI gemm stays vacuous; that is accepted here via `--known-vacuous`. Keep #450 open. The `0x75a3` sibling object is a follow-up — it changes what the case represents.
- Copilot exact `expected_error`: resolved
- Copilot dashboard description: resolved
- **Medium** (missing reports fail open): resolved
- Copilot survey invocation: resolved

**Still open from last review / Copilot**
- **Medium:** `--known-vacuous` is parsed and conflict-checked, then ignored unless `--vacuous-only`. Copilot flagged this too.

The self-retiring `--known-vacuous` design is the right answer for arming the sweep without blessing a broken row.

recipes/sanitizers/daily-consan-gemm.yaml
amd-vivekag: Resolved for this PR's revised claim — thanks. The ROCm 10 numbers check out; `lenient` is a precondition, not a CI fix. Keep #450 open. Leave the 0x75a3 sibling object for a follow-up; it changes what the case represents.

scripts/sanitizers/compare_verdict_baselines.py
amd-vivekag: Resolved — `--require` plus the survey wiring closes the fail-open.

scripts/sanitizers/compare_verdict_baselines.py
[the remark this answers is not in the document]
amd-vivekag: Resolved — exact match on status, verdict, and the full reason set, with load-time validation.

recipes/sanitizers/daily-consan-gemm.yaml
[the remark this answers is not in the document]
amd-vivekag: Resolved — the dashboard description now describes the lenient coverage check.

.github/workflows/sanitizers-nightly.yml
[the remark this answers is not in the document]
amd-vivekag: Resolved — the survey job now invokes the sweep with `--require` and `--known-vacuous`.

scripts/sanitizers/compare_verdict_baselines.py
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed by forwarding, in `4314efc`.

You had it exactly right, including the part that makes it more than a plumbing slip: the list was parsed **and** conflict-checked and then dropped at this call site, so `--known-vacuous <case> <root>` without `--vacuous-only` still failed the declared row as an ordinary vacuous report — and did so while printing the "declare it as an `expected_error`" advice, which is the one thing a known-vacuous row must never be told to do. A caller who declared the row would have read that message as the check not knowing about the declaration, when in fact it had validated it moments earlier.

## Why forward rather than reject

Of the two resolutions you offered, forwarding is the one that follows the semantics already in the file rather than narrowing them:

- `--vacuous-only` selects which **halves** run, nothing more. It exists for result trees that do not contain the gated cases at all (the survey job's separate output root). It does not change what the sweep *means*, and the sweep is the same code over the same root either way.
- `--require` is the companion option and also describes the sweep. It is already honoured on both paths, with `test_require_also_applies_to_the_gate_mode` asserting exactly that. Rejecting `--known-vacuous` outside `--vacuous-only` would have left two options that describe the same pass disagreeing about where they are legal, for no reason a caller could predict.
- The `expected_error` conflict check already runs unconditionally, *ahead* of the mode branch. That is only coherent if the option is meaningful on every path — which is the reading you took, and the code agrees with you.

`_sweep_vacuous` already had the parameter, so the fix is the one-line forward at the call site. I also stated the invariant in the module docstring, next to the `--vacuous-only` note, since the whole defect was that nothing wrote it down.

## Verification

The nightly lane is unaffected — it passes `--vacuous-only`. Replayed over the real survey trees of runs 33755195031, 33872660689, 33965800908 and 34032862308: still exit 0, both blocked rows reported, six reports judged.

For the path that actually changed, I built a gate+sweep tree from the last-green gate run [33507543444](https://github.com/ROCm/aorta/actions/runs/33507543444) plus the real vacuous ROCm 10 `consan-gemm` report from [34032862308](https://github.com/ROCm/aorta/actions/runs/34032862308):

```
--known-vacuous informational/consan-gemm <root>     pre-fix: exit 1   post-fix: exit 0
  pre-fix   consan_gemm: ... zero findings -- the run produced no sanitizer signal at all.
            Either the recipe is misconfigured (see ROCm/aorta#450) or ... declare it as an
            'expected_error' entry ...
  post-fix  consan_gemm: KNOWN VACUOUS, not failing -- ... reason
            'consan_strict_load_rejection', and zero findings. This row is on the caller's
            --known-vacuous list: the outcome is wrong and tracked elsewhere, not accepted.
```

The gate half is untouched: with the same declared row present, blanking the expected finding shape on `consan-racy` still fails the run.

Two tests, in the known-vacuous section and named to mirror the `--require` one:

- `test_known_vacuous_also_applies_to_the_gate_mode` — asserts the row is not failed on the gate path, and that the `expected_error` advice does not appear for it. I checked this one fails against the pre-fix code (`assert 1 == 0`) rather than trusting that it covers the defect.
- `test_known_vacuous_conflict_is_caught_in_the_gate_mode_too` — forwarding the list must not open a path on which a row that is both `--known-vacuous` and a declared `expected_error` is accepted. Still exits 2 there.

532 tests pass across `tests/sanitizers` and `tests/instrumentation/rocjitsu_sanitizers`; `pre-commit run --all-files` clean.

vivekkhandelwal1: Fixed in `4314efc`, taking your `suggestion` block as written, and the test is next to `test_require_also_applies_to_the_gate_mode` as you asked.

Your reading of the blast radius is right on both halves: the nightly is fine today because it passes `--vacuous-only`, and the conflict check running for every invocation is what makes the drop worse than a plain missing argument. The option was accepted, validated against the baselines, and then discarded, so `compare_verdict_baselines.py --known-vacuous informational/consan-gemm <root>` failed that row as an ordinary vacuous report *and* printed the "declare it as an `expected_error`" advice — the one resolution a known-vacuous row must not be pushed toward, since it asserts the outcome is correct.

## Forwarded, not rejected

Both you and Copilot named the same two options, so briefly on why forwarding is the truer one rather than just the smaller diff:

- `--vacuous-only` chooses which **halves** run — it is there for trees without the gated cases. It does not change what the sweep means; the sweep is identical code over the same root on both paths.
- `--require` also describes the sweep and is already honoured on both paths, which is precisely what `test_require_also_applies_to_the_gate_mode` pins. Making its companion path-sensitive would have been an arbitrary distinction.
- The conflict check you pointed at sits *ahead* of the mode branch, so the file already treats the option as meaningful everywhere.

`_sweep_vacuous` already took `known_vacuous`, so this is the one-line forward. I also recorded the invariant in the module docstring beside the `--vacuous-only` note — the defect existed because that was never written down anywhere a reader or a test could check it.

## Checks

Nightly lane unchanged (it uses `--vacuous-only`): replayed over the real survey trees of runs 33755195031, 33872660689, 33965800908 and 34032862308 — exit 0, both blocked rows reported, six reports judged, `(2 known-vacuous)`.

For the changed path I used real reports rather than fixtures: the last-green gate tree from [33507543444](https://github.com/ROCm/aorta/actions/runs/33507543444) plus the vacuous ROCm 10 `consan-gemm` report from [34032862308](https://github.com/ROCm/aorta/actions/runs/34032862308).

| `--known-vacuous informational/consan-gemm <root>` | pre-fix | post-fix |
|---|---|---|
| exit | 1 | 0 |
| `consan_gemm` row | failed as an ordinary vacuous report, with the `expected_error` advice | `KNOWN VACUOUS, not failing`, reason `consan_strict_load_rejection` |

Gate half verified still armed on that same tree: blanking the expected finding shape on `consan-racy` still fails, with the known-vacuous row present.

Two tests:

- `test_known_vacuous_also_applies_to_the_gate_mode` — the gate-mode test you asked for. Confirmed it fails against the pre-fix code (`assert 1 == 0`, with the row failed as an ordinary vacuous report in the captured stdout), so it covers the defect rather than merely passing.
- `test_known_vacuous_conflict_is_caught_in_the_gate_mode_too` — the conflict check you noted runs everywhere; this pins that forwarding did not open a path where a contradictory row slips through. Still exit 2.

532 tests across `tests/sanitizers` and `tests/instrumentation/rocjitsu_sanitizers`, 3 GPU tests skipped on a non-GPU host; `pre-commit run --all-files` clean.

vivekkhandelwal1:
Two follow-ups on this revision, both pushed in `4314efc`.

**1. `--known-vacuous` reached the sweep only under `--vacuous-only`** — reported independently by @Copilot and @amd-vivekag. Forwarded rather than rejected, since `--vacuous-only` selects which halves run and not what the sweep means, and `--require` already applies on both paths. Detail in the two threads.

**2. This PR was still armed to auto-close #450**, despite the body saying it should not be, and despite @amd-vivekag asking to keep #450 open.

The cause is that GitHub's closing-keyword parser is not sentence-aware. The body contained the literal string `close #450` twice — once in the opening disclaimer (`see "Does this close #450?" below`) and once in the section heading `## Does this close #450?`. Both are arguing *against* auto-closing, and both matched. `closingIssuesReferences` confirmed it: `totalCount=1 -> #450 (OPEN)`, so merging would have closed the issue this PR deliberately does not close.

Reworded both to "Does this close the issue?" — no other body text changed. `closingIssuesReferences` is now `totalCount=0`.

Worth noting because it is the same failure shape as the defect the threads above are about, and as #450 itself: something that reads as a declaration of intent, is accepted, and quietly does the opposite. `Refs #450` was doing nothing; the prose next to it was doing the linking.

amd-vivekag (commented):
## Review summary

**Verdict: ship** (one Low nit, author's discretion — not blocking)

Re-review of `4314efc`. My Medium is fixed exactly as suggested: `known_vacuous` is now forwarded at the gate-path call site, with `test_known_vacuous_also_applies_to_the_gate_mode` and a companion asserting the conflict check still bites there. Forwarding rather than rejecting is the right call — the unconditional conflict check already treated the option as meaningful on every path, so rejecting it would have left `--require` and `--known-vacuous` disagreeing about where they are legal. Verifying the new test fails against the pre-fix code, and recording the invariant in the docstring, are both the right instincts.

**Prior comments — all resolved**
- **High** (ROCm 10 validation): resolved as a claim; CI gemm stays vacuous and is now suppressed rather than blessed. #450 stays open.
- **Medium** (`--known-vacuous` dropped on the gate path): resolved in `4314efc`.
- **Medium** (missing reports fail open): resolved via `--require`.
- Copilot exact `expected_error`: resolved.
- Copilot dashboard description: resolved.
- Copilot survey invocation: resolved.

Nothing still open from any prior review.

**New (Low), confirming Copilot's 11:55 note.** The sweep's "declare it as an `expected_error`" advice is a guaranteed dead end for `informational/waitcheck-gemm`, because `_case_key` maps it to the gated key `waitcheck_gemm` and `_load_baselines` rejects `expected_error` there. Following the advice gives exit 2, not a bypass, so this is misleading guidance rather than a hole. Inline below. I'd take it or wave it off; either way I'll approve on the next push.

scripts/sanitizers/compare_verdict_baselines.py
```diff
@@ -70,17 +241,247 @@ def _compare_case(name: str, report: SanitizerReport, expected: dict[str, object
...
+
+    reasons = _reasons(report)
+    shape = _vacuous_shape(report, reasons)
+
+    declared = expected.get("expected_error") if isinstance(expected, dict) else None
+    if declared is None:
+        return [
+            f"{shape} -- the run produced no sanitizer signal at all. Either the recipe "
+            f"is misconfigured (see ROCm/aorta#450) or, if this outcome is intended, "
+            f"declare it as an 'expected_error' entry in {_BASELINES}."
```
amd-vivekag: **[Low]** This advice is impossible for one row: `informational/waitcheck-gemm`.

`_case_key` maps that directory to the gated key `waitcheck_gemm`, and `_load_baselines` rejects `expected_error` on a gated key — so an operator who follows this message gets exit 2. Confirming Copilot's 11:55 note. Not a bypass and masked today by `--known-vacuous`, but it bites whoever removes that suppression. A `suggestion` block would need the case name plumbed in, so prose instead.

Fix: pass `name` into `_vacuous_problems` and, when `name in _CASES`, point at `--known-vacuous` or renaming the case directory instead of at `expected_error`.

scripts/sanitizers/compare_verdict_baselines.py
amd-vivekag: Resolved — forwarding is the right resolution, and verifying the new test fails against the pre-fix code is what makes it worth having. Thanks for the docstring invariant too.

scripts/sanitizers/compare_verdict_baselines.py
vivekkhandelwal1: Fixed in `77f9b9a`, and it is worse than "gets exit 2" — worth recording because it changes the priority.

The rejection is raised by `_load_baselines`, so it is not scoped to the offending row. An operator who follows the advice for `informational/waitcheck-gemm` does not get one failing case; the baselines file stops loading and the comparator refuses to judge *anything*:

```
$ # waitcheck_gemm.expected_error added, exactly as the message advised
$ compare_verdict_baselines.py --vacuous-only <root>
error: could not read baselines …/verdict_baselines.json: waitcheck_gemm.expected_error
is not allowed on a gated case: waitcheck_gemm is also the baseline key for a survey
case directory, so the exemption could not be attributed to one report or the other
exit 2
```

So the guidance did not just fail to help — it talked the operator into breaking the whole check. Agreed on the reasoning for fixing it under the suppression rather than after: the person who hits it is whoever removes `--known-vacuous`, and they arrive with a message that names the remedy the tool then refuses.

## What changed

Took your fix as directed — `name` plumbed into `_vacuous_problems` (with `case_dir` alongside, since `--known-vacuous` takes the directory and the key is a lossy transform of it), and `name in _CASES` selects the guidance. The advice now comes from `_vacuous_remedy`:

| row | guidance |
|---|---|
| ordinary survey row | unchanged — declare an `expected_error` |
| colliding survey directory | `--known-vacuous <case-dir>`, plus renaming as the route to a declaration, plus *why* the declaration is refused |
| the gated report itself | neither — a regression to fix in the recipe, or a baseline to update |

**One deliberate step beyond your suggestion, on the third row.** `name in _CASES` is true for two different situations: the colliding survey directory, and the gated report itself (`waitcheck/`). Pointing the latter at `--known-vacuous` would have been a second impossible remedy, because the sweep cannot rescue a gated case — `_compare_case` judges it against its own baseline and fails it regardless. Checked rather than assumed:

```
$ compare_verdict_baselines.py --known-vacuous waitcheck <vacuous gate tree>
waitcheck_gemm: overall_verdict='error', expected 'warn'
waitcheck_gemm: check 'waitcheck' findings do not contain expected shape 'missing s_waitcnt'
exit 1
```

That is Copilot's framing too — its 11:55 note flagged the advice as impossible "for gated keys" *and* separately wrong for the collision — so both sub-cases needed distinct guidance rather than one branch.

I kept it as prose in the message rather than restructuring the remedy into the baselines file: the constraint is a property of the key mapping, not of any declaration, so there is nothing for the file to hold.

## Tests

Three, next to `test_expected_error_is_rejected_on_a_gated_key` since that test is precisely what the old advice led to:

- `test_vacuous_advice_is_followable_for_a_colliding_survey_key` — asserts the impossible string is gone, the `--known-vacuous informational/waitcheck-gemm` advice is present, and then **follows that advice and asserts exit 0**. Guidance that is merely different is not the fix; guidance that works is.
- `test_vacuous_advice_for_a_gated_report_does_not_point_at_the_sweep` — the gated report gets neither remedy.
- `test_vacuous_advice_still_points_at_expected_error_for_an_ordinary_row` — pins that the change is scoped, so a future edit cannot strip the correct advice from every row.

The first two fail against the pre-fix code (verified by reverting the script and re-running); the third passes both ways by design, since it guards against over-application.

Real-tree replay unchanged: the wired nightly invocation is exit 0 on the survey trees of runs 33755195031, 33872660689, 33965800908 and 34032862308, and the gate trees are unmoved (33507543444 exit 0, 34032862308 exit 1 on the standing #453 row).

535 tests across `tests/sanitizers` and `tests/instrumentation/rocjitsu_sanitizers`; `ruff` and `pre-commit run --all-files` clean.

amd-vivekag (approved):
No further review comments — approving. Every finding from the previous rounds is fixed and verified; the branch is `BEHIND` `main`, so update it before merging.

scripts/sanitizers/compare_verdict_baselines.py
amd-vivekag: Resolved, and you were right that it deserved higher priority than I gave it — `_load_baselines` raising for the whole file means the advice broke every row, not just the one. Splitting the gated report from the colliding directory was the correct call; pointing the gated case at `--known-vacuous` would have been a second dead end. Verified the branch dispatch routes every real case directory to guidance that works, and following-the-advice-and-asserting-exit-0 is the right shape for that test.

PR #502, opened 2026-09-17
fix(sanitizers): ignore non-applicable coverage-site records

amd-vivekag:
## Summary

- Exclude `coverage_site disposition=not_applicable` debug records from reconciliation against RocJITsu's `*_discovered` counts.
- Keep strict reconciliation for every discovered supported/unsupported site.
- Cover access, barrier, atomic, and fence records.

## Root cause

RocJITsu intentionally excludes `NotApplicable` policy decisions when computing `*_discovered`, but still itemizes those decisions at debug log level. Aorta counted every itemized line, so a valid record such as `access_discovered=0` plus four `disposition=not_applicable` lines failed with `access site count mismatch`.

This caused both ConSan controls in sanitizer nightly run 35189763335 to be rejected during parsing even though the hook reported complete analysis.

## Validation

- Exact logs from nightly run 35189763335 now evaluate to:
  - `consan-clean`: `pass`, 0 findings
  - `consan-racy`: `fail`, 64 findings
- `tests/instrumentation/rocjitsu_sanitizers`: 170 passed, 3 skipped
- Ruff and IDE lints: clean

This is independent of the stale waitcheck bundle. The RocJITsu artifact workflow also needs its GCC 13 build fix so aorta can consume the #11718 binary.


vivekkhandelwal1 (commented):
**Superseded — see the approval at [pullrequestreview-5235381967](https://github.com/ROCm/aorta/pull/502#pullrequestreview-5235381967).**

This earlier review withheld approval over a test-narrowness gap. That gap is real and is now filed as an inline comment instead, since it is a test-coverage improvement rather than a correctness objection and does not warrant holding the PR. The findings and the evidence behind them are carried in the approving review; the lengthy analysis narrative that was here has been removed as noise.

src/aorta/instrumentation/rocjitsu_sanitizers/consan_coverage.py
```diff
@@ -368,7 +368,9 @@ def parse_coverage_decision(log_text: str) -> CoverageDecision:
             retained = [
                 site
                 for site in site_records
-                if site.identity == record.identity and site.kind == kind
+                if site.identity == record.identity
+                and site.kind == kind
+                and site.disposition != "not_applicable"
```
vivekkhandelwal1, line 373: Non-blocking, and the only real gap I found: this line's *effect* is pinned, but its *narrowness* is not. Four independent mutations of it each pass the full 621-test suite:

- key on `outcome` instead of `disposition`
- key on `lowering_reason`
- widen the exclusion to drop `unsupported` as well
- invert it into an allowlist that keeps only `not_applicable`

The only mutation that is caught is weakening the adjacent `!=` to `>`, and that is caught by the pre-existing `test_partially_itemized_site_kind_is_still_a_parse_error` rather than by anything this PR adds. So a later edit that quietly grows what gets excluded would land green.

Separately, a five-line change mirroring upstream's own shape would make the resulting message accurate rather than reporting a "count mismatch" for a disposition it does not recognise: an explicit allowlist of the three documented values, with anything else treated as unparseable. I tested that variant — 621 passed, and the two real logs still evaluate correctly. Worth doing, not worth holding the PR for.

src/aorta/instrumentation/rocjitsu_sanitizers/consan_coverage.py
```diff
@@ -368,7 +368,9 @@ def parse_coverage_decision(log_text: str) -> CoverageDecision:
...
             ]
             discovered = counts[f"{kind}_discovered"]
```
vivekkhandelwal1, line 375: A latent second divergence from the upstream gate, recorded here rather than asked of this PR. Upstream collapses its ledger by physical text offset before comparing against `*_discovered`; this counts raw records. Two itemized records sharing a `text=` offset therefore reconcile differently in the two implementations.

I found no instance of it in the real logs, so it is not a live defect. What makes it worth the note is the symptom: it surfaces as the *same* `site count mismatch` string this PR is fixing, so if it ever fires it will read as a regression of this change. Not a one-liner either, since `_SiteRecord` does not capture `text`.

tests/instrumentation/rocjitsu_sanitizers/test_consan.py
```diff
@@ -561,6 +561,29 @@ def test_partially_itemized_site_kind_is_still_a_parse_error() -> None:
...
+                "outcome=not_applicable lowering_reason=semantic_not_applicable "
+                "resource_reason=none container=k scope=kernel text=0xc "
+                "mnemonic=unknown"
+            ),
+        )
+    )
+
+    _waitcheck, consan = evaluate_record_replay(ProcessResult(("app",), 0, output, ""), strict=True)
+
+    assert consan.verdict is Verdict.PASS
```
vivekkhandelwal1, line 584: This assertion is what pins the fix's effect. Consider adding its counterpart in the same test — that a *discovered* site relabelled `not_applicable` still fails closed:

```python
assert consan.verdict is Verdict.ERROR
assert "site count mismatch" in str(consan.reason)
```

I verified that behaviour holds today on all four kinds, so the assertion would pass as written; it simply isn't guarded. This pairing is also what would catch the mutations described on the changed line in `consan_coverage.py`. An exclusion asserted only by what it lets through, with no counterpart asserting what it must still reject, is the shape that lets a filter widen unnoticed.

tests/instrumentation/rocjitsu_sanitizers/test_consan.py
```diff
@@ -561,6 +561,29 @@ def test_partially_itemized_site_kind_is_still_a_parse_error() -> None:
...
+@pytest.mark.parametrize("kind", ("access", "barrier", "atomic", "fence"))
+def test_not_applicable_site_is_excluded_from_discovered_count(kind: str) -> None:
+    # The hook itemizes policy decisions that it deliberately excludes from
+    # *_discovered. Those debug records must not make otherwise self-consistent
+    # coverage look malformed (nightly run 35189763335).
+    output = "\n".join(
+        (
+            _healthy_evidence(),
+            (
+                f"{_PREFIX} coverage_site reader=1 load=1 kind={kind} "
```
vivekkhandelwal1, line 573: Worth knowing how much of this parametrization is load-bearing. `_healthy_evidence()` carries `access_discovered=2`, so the `access` case genuinely exercises a kind that was discovered, while `barrier`, `atomic` and `fence` each run against `discovered=0`. The `discovered and not retained` branch is therefore only reached for `access`.

Not a defect: the loop has no per-kind branch, so the symmetry is structural, and I confirmed all four behave identically in both directions. But neither real log contains a barrier, atomic or fence site, so those three cases rest on synthetic input and pin less than a count of four suggests.

vivekkhandelwal1 (approved):
**Approving** — reviewed at `7e3502677479308d66448a0b7d0fcbad21280b5d`, base `73409f79`. Stating the SHA because classic branch protection is disabled on this repo, so this approval is not dismissed by a later push.

The change is correct and I could not break it.

- **The premise is confirmed at source, not inferred from the counts agreeing again.** RocJITsu ships its own parser for these records at `emulation/rocjitsu/tests/dbi/consan/consan_coverage_gate.py` on branch `shared/rocjitsu/sanitizers`, and lines 539-547 perform the identical `disposition != "not_applicable"` exclusion, commented "Filter exclusions remain in the verbose inventory but are not instrumentation sites and therefore do not enter the aggregate." The reference doc independently gives `disposition` a closed three-value set and the accounting identity `discovered = supported + unsupported`. So this restores agreement with the gate this parser was ported from, rather than loosening a check. The doc's "`NotApplicable` records are omitted from output" versus the logs containing them resolves via log level: aorta sets `RJ_CONSAN_LOG=3`.
- **A real coverage failure cannot be laundered through the new filter.** Relabelling a *discovered* site `not_applicable` still fails closed on all four kinds, and `_failure_attributions` runs over the unfiltered record list, so a record carrying `outcome=placement_or_lowering_failed` or `resource_failed` still surfaces and still rejects. Near-miss tokens (`NOT_APPLICABLE`, `notapplicable`, `not_applicable2`) all fail closed. The one newly-accepted shape is strictly dominated by one `main` already accepts: an object reporting `discovered=0` that itemizes nothing passes on both revisions.
- **The validation claim reproduces in both directions.** Replaying the two real `consan.log` artifacts from run 35189763335: on `main` the archived parse errors come back verbatim, including both reader IDs; at this head, `consan-clean: pass` with 0 findings and `consan-racy: fail` with 64, all 64 matching the `auto replay diagnostic` shape the recipe baseline requires. Suites: `rocjitsu_sanitizers` 166 passed on `main` to **170 passed, 3 skipped** here, the delta being exactly the four new cases; `tests/sanitizers` 451 passed on both.
- A full revert of the changed line is **caught**, on all four kinds.

Four non-blocking comments are inline: a test-narrowness gap, a latent second divergence from the upstream gate, and a note on how much of the new parametrization is load-bearing. None is substantive enough to hold the PR.

**One correction to the description, because it will be read as a status claim.** This does not turn the sanitizer nightly green. The gating job fails three scenarios; this clears `consan_clean` and `consan_racy`, while `waitcheck_gemm` fails independently with `worklist_not_fully_checked` and has zero ConSan coverage records in its log — that one is aorta#453, upstream rocm-systems#11428. The independence claim in the description is true; the inference a reader draws from it will not be.

Not verified: only the `access` parameter exercises a kind that was actually discovered, and no real log in hand contains a barrier, atomic or fence site, so three quarters of the new parametrization rests on synthetic input.

Issue #330, opened 2026-08-05
Pin sanitizers-nightly RocJITsu build to an immutable reviewed commit


amd-vivekag:
The `sanitizers-nightly.yml` workflow currently checks out RocJITsu via the mutable branch ref `shared/rocjitsu/sanitizers` (env `ROCJITSU_REF`). Per the PR #324 review, a Phase-1 gate must pin an **immutable** input so a branch retag cannot silently change gate behavior.

**Task**
- Replace `ROCJITSU_REF` with the reviewed RocJITsu commit SHA (40-char) in `.github/workflows/sanitizers-nightly.yml`.
- Record the chosen commit in the PR/gate docs.

Context: the clone is already token-free (rocm-systems is public) and uses `${RUNNER_TEMP}` paths with always-run cleanup; only the immutable pin remains.

PR #505, opened 2026-09-17
fix(sanitizers): make ConSan Sampled the guarded default

### LINKED ISSUE #503: sanitizers: migrate ConSan guardrails to Sampled mode
## Problem

RocJITsu now defaults to and recommends ConSan MOI Sampled mode. Record/Replay and Inline Shadow are expected to be retired, but aorta currently overrides RocJITsu's default with:

```python
RJ_CONSAN_MODE=record-replay
```

Aorta's finding parser is also Record/Replay-specific: it recognizes `auto replay diagnostic` and `auto replay` summaries but not Sampled conflict records or summary counts. Once Record/Replay is removed, ConSan recipes and the sanitizer dashboard will fail during hook configuration or silently miss Sampled findings.

## Required behavior

- Make Sampled the ConSan mode requested by aorta.
- Parse detailed `ConSan MOI auto sampled conflict` records.
- Parse summary-only Sampled conflicts from `sampled_conflicts` and `sampled_immediate_conflicts` without double-counting detailed examples.
- Preserve compatibility with Record/Replay output from older downloaded bundles and saved logs.
- Continue using strict static/dynamic coverage validation.
- Make the nightly clean/racy controls deterministic by selecting the `max` Sampled preset (workgroup and cell stride 1); the ordinary 1/256 default may statistically miss the positive racy control.
- Scrub inherited mode/preset environment values so callers cannot weaken or accidentally alter the pinned gate contract.
- Update tests and docs/comments that describe Record/Replay as the only supported mode.

## Acceptance criteria

- Clean Sampled evidence produces `pass`.
- Sampled conflict evidence produces `fail` with correctly attributed findings.
- Summary-only conflicts remain visible.
- Record/Replay fixtures continue to parse.
- `run_consan()` exports `RJ_CONSAN_MODE=sampled` and `RJ_CONSAN_MOI_SAMPLED_PRESET=max`.
- Existing sanitizer tests and the nightly recipe/baseline contract remain green.

## Context

- RocJITsu commit `164c20fae8c` made Sampled the default and updated its guidance.
- The `not_applicable` coverage-site compatibility fix is tracked separately in #502 and remains necessary because static coverage logging is shared across modes.

amd-vivekag:
## Summary

- Make aorta request RocJITsu's retained default ConSan detector after Record/Replay and the explicit Sampled name were removed.
- Pin `RJ_CONSAN_PRESET=max` and scrub every inherited `RJ_CONSAN_*` control so the clean/racy guardrails remain deterministic.
- Parse current detail conflicts, summary-only conflicts, and immediate device conflicts without mistaking benign access evidence for races.
- Preserve legacy Sampled and Record/Replay parser/evaluator compatibility for older saved logs.
- Update the racy baseline, dashboard reproduction environment, and current mode documentation.

## Safety properties

- Every default-detector report snapshot is required and internally cross-checked; one valid snapshot cannot mask a truncated sibling.
- Detailed examples and summary counts are reconciled per reader and per schema, so legacy details cannot suppress current summaries or vice versa.
- Evidence-loss/malformed counters and a hook that ignores the pinned mode fail closed.
- Current `mode=default` / `ConSan auto report` / `ConSan conflict` and legacy Sampled/Record-Replay schemas are both covered.

## Compatibility / dependency

The latest-successful bundle moved while this PR was under review. Artifact run [35300134760](https://github.com/ROCm/rocm-systems/actions/runs/35300134760) publishes commit `9d72e86f2999`, which contains RocJITsu's `5d5a0f1` mode/naming simplification. This PR follows that current contract while retaining parsers for old saved logs. #502 is merged and is the parent of this branch.

## Validation

- `tests/instrumentation/rocjitsu_sanitizers/test_consan.py`: 87 passed
- `tests/instrumentation/rocjitsu_sanitizers tests/sanitizers`: 667 passed, 3 skipped
- Ruff, pre-commit, diff whitespace, and IDE lints: clean
- planner-rubric evaluation: 100/100 functional requirements
- Current-head MI350 [sanitizer-nightly run 35318482298](https://github.com/ROCm/aorta/actions/runs/35318482298), using RocJITsu bundle `9d72e86f`:
  - gate job: success
  - waitcheck: expected `warn`, execution complete, 64 findings
  - default-detector clean: `pass`, `dynamic_complete=true`, 0 conflicts
  - default-detector racy: `fail`, `dynamic_complete=true`, 1 attributed `ConSan conflict`
  - both controls: 0 dropped/saturated/malformed windows

The non-gating survey remains responsible for confirming the `consan_tiny` observation.

Fixes #503.


src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -40,11 +59,43 @@
...
+    "RJ_CONSAN_MOI_CELL_SAMPLE_OFFSET",
+    "RJ_CONSAN_MOI_CELL_SAMPLE_STRIDE",
+    "RJ_CONSAN_MOI_RUNTIME_SAMPLE_OFFSET",
+    "RJ_CONSAN_MOI_RUNTIME_SAMPLE_STRIDE",
+    "RJ_CONSAN_MOI_SAMPLED_BANKS",
+    "RJ_CONSAN_MOI_SAMPLE_OFFSET",
+    "RJ_CONSAN_MOI_SAMPLE_STRIDE",
+    "RJ_CONSAN_MOI_WORKGROUP_SAMPLE_OFFSET",
+    "RJ_CONSAN_MOI_WORKGROUP_SAMPLE_STRIDE",
+)
```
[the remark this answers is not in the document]
amd-vivekag: Good catch - fixed in 3de5d4f9. Confirmed against upstream `USAGE.md`: an unselected epoch has its report reset "without copying, decoding, analyzing, or rendering it", so "diagnostics in an unselected epoch do not contribute to the final verdict", and `manual` analyses nothing at all unless a harness calls `rj_dbi_consan_begin_epoch_analysis_window()` - which `run_consan` never does. So an inherited value really could drop the positive control's race even at `preset=max`.

Added `RJ_CONSAN_MOI_EPOCH_ANALYSIS` to the scrubbed set alongside the sampling selectors and the same-value opt-in, with a comment recording why. The gate-override test is parametrised over the whole tuple, so it now covers this variable too.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -182,20 +359,32 @@ def parse_record_replay_output(output: str) -> ParsedCombinedOutput:
...
+                    severity=FindingSeverity.RACE,
+                    code="sampled_conflict",
+                    message=line,
+                    kernel_name=fields.get("kernel"),
+                    code_object=fields.get("code_object"),
+                    metadata=tuple(sorted(fields.items())),
+                )
+            )
+        elif _AUTO_SAMPLED_SUMMARY.search(line):
+            sampled_summaries.append(fields)
```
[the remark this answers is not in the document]
amd-vivekag: Good catch - fixed in 3de5d4f9. I reproduced both halves before acting, and each really did reach a clean `pass`:

- report line truncated before `sampled_conflicts=` -> `pass`, 0 findings
- `auto report` line lost entirely -> `pass`, 0 findings

Two changes. The summary is now recognized from `visible_sampled`, the second field the renderer appends, so a line truncated between there and the counters is still seen as a Sampled summary and fails closed on the missing ones. And every applicable Sampled reader is reconciled against the coverage records, so a summary that never arrived is fatal rather than silently unaccounted for.

One correction to the suggested shape, worth recording: I first keyed on the *first* field, `sampled_watchpoints`, and that broke every healthy run - `auto report plan` is an allocation line that carries the same name and no counters. The real MI350 logs went red until I moved the marker to `visible_sampled`, which `plan`/`buffer`/`cleanup` never emit. Pinned that with its own regression test.

Reconciliation is scoped to *applicable* readers, since a loaded helper object that discovers no site is never instrumented and legitimately publishes no report - both nightly controls carry one. Verified against the real gfx950 logs: clean still `pass`, racy still `fail` with the conflict attributed.

scripts/sanitizers/gen_sanitizer_dashboard.py
```diff
@@ -2287,6 +2287,12 @@ def run_area_files(case_dir: Path) -> list[tuple[str, int]]:
         "purpose": "ConSan detection mode",
         "consan_only": True,
     },
+    {
+        "var": "RJ_CONSAN_MOI_SAMPLED_PRESET",
+        "set_by": "aorta",
+        "purpose": "max Sampled preset for deterministic guardrail coverage",
+        "consan_only": True,
+    },
```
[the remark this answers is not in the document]
amd-vivekag: Good catch - fixed in e25a1973. Both descriptions were mine to keep current and I left them behind when I added the fifth entry.

Rather than bump four to five, I phrased both by the `consan_only` flag that actually selects the entries, so the next variable added here cannot silently falsify the comment the same way. Swept the rest of the contract for the same hardcoded-count shape: `required_env_note` already derives its prose from the passed list, and the dashboard test asserts the explicit variable list, so neither carries a count to drift.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -29,6 +29,34 @@
...
+# Keyed on ``visible_sampled``, which the renderer writes second, rather than on
+# a conflict counter near the end of the line: a line truncated between the two
+# then still reports as a Sampled summary and fails closed on the missing
+# counters instead of going unrecognized. It cannot key on the first field
+# (``sampled_watchpoints``) because the ``auto report plan`` allocation line
+# carries that name too and no counters, so every healthy run would fail. Any
+# truncation earlier still than this is caught by the reconciliation against the
+# coverage records in _require_sampled_summaries.
+_AUTO_SAMPLED_SUMMARY = re.compile(r"\bauto\s+report\b.*\bvisible_sampled=", re.IGNORECASE)
+_SAMPLED_SUMMARY_COUNTS = (
```
vivekkhandelwal1, line 47: **Blocking.** These four are the only Sampled counters read, and the real `auto report` line carries ten more that report lost or malformed evidence — none consulted.

Taking run 35220499336's real `consan-clean/consan/consan.log` and flipping exactly one counter from `0` to `9`, the verdict stays `pass` for every one of `sampled_dropped_windows`, `sampled_saturated_windows`, `sampled_stale_snapshots`, `sampled_incomplete_snapshots`, `sampled_malformed_snapshots`, `sampled_malformed_sync`, `sampled_static_mapping_malformed`, `sampled_unsupported_sync`, `sampled_pending_acquire_malformed` and `sampled_changed_snapshots`. On the same log `sampled_immediate_conflicts=4` gives `fail` and deleting `sampled_conflict_pairs_without_example` gives `error`, so the four that are read do bite.

They are already in the `fields` dict at this point, so the fix needs no new parsing:

```python
_SAMPLED_INTEGRITY_COUNTS = (
    "sampled_dropped_windows",
    "sampled_saturated_windows",
    "sampled_stale_snapshots",
    "sampled_incomplete_snapshots",
    "sampled_malformed_snapshots",
    "sampled_malformed_sync",
    "sampled_static_mapping_malformed",
    "sampled_unsupported_sync",
    "sampled_pending_acquire_malformed",
)
```

and a non-zero value raises from `_sampled_totals`, or routes to `consan_coverage_incomplete`.

This is the counterpart of something already enforced for the engine being replaced: `consan_coverage.py`'s `_VERDICT_COUNTS[1:]` requires `dynamic_incomplete` and `replay_metadata_full` among others to be zero, and `replay_metadata_full=3` yields `error` on both `main` and this head. Dropped windows or malformed snapshots mean the pairwise analysis saw less than it should have, which surfaces as fewer conflicts, which is `pass`.

Two questions rather than assumptions. Does the hook set `dynamic_incomplete` on the analysis verdict line whenever any of these goes non-zero? If so the run already fails closed and this reduces to a comment saying why these are safe to ignore, which is still worth having. And is `sampled_changed_snapshots` benign? The membership of the set is yours; that none of the ten is read is the point.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -157,12 +239,150 @@ def _parse_waitcheck(
...
+    instrumentation and stay healthy either way. Reconciling against coverage
+    keeps the absence itself fatal rather than silently unaccounted for. Only
+    applicable readers are required: a loaded object with no discovered site is
+    never instrumented and publishes no report, which is the ordinary shape of
+    the runtime helper objects that accompany a repro.
+    """
+    missing = sorted(
+        str(record.reader)
+        for record in coverage
+        if record.engine == "sampled" and record.applicable and str(record.reader) not in totals
```
vivekkhandelwal1: The requirement is keyed on what the log says the engine was, but `run_consan` is the thing that pinned `RJ_CONSAN_MODE=sampled` two hundred lines down. So if the hook does not honour that request, no Sampled requirement is imposed at all: a log whose coverage records carry `engine=record_replay` or `engine=inline_shadow` and which contains no conflict evidence of either dialect evaluates to `pass` with zero findings. I confirmed that for all three non-Sampled engine values.

That matters because of how the bundle is chosen. `sanitizers-nightly.yml` resolves `ROCJITSU_REF: shared/rocjitsu/sanitizers` as latest-successful at run time, there is no commit pin, this PR adds none, and the workflow's own `TODO(#330)` says the artifacts are ephemeral 30-day ones. The dependency is satisfied today — `6e6574b59b8a` is 6 commits ahead of `164c20fae8c` — but it is satisfied by the branch happening to be ahead, not by anything that fails if an older bundle is served.

If an older hook rejects the unknown mode or preset outright, the non-zero exit fails closed and there is no issue. If it ignores them and runs a different engine, this is a false clean.

Since aorta pins the mode itself, the strictly fail-closed form asserts what it requested rather than what it was told:

```python
if not any(record.engine == "sampled" for record in coverage):
    raise ValueError("ConSan ran without Sampled evidence despite RJ_CONSAN_MODE=sampled")
```

Worth noting that this predicate is not pinned by a test either: rewriting it as `record.engine != "record_replay"` leaves the whole suite green.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -82,6 +151,19 @@ def _int(fields: dict[str, str], key: str) -> int:
     return value
 
 
+def _required_int(fields: dict[str, str], key: str) -> int:
```
vivekkhandelwal1, line 139: Non-blocking, narrowness. The docstring's reasoning is right and the guard works — dropping any one of the four counters from the real log yields `error`. But only one of them is pinned by a test. Restricting the raise to `key == "sampled_conflicts"`, so the other three silently read as zero, leaves the entire suite green, because the truncation test cuts before all four and so still trips on the one that stays enforced.

Since zero is "the one value this must not invent", it is worth parametrizing a test over the four names in `_SAMPLED_SUMMARY_COUNTS` and dropping each in turn. That is the mutation that catches this requirement being relaxed for one counter later, rather than only catching it being deleted wholesale.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -205,8 +437,13 @@ def parse_record_replay_output(output: str) -> ParsedCombinedOutput:
                     metadata=tuple(sorted(summary.items())),
                 )
             )
+    sampled_totals = _sampled_totals(sampled_summaries)
+    findings.extend(
+        _sampled_summary_findings(sampled_totals, _itemized_by_reader(sampled_details))
```
vivekkhandelwal1, line 479: Non-blocking, narrowness. The per-engine split is correct, and I verified both directions behave: a Record/Replay detail record does not satisfy a Sampled summary, and the reverse also holds. Only one direction is tested, though.

`test_legacy_and_sampled_evidence_do_not_cross_contaminate` catches widening the *replay* side (crediting sampled details against a replay summary). Widening this side — `_itemized_by_reader([*sampled_details, *replay_details])` — leaves the suite green, and that is the direction where a Sampled shortfall finding gets suppressed by unrelated legacy records sharing the log. Given the PR claims examples and counts are reconciled per engine "so one cannot suppress the other", the mirror-image assertion is worth adding.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -29,6 +29,34 @@
...
+#
+# Keyed on ``visible_sampled``, which the renderer writes second, rather than on
+# a conflict counter near the end of the line: a line truncated between the two
+# then still reports as a Sampled summary and fails closed on the missing
+# counters instead of going unrecognized. It cannot key on the first field
+# (``sampled_watchpoints``) because the ``auto report plan`` allocation line
+# carries that name too and no counters, so every healthy run would fail. Any
+# truncation earlier still than this is caught by the reconciliation against the
+# coverage records in _require_sampled_summaries.
+_AUTO_SAMPLED_SUMMARY = re.compile(r"\bauto\s+report\b.*\bvisible_sampled=", re.IGNORECASE)
```
vivekkhandelwal1: Non-blocking, narrowness. The comment above gives two reasons both halves are required: `auto report` also prefixes plan and allocation lines carrying no counters, and the `sampled_*` names also appear in the `RJ_CONSAN_MOI_REQUIRE_DIAGNOSTICS` rejection message. The first is tested — keying on `sampled_watchpoints=` instead is caught by `test_sampled_report_plan_line_is_not_mistaken_for_a_summary`. The second is not: dropping the `auto report` half entirely, leaving just `visible_sampled=`, leaves the suite green, because the plan-line fixture carries no `visible_sampled=` token and nothing exercises a non-report line that does.

A fixture for the rejection message, if its text carries `visible_sampled=`, would pin the half of the rationale that is currently only asserted in prose.

Separately, and this one is a question rather than a finding: keying on `visible_sampled` rather than a late counter is justified above on the grounds that a line truncated between the two "still reports as a Sampled summary and fails closed on the missing counters instead of going unrecognized". I could not make that distinction observable — keying on the last counter instead also leaves the suite green, and both variants fail closed, one through `_required_int` and the other through `_require_sampled_summaries`. That reads as defence in depth working as intended, but it does mean the stated reason for the key choice has no test behind it.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -40,11 +68,52 @@
...
+_CONSAN_SAMPLED_PRESET = "max"
+
+# Everything inherited from the caller's shell that could weaken the gate, of
+# which there are three kinds. A preset supplies DEFAULTS for the selection
+# knobs only, so an inherited selector silently thins ``max`` back out, and
+# mixing the coupled with the per-axis selectors is a hard hook config error.
+# The same-value opt-in suppresses a class of conflicts outright. The epoch
+# selector decides which epochs are analysed at all. Scrub all three so the
+# pinned mode and preset are the whole contract.
+_SAMPLED_GATE_OVERRIDES = (
```
vivekkhandelwal1: This works, and the way it is tested is the right way round. Setting all 11 of these in the real process environment, none reaches the hook, and `RJ_CONSAN_MODE=sampled`, `RJ_CONSAN_MOI_SAMPLED_PRESET=max`, `RJ_CONSAN_POLICY=strict` and `HSA_TOOLS_DISABLE_REGISTER=1` all arrive pinned over hostile inherited values. Because the test spells the list out instead of importing it, deleting a single name is caught — I checked with `RJ_CONSAN_MOI_SAMPLED_BANKS` and `RJ_CONSAN_MOI_EPOCH_ANALYSIS`, and both go red. The reasoning about `manual` and `nth:N` epoch selection dropping the positive control's race is the kind of thing that is easy to miss and expensive to debug later.

One question. `RJ_CONSAN_MOI_REQUIRE_DIAGNOSTICS` is named in the comment on the summary regex, so it exists, and it is not scrubbed. By its name it would make the hook stricter rather than weaker, so it is probably harmless here — but if it changes what the hook emits, an inherited value changes the log shape the parser sees. Is it deliberately left inherited?

Also worth considering alongside the integrity-counter finding: if there is a hook variable governing sampled window capacity, an inherited value could induce dropped or saturated windows, which currently yields `pass`.

recipes/sanitizers/fixtures/expected/verdict_baselines.json
```diff
@@ -23,7 +23,7 @@
       "consan": "fail"
     },
     "finding_shape": {
-      "consan": "auto replay diagnostic"
+      "consan": "ConSan MOI auto sampled"
```
vivekkhandelwal1: Checking this is the intended trade rather than flagging a defect. The verdict either side is unchanged — `overall_verdict` and `checks.consan` stay `fail` — so nothing was relaxed from failing to passing, which is the thing worth confirming about a baseline edit.

The new string is the common prefix of all three Sampled finding messages: the raw `ConSan MOI auto sampled conflict ...` detail line, `ConSan MOI auto sampled conflicts reported by the report summary only: ...`, and `ConSan MOI auto sampled immediate conflicts counted on the device: ...`. So where `auto replay diagnostic` pinned the itemized-detail form specifically, this now accepts any of the three. Given the summary-only case is described as the common one rather than a corner, accepting all three looks deliberate and right.

The consequence to be aware of is that the racy gate can no longer tell an itemized conflict from a summary-only count, so if the reproducer ever degraded to being detected only as an unitemized count, this case would still report green. If you want that distinguished, the per-code grouping the dashboard already does in `test_summarize_consan_keeps_the_three_sampled_finding_kinds_apart` is the place it exists.

vivekkhandelwal1 (changes requested):
**Not approving — requesting changes** on `e25a1973eaa6d52b1f47c13ea2665c3ca525a98c`. One blocking finding, and it is in the property this PR exists to establish. Everything else below is non-blocking; the change is well built and the test suite is unusually strong, so the block is narrow and the fix is small.

Stating the SHA explicitly because classic branch protection is disabled on this repo, so a review is not dismissed by a later push.

## Blocking: the Sampled engine's own evidence-integrity counters are never read, so a run that reports malformed or incomplete evidence still reports `pass`

The stated safety property is that "malformed or truncated Sampled output remains fail-closed". That holds for the shapes the parser was written against, and I confirmed it across roughly thirty adversarial inputs. It does not hold for the hook telling us in its own counters that its evidence was malformed or incomplete.

The real `auto report` line from run 35220499336 — the exact line `_AUTO_SAMPLED_SUMMARY` already matches and `_kv` already parses into `fields` — carries these, all zero on that run:

```
sampled_dropped_windows=0 sampled_saturated_windows=0 sampled_stale_snapshots=0
sampled_incomplete_snapshots=0 sampled_changed_snapshots=0 sampled_malformed_snapshots=0
sampled_malformed_sync=0 sampled_static_mapping_malformed=0 sampled_unsupported_sync=0
sampled_pending_acquire_malformed=0
```

`_SAMPLED_SUMMARY_COUNTS` reads four counters, none of them these. Reproduction, taking that run's real `consan-clean/consan/consan.log` unmodified and flipping exactly one counter from `0` to `9`:

| counter set to 9 in the real clean log | verdict |
|---|---|
| `sampled_dropped_windows` | `pass` |
| `sampled_saturated_windows` | `pass` |
| `sampled_incomplete_snapshots` | `pass` |
| `sampled_malformed_snapshots` | `pass` |
| `sampled_malformed_sync` | `pass` |
| `sampled_stale_snapshots` | `pass` |
| `sampled_static_mapping_malformed` | `pass` |
| `sampled_unsupported_sync` | `pass` |
| `sampled_pending_acquire_malformed` | `pass` |
| `sampled_changed_snapshots` | `pass` |

Controls on the same log, to show the harness is sound rather than inert: `sampled_immediate_conflicts=4` gives `fail`, `sampled_conflicts=4` gives `error` via the internal cross-check, and deleting `sampled_conflict_pairs_without_example` gives `error`. So the four counters that are read do bite; the integrity counters are simply not consulted.

What makes this a regression rather than a pre-existing gap is the asymmetry with the engine being replaced. `consan_coverage.py` already enforces exactly this class of counter for Record/Replay: `_VERDICT_COUNTS[1:]` requires `dynamic_incomplete`, `replay_unsupported_access`, `replay_unsupported_atomics`, `replay_unsupported_fences` and `replay_metadata_full` to all be zero, and a non-zero `replay_metadata_full` yields `consan_coverage_incomplete` (verified: `replay_metadata_full=3` gives `error` on both `main` and this head). This PR moves the gate onto an engine whose equivalent counters are unchecked, from one whose were checked. The direction of the resulting failure is the bad one: windows dropped or snapshots malformed means the host's pairwise analysis saw less than it should have, which shows up as fewer conflicts, which is `pass`.

Minimal fix, in `_sampled_totals` where the fields are already in hand — no new parsing, and it mirrors what `consan_coverage.py` does for the other engine:

```python
_SAMPLED_INTEGRITY_COUNTS = (
    "sampled_dropped_windows",
    "sampled_saturated_windows",
    "sampled_stale_snapshots",
    "sampled_incomplete_snapshots",
    "sampled_malformed_snapshots",
    "sampled_malformed_sync",
    "sampled_static_mapping_malformed",
    "sampled_unsupported_sync",
    "sampled_pending_acquire_malformed",
)
```

then raise, or route to `consan_coverage_incomplete`, when any is non-zero.

Two things I could not settle and would rather you decide than have me guess:

- I have no access to the Sampled renderer source, so the semantics above are read off the names and off the line's own convention (`dropped_records`/`capacity`, `dropped_barriers`/`barrier_capacity`, and `sampled_claimed_windows`/`dropped`/`saturated` as a triple). If `dynamic_incomplete` on the analysis verdict line is already set whenever any of these goes non-zero, then the run does fail closed today and the whole finding collapses to a comment saying so. That is worth writing down either way, because nothing currently records it.
- `sampled_changed_snapshots` may well be benign. The exact membership of the set is yours; that none of the ten is read is the defect.

## Non-blocking

**The guard added after the GPU run was never exercised on GPU.** Run 35220499336 ran at 12:19:55Z on `4f94328`, which is not in this branch's history. All three current commits postdate it: `6c46031` at 12:47:33Z, `3de5d4f` at 12:57:14Z, `e25a197` at 13:11:55Z. `3de5d4f` is the commit that introduced `_require_sampled_summaries`, which can turn a clean pass into `consan_output_parse_error` when an applicable Sampled reader publishes no report line — so the one guard capable of converting the clean case from `pass` to `error` is not covered by the evidence offered for the clean case. I checked this the only way available without hardware, by re-running the uploaded logs through this head, and it is fine: `consan-clean` gives `pass` and `consan-racy` gives `fail` with one `sampled_conflict`. That confirms the claim, and it is real bundle output rather than a fixture, but it is a replay of a log rather than a run. A dispatch on the current head before merge would close it properly.

**The bundle dependency is satisfied today but pinned nowhere.** `164c20fae8c` is "Improve ConSan Sampled coverage and make it the default" (2026-09-16 18:14Z), and the published bundle `6e6574b59b8a` from run 35209663700 is 6 commits ahead of it with 0 behind, so the requirement genuinely holds right now. But `sanitizers-nightly.yml` resolves `ROCJITSU_REF: shared/rocjitsu/sanitizers` as latest-successful at run time, this PR touches neither the workflow nor the downloader, and the workflow's own `TODO(#330)` notes these are ephemeral 30-day artifacts. So the dependency rests on that branch staying ahead, with nothing in-tree that fails if an older bundle is ever served. Given the nightly has been red for a long run of nights under #453, the version of this that would bite is quiet rather than loud.

**Verdict shape assertions and mutation gaps** are on the inline comments; the substantive ones are the engine predicate in `_require_sampled_summaries`, the untested direction of the cross-engine claim, and the `auto report` half of the summary regex.

## Verified positively

Worth saying plainly, because most of this change stands up to deliberate attack:

- **Fail-closed, adversarially.** Dropping each of the five required summary fields in turn, truncating the summary at three different points, removing the summary entirely, four flavours of internally inconsistent counters, `dynamic_complete=false`, and absent coverage records all yield `error`, never `pass`.
- **Benign records cannot launder a real conflict.** I could not construct one. Benign `auto sampled reader=...` evidence and the `omitted=N after log limit=N` record both stay clean; a conflict alongside them still fails; a conflict whose detail record is reshaped to dodge `_AUTO_SAMPLED_CONFLICT` entirely still fails, because the summary counters are the authority and the detail regex can only ever cause over-reporting. Deriving the shortfall as `max(pairs_without_example, conflicts - itemized)` is what makes that true, and reverting it to either term alone is caught by a test.
- **Per-reader and per-engine reconciliation both work.** A conflict detail for reader 1 does not satisfy reader 2's summary, and a Record/Replay detail record does not satisfy a Sampled summary — each produces two findings rather than suppressing one.
- **The Record/Replay alias is an alias, not a lenient fallback.** `parse_record_replay_output is parse_consan_output` and `evaluate_record_replay is evaluate_consan_output` are both `True`, and there is no `try`/`except` inside the parse body, so a new-format rejection cannot land in a permissive second parse. Across 14 old-format shapes the verdicts are byte-identical to `origin/main`.
- **Environment scrubbing does what it says.** Setting all 11 names in the real process environment, zero reach the hook, and `RJ_CONSAN_MODE=sampled`, `RJ_CONSAN_MOI_SAMPLED_PRESET=max`, `RJ_CONSAN_POLICY=strict` and `HSA_TOOLS_DISABLE_REGISTER=1` are all sent as pinned, overriding hostile inherited values.
- **Neither suspicious fixture is what it looks like.** `lds_dispatch.hip` is comment-only — filtering the diff to non-comment added and removed lines leaves nothing — and it is the race-free dispatch control, not the racy reproducer. The racy case uses `fixtures/repro/consan_lds_race_2wave.hip`, untouched here. `verdict_baselines.json` changes one field, `consan_racy.finding_shape.consan`, while `overall_verdict` and `checks.consan` stay `fail`. No expectation was relaxed.
- **The tests are genuinely load-bearing.** Of 25 mutations, 19 are caught, including every full revert but two: the required-field raise, the internal cross-check, per-reader reconciliation, the replay-side split, the `max(...)` shortfall, the coverage reconciliation, the widened conflict regex, the immediate/host split, the scrub loop, dropping a single name from the scrub list, and the mode and preset pins. Spelling the scrub list out in the test rather than importing it is why dropping one name is caught, and that was a good call.

## Limits on the verdict

- No GPU and no dispatch, so every number here is CPU-local, against real uploaded logs and synthetic inputs. **I cannot confirm the racy reproducer still genuinely races under Sampled**, only that the log it produced parses to `fail` for the right reason. What would settle it is a dispatch on this head, or the negative control: the same reproducer built race-free and confirmed to still report `pass`, which distinguishes "the race is detected" from "this case reports `fail` under Sampled".
- The integrity-counter semantics are inferred from naming, as described above.
- `test_consan.py` is 66 passed and `tests/instrumentation/rocjitsu_sanitizers` plus `tests/sanitizers` is 646 passed / 3 skipped on this head, against 42 and 621 / 3 on `origin/main` (py3.11). The body's 57 and 637 / 3 do not reproduce at any of the three commits; they look like a reading from an earlier state. Not a problem in itself, the real counts being higher, but the numbers as written are not checkable.
- All checks now pass, including `pytest (GPU, MI350)` and `workload regression (GPU, MI350)`, and `mergeStateStatus` has moved from `BLOCKED` to `CLEAN` — so this can merge as soon as it is approved. #502 landed as `d2fa5f1` and is already the parent of this branch's first commit, so the dashboard dependency is satisfied and merge order no longer arises.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
amd-vivekag: Thanks — fixed in `69773970`. Added a parametrized mutation test over all four `_SAMPLED_SUMMARY_COUNTS`, deleting each field independently and asserting the named counter causes `ERROR`. This catches narrowing `_required_int` enforcement to only one field rather than merely deleting the whole counter block.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
amd-vivekag: Thanks for the careful mutation pass — confirmed and fixed in `69773970`.

I traced the counters through RocJITsu's `evaluate_auto_moi_report_trust`. Seven of the ten you listed already feed `dynamic_incomplete`; I now also require them directly from the per-reader summary because that and the aggregate verdict are separate log records. I additionally fail closed on `sampled_static_mapping_malformed`, which upstream currently omits from `dynamic_incomplete` but cannot be trusted for safe attribution.

I left two non-fatal, with tests and the upstream reason pinned:

- `sampled_saturated_windows`: `USAGE.md` explicitly says saturation is bounded sampling and is “reported separately from loss”.
- `sampled_pending_acquire_malformed`: it makes `synchronization_evidence_complete=false`; the analyzer then refuses ordering suppression, biasing toward extra findings rather than hiding one.

`sampled_changed_snapshots` is not benign — upstream includes it in `sampled_unusable_snapshot_count()` — so it is in the failing set. A parametrized mutation test now sets each of the eight incomplete counters nonzero and requires `ERROR`; two separate cases hold the conservative exceptions.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
amd-vivekag: Thanks — fixed in `69773970`. Added the mirror case: one Record/Replay detail shares a reader with a Sampled summary that claims one example but has no Sampled detail. The Sampled summary-only finding must still appear. Together the two tests now pin both directions of the per-engine split.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
amd-vivekag: Thanks for probing this. I checked the current upstream tree: `visible_sampled=` is emitted only by the actual Sampled report renderer (plus docs/tests); the `REQUIRE_DIAGNOSTICS` rejection does not carry it. So there is no real non-report fixture with that marker to add without inventing an unsupported log shape.

I did update the comment in `69773970` to remove the now-stale rejection-message rationale. I am keeping the `auto report` half as a record-kind boundary against future diagnostics, while the tested fail-closed properties remain independent of the exact marker: truncation before the counters errors through `_required_int`, and loss before the marker or of the whole line errors through applicable-reader reconciliation. Happy to add a fixture if upstream introduces a second real `visible_sampled=` record.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
amd-vivekag: Thanks — `RJ_CONSAN_MOI_REQUIRE_DIAGNOSTICS` is not harmless for this gate: inherited `=1` turns the clean control into an error. The proactive sweep also found `SAMPLED_CHECK`, REQUIRE/FORBID controls, report-buffer sizing, conflict limits, and tracking controls that can change evidence or outcomes.

Fixed in `69773970` by scrubbing the entire inherited `RJ_CONSAN_MOI_*` namespace before setting the one gate-owned value, `RJ_CONSAN_MOI_SAMPLED_PRESET=max`. That also future-proofs the gate against new MOI controls. The test sets every current upstream risk plus a synthetic future-prefixed variable and verifies only the pinned preset reaches the hook.

recipes/sanitizers/fixtures/expected/verdict_baselines.json
amd-vivekag: Thanks for checking the trade-off explicitly. This is deliberate: the gated contract is “the known racy control produces a Sampled race signal”, including the valid summary-only and device-immediate forms, rather than “it must retain an itemized host-analysis example”. The dashboard keeps the three codes separate for diagnosis, while the baseline stays robust when the bounded example limit is exhausted.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
amd-vivekag: Good catch — fixed in `69773970`.

The mode-neutral parser still accepts legacy Record/Replay logs when called directly, but the execution path now passes `expected_mode=ConSanMode.SAMPLED` and requires the coverage records to contain exactly that engine. I tested `record_replay`, `inline_shadow`, and `supercollider`; all now return `consan_output_parse_error` instead of a clean result, while a Sampled record passes.

This keeps saved-log compatibility without allowing the hook invoked by `run_consan` to ignore the mode aorta pinned.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -29,6 +29,46 @@
...
+#
+# Keyed on ``visible_sampled``, which the renderer writes second, rather than on
+# a conflict counter near the end of the line: a line truncated between the two
+# then still reports as a Sampled summary and fails closed on the missing
+# counters instead of going unrecognized. It cannot key on the first field
+# (``sampled_watchpoints``) because the ``auto report plan`` allocation line
+# carries that name too and no counters, so every healthy run would fail. Any
+# truncation earlier still than this is caught by the reconciliation against the
+# coverage records in _require_sampled_summaries.
+_AUTO_SAMPLED_SUMMARY = re.compile(r"\bauto\s+report\b.*\bvisible_sampled=", re.IGNORECASE)
```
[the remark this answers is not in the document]
amd-vivekag: Good catch — fixed in `f7f60d74`.

The parser now recognizes every actual per-snapshot record from its prefix, `auto report reader=`, before any Sampled field. After coverage identifies which readers ran Sampled, every such record for those readers is passed through the complete required/integrity counter validation. A valid summary can therefore no longer mask a truncated sibling.

The exact prefix still excludes the sibling `auto report plan`, `buffer`, and `cleanup` lifecycle records. Added the reported same-reader shape as a regression test: one complete summary plus a second line truncated before `visible_sampled=` must return `ERROR`. The stronger reconciliation also exposed an old synthetic two-reader test with no reader-2 coverage record; I corrected that fixture rather than weakening the production check.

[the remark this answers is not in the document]
amd-vivekag: Good catch — fixed in `f7f60d74`, then superseded by the stronger current-contract handling in `60ae7d65`.

The parser now recognizes every actual per-snapshot record from the stable `auto report reader=` prefix, before any mode-specific field. Coverage then identifies which readers use the retained default detector, and every report record for those readers must satisfy the complete current or legacy counter schema. One valid snapshot can no longer mask a truncated sibling.

While validating this on GPU, the latest published RocJITsu bundle exposed an upstream change that landed after the previous run: Record/Replay and the explicit Sampled mode were removed in `5d5a0f1`, and the grammar/env namespace was flattened. `60ae7d65` therefore also pins `mode=default` plus `RJ_CONSAN_PRESET=max`, parses current `mode=default` / `ConSan auto report` / `ConSan conflict` records, and retains the old schema only for saved logs. The same-reader valid-plus-truncated case remains pinned as `ERROR`.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -29,6 +29,40 @@
...
+_SAMPLED_SUMMARY_COUNTS = (
+    "sampled_conflict_examples",
+    "sampled_conflict_pairs_without_example",
+    "sampled_conflicts",
+    "sampled_immediate_conflicts",
+)
+# These mean evidence was lost, unusable, or cannot be attributed safely.
+# RocJITsu folds all but static_mapping_malformed into dynamic_incomplete; check
+# the summary independently as well because the two are separate log records.
+_SAMPLED_INCOMPLETE_COUNTS = (
```
vivekkhandelwal1, line 56: `suppressed_uniform_write_conflicts` sits on the same record and is in neither tuple. Flipping it on the real clean log from this head's nightly leaves the gate green:

```
suppressed_uniform_write_conflicts=9  ->  state=ran  verdict=pass  findings=0
```

It differs in kind from the eight here, which is why I am not asking for a change: it does not report lost evidence, it counts conflicts that were **found and then suppressed**. Both routes to a non-zero value are already shut. Upstream `docs/consan/EXPERT_CONTROLS.md` at bundle `9d72e86f` says of `RJ_CONSAN_ALLOW_PROVABLY_SAME_VALUE_WRITE_RACES` that "The default is `0`; sampling presets do not enable it", and `_CONSAN_ENV_PREFIX` now scrubs the variable so it cannot be inherited.

What reading it would buy is turning that argument into a check. A non-zero value means a real conflict was detected and dropped before the verdict, and it is the one counter that would prove the scrub actually held on a machine nobody here configured. One more entry, fatal or warn, closes the reasoning loop instead of resting it on the default.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -157,12 +237,178 @@ def _parse_waitcheck(
...
+    for summary in summaries:
+        reader = summary.get("reader")
+        if reader is None:
+            raise ValueError("ConSan field reader is missing from a Sampled report summary")
+        _required_int(summary, "reader")
+        counts = {key: _required_sampled_int(summary, key) for key in _SAMPLED_SUMMARY_COUNTS}
+        incomplete = {
+            key: _required_sampled_int(summary, key) for key in _SAMPLED_INCOMPLETE_COUNTS
+        }
+        nonzero_incomplete = [key for key, value in incomplete.items() if value != 0]
```
vivekkhandelwal1, line 272: Not a requested change; the guard and both exclusions are right, and I verified them against upstream at the bundle commit this head actually ran rather than accepting the reply. `docs/consan/EXPERT_CONTROLS.md` at `9d72e86f`: "Fail if evidence was truly dropped. ConSan bounded saturation is reported separately from loss", and "A later valid identity after every bank fills is `saturated_windows`; malformed publication or true evidence loss uses separate counters and makes the analysis incomplete." Treating saturation as fatal here would have been the over-correction.

The residual worth a thought: saturation does mean a later window identity was not retained, and `docs/consan/MODES.md` lists bounded retention among the limits that "can hide conflicts". It is `0` on both controls today, in the archived logs as well as in your PR body. So if `saturated_windows` ever goes non-zero on `consan_racy`, the gate keeps reporting `fail` for the right reason while sensitivity has quietly narrowed, and nothing says so. A warn-severity finding on non-zero saturation would surface that without arming it as an error.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -157,12 +237,178 @@ def _parse_waitcheck(
...
+    The conflict counters are the only place a Sampled race with no logged
+    example is visible, so a log that lost them cannot be distinguished from a
+    clean run by the coverage records alone -- those describe static
+    instrumentation and stay healthy either way. Reconciling against coverage
+    keeps the absence itself fatal rather than silently unaccounted for. Only
+    applicable readers are required: a loaded object with no discovered site is
+    never instrumented and publishes no report, which is the ordinary shape of
+    the runtime helper objects that accompany a repro.
+    """
+    missing = sorted(
```
vivekkhandelwal1, line 305: The per-reader reconciliation closes the cases it was written for — I confirmed a sibling truncated before its counters, one with a non-integer counter, one with an inconsistent conflict count and one carrying `dropped_windows=7` all error, and so does losing every record. One narrower case survives: an entire `auto report reader=` record can be **deleted**, because a sibling record for the same reader keeps that reader in `totals` while the coverage records stay healthy.

The precondition is the ordinary shape rather than a contrived one — both real controls from this head carry exactly two records per reader, the second being the emptied post-reclaim snapshot. Reproduction on the real clean log, making the first record a *consistent* summary-only conflict (`conflicts=1`, `conflict_examples=0`, `conflict_pairs_without_example=1`, i.e. the form the updated racy baseline now accepts):

```
summary-only conflict present      ->  verdict=fail  findings=1  (sampled_conflict_summary)
...same log, that one line deleted ->  verdict=pass  findings=0
```

Two things bound it, and are why this is a note rather than a request. The log declares no snapshot count anywhere — `auto report plan` and `cleanup` carry allocation and lifecycle fields only — so there is nothing to reconcile a per-record count against without a new upstream field. And the `ConSan analysis verdict` record is the final line of both real logs (76 of 76, 78 of 78), so the realistic loss mode, a truncated tail, takes the verdict record with it and already errors.

Worth one clause in the docstring: the guarantee is per applicable reader, not per record.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -29,6 +29,40 @@
...
+# a race report.
+_AUTO_SAMPLED_CONFLICT = re.compile(r"\bauto\s+sampled\s+conflict(?:\s|$)", re.IGNORECASE)
+_DEFAULT_CONFLICT = re.compile(r"\bConSan\s+conflict(?:\s|$)", re.IGNORECASE)
+# The actual per-snapshot record begins ``auto report reader=``. Match that
+# prefix, before any mode-specific field: every such record belonging to a
+# Sampled reader must carry the complete Sampled schema, so one valid snapshot
+# cannot mask a second truncated one. The exact prefix excludes the sibling
+# ``auto report plan|buffer|cleanup`` records, which carry allocation/lifecycle
+# fields rather than evidence counters.
+_AUTO_REPORT = re.compile(r"\bauto\s+report\s+reader=", re.IGNORECASE)
```
vivekkhandelwal1, line 46: Narrowness, minor, recorded so the gap is known rather than as a request. This is the one of my six earlier mutations still not caught: rewriting it as

```python
_AUTO_REPORT = re.compile(r"auto\s+report\s+reader", re.IGNORECASE)
```

— dropping the `=` and the word boundary — leaves the suite at `667 passed, 3 skipped`.

It is benign on every log shape I can see, which is presumably why no test notices. It admits only sibling fields whose name begins with `reader`, and the current schema has exactly one; the `auto report plan|buffer|cleanup reader=` records still fail to match, because `reader` has to follow `report` directly. The comment above already states the exclusion that matters, and the four mutations that would genuinely widen this record class are all caught.

vivekkhandelwal1 (approved):
**Approving** at `60ae7d65fdddbfa982154d483a063f7620170077`.

Classic branch protection is disabled on this repository, so this approval is not dismissed by a later push. It is scoped to that SHA and to nothing after it.

This supersedes my CHANGES_REQUESTED on `e25a197` ([pullrequestreview-5238289988](https://github.com/ROCm/aorta/pull/505#pullrequestreview-5238289988)). The blocking finding there is fixed, the engine gap is closed, and no over-correction was introduced. **No blocking findings remain.** Four non-blocking notes are inline.

The head moved twice while I was working, so to be unambiguous: my findings cover `60ae7d6`, not `f7f60d7`. That matters more than usual here, because `f7f60d7` was broken on hardware and `60ae7d6` is the fix.

### The integrity counters are enforced

Re-ran the ten-counter flip probe, now against the real gfx950 `consan-clean` log captured at this head (nightly [35318482298](https://github.com/ROCm/aorta/actions/runs/35318482298), bundle `9d72e86f`), one counter at a time from `0` to `9`, through `evaluate_consan_output` with the mode the gate pins:

| counter | verdict |
|---|---|
| `dropped_windows` | `error` |
| `stale_snapshots` | `error` |
| `incomplete_snapshots` | `error` |
| `changed_snapshots` | `error` |
| `malformed_snapshots` | `error` |
| `malformed_sync` | `error` |
| `static_mapping_malformed` | `error` |
| `unsupported_sync` | `error` |
| `saturated_windows` | `pass` (excluded by design) |
| `pending_acquire_malformed` | `pass` (excluded by design) |

Eight of the ten now fail closed where all ten previously reported `pass`. Controls confirm the harness was live: `immediate_conflicts=4` gives `fail` with one finding, `conflicts=4` gives `error` on the consistency check, and deleting `conflict_examples` or `conflict_pairs_without_example` gives `error` naming the missing field.

I did not take the two exclusions on trust. Both claims check out against upstream at the exact bundle commit this head ran, `9d72e86f`: `docs/consan/EXPERT_CONTROLS.md` states "Fail if evidence was truly dropped. ConSan bounded saturation is reported separately from loss", and, on the counter itself, "A later valid identity after every bank fills is `saturated_windows`; malformed publication or true evidence loss uses separate counters and makes the analysis incomplete." Making saturation fatal would have been the over-correction. `changed_snapshots` being treated as non-benign is also right, and it is in the failing set.

### No over-correction, confirmed on hardware and not only locally

The real clean log evaluates to `pass` with zero findings and the real racy log to `fail` with its one attributed `ConSan conflict`, unmodified, at this head. Independently, the gate job of nightly [35318482298](https://github.com/ROCm/aorta/actions/runs/35318482298) on `60ae7d6` succeeded with `consan_clean: ok (pass)` and `consan_racy: ok (fail)` — byte-identical control verdicts to the `main` dispatch [35237696462](https://github.com/ROCm/aorta/actions/runs/35237696462).

### Worth recording: `f7f60d7` was vacuous on hardware, and the design failed closed

The intermediate head was red on the thing it gates, and the mechanism is instructive rather than a reproach. Nightly [35315430164](https://github.com/ROCm/aorta/actions/runs/35315430164) on `f7f60d7`:

```
consan_clean: overall_verdict='error', expected 'pass'
consan_racy:  overall_verdict='error', expected 'fail'
reason 'consan_output_parse_error: missing ConSan coverage record' ... no sanitizer signal at all
```

Its archived `consan-clean/consan/consan.log` is two lines, and the hook says why verbatim:

```
[rocjitsu-dbi-hooks] invalid RJ_CONSAN_MODE='sampled'; expected default|supercollider
```

Two things follow. The dependency hazard raised on `e25a197` — the bundle resolved as latest-successful with no commit pin — is no longer hypothetical; it fired, mid-review, and removed the mode name the gate was pinning. And when it fired the gate went **red, not falsely green**, which is the property this PR exists to establish. A commit pin on the bundle is still worth having, and is out of scope here.

### Verified positively

- **Engine predicate closed.** Every non-pinned engine now errors: `record_replay`, `sampled`, `inline_shadow` and an unrecognised value are rejected by the coverage grammar as `unsupported mode`; `supercollider`, which is a real current mode, is rejected by `_require_expected_engine` as "expected the pinned engine default". Previously each of these was a clean `pass`.
- **Per-snapshot validation holds.** On the real two-record logs: a sibling truncated before its counters, a sibling with a non-integer counter, a sibling carrying an inconsistent conflict count, and a sibling carrying `dropped_windows=7` all error, as does losing every record. One narrower case survives and is inline.
- **Legacy compatibility is real, and scoped correctly.** The pre-migration `engine=record_replay` logs still parse for direct callers (`pass` on clean, `fail` with 64 findings on racy), while the same logs are rejected on the gate path because it pins the current mode. That is the right way round.
- **The namespace scrub is safe.** I audited every `RJ_CONSAN_*` reference in the tree: nothing outside tests sets one, and each `RJ_CONSAN_REQUIRE_RECORDS` mention is a comment noting the hook derives it from `POLICY=strict`. The growth-cap variables appear only in docs and a repro script, so widening the scrub strips nothing in-tree that was relied on.
- **Five of the six narrowness mutations I reported are now caught**: requiring only `sampled_conflicts`, crediting replay details against sampled summaries, the predicate rewritten as `!= "record_replay"`, dropping the `auto report` half of the record regex, and one of the two anchor variants. Three fresh mutations against the new integrity guard are also caught: requiring only a subset of the counters, keying the guard on the conflict counters instead of the integrity ones, and inverting its comparison. The sixth old mutation is still missed and is inline.
- **Counts reproduce.** `tests/instrumentation/rocjitsu_sanitizers tests/sanitizers` gives `667 passed, 3 skipped` on this head against `621 passed, 3 skipped` on `origin/main` (`d2fa5f1`), +46 tests, green both sides, each worktree with its own venv so the editable install cannot cross-resolve. `test_consan.py` alone gives `87 passed`. Both figures in the PR body are exact; the previous pass could not reproduce the figures stated then.

### Limits on this verdict

- Every parser result above is log replay on CPU. I ran nothing on a GPU and dispatched nothing.
- Hardware provenance, which was an open limit last pass, is now satisfied: `pytest (GPU, MI350)` ran and passed on this exact head, and so did the gate job of the nightly dispatched on it. I did not dispatch either.
- `workload regression (GPU, MI350)` was still `QUEUED` when I posted; the other 16 checks are `SUCCESS`. `mergeStateStatus` reads `BLOCKED` solely because of my own CHANGES_REQUESTED on `e25a197` — it is the only non-comment review on the PR, and no check is failing.
- The two conservative counter exclusions rest on upstream prose at bundle `9d72e86f`, which I read; I did not exercise either counter non-zero on hardware, and no archived log has one non-zero.
- The PR title still says "make ConSan Sampled the guarded default" while the change now pins `mode=default`. The body explains the pivot, but the title is what survives in history.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
```diff
@@ -656,9 +941,14 @@ def run_consan(
...
-    # no auto-registration, record/replay mode, and the requested policy.
+    # no auto-registration, the retained default detector at the ``max`` preset,
+    # no inherited ConSan controls, and the requested policy.
     env["HSA_TOOLS_DISABLE_REGISTER"] = "1"
-    env["RJ_CONSAN_MODE"] = ConSanMode.RECORD_REPLAY.value
+    for name in tuple(env):
+        if name.startswith(_CONSAN_ENV_PREFIX):
+            env.pop(name)
+    env["RJ_CONSAN_MODE"] = ConSanMode.DEFAULT.value
+    env["RJ_CONSAN_PRESET"] = _CONSAN_SAMPLED_PRESET
```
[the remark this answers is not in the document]
amd-vivekag: Thanks — this was a real moving dependency, and the latest-successful artifact changed again while the PR was under review.

Before #505 merged I updated its body to artifact run [35300134760](https://github.com/ROCm/rocm-systems/actions/runs/35300134760) at `9d72e86f`, which contains the `5d5a0f1` mode/naming simplification, and reran the current head against it: [sanitizer nightly 35318482298](https://github.com/ROCm/aorta/actions/runs/35318482298) passed the gating job with waitcheck=`warn`, default-detector clean=`pass`, and racy=`fail` with one current `ConSan conflict`. The merged code pins `mode=default` plus `RJ_CONSAN_PRESET=max`; the old artifact claim is no longer in the PR body.

src/aorta/instrumentation/rocjitsu_sanitizers/consan_coverage.py
```diff
@@ -252,7 +269,16 @@ def _parse_verdict(payload: str, line_number: int) -> AnalysisVerdict:
...
     )
+    legacy_present = [name in fields for name in _LEGACY_VERDICT_COUNTS]
+    if any(legacy_present) and not all(legacy_present):
+        raise CoverageParseError(f"{context}: incomplete legacy replay counters")
     counts = {name: _count(fields, name, context) for name in _VERDICT_COUNTS}
+    counts.update(
+        {
+            name: _count(fields, name, context) if all(legacy_present) else 0
+            for name in _LEGACY_VERDICT_COUNTS
+        }
```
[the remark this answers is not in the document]
amd-vivekag: Good catch. #505 merged while I was preparing this fix, so it is now isolated in follow-up [#508](https://github.com/ROCm/aorta/pull/508) (`7e3c20d7`).

Coverage and verdict records now retain explicit `current` versus `legacy` provenance, and the parser requires their observed schema sets to agree. A current `mode=default` record may omit the removed replay counters; a legacy `flavor`/`engine` record paired with a verdict missing all four now fails with `coverage/verdict schema mismatch` instead of supplying zeros. Added the exact all-four-deleted regression; 146 focused tests pass.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
amd-vivekag: Thanks for closing the reasoning loop. I agreed this is worth enforcing and added it to follow-up [#508](https://github.com/ROCm/aorta/pull/508) in `d61fa29b`.

Current flattened reports must contain `suppressed_uniform_write_conflicts`, and any nonzero value is now a parse error: live execution clears the only expert opt-in that permits suppression, so nonzero means the pinned contract was not honored. Legacy Sampled reports predate the field and remain compatible. Added the exact `=9` regression; the current green MI350 clean/racy logs still parse `pass`/`fail`.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
amd-vivekag: Thanks — added the boundary explicitly in follow-up [#508](https://github.com/ROCm/aorta/pull/508), commit `b2694541`.

The docstring now states that the presence guarantee is per applicable reader, not per snapshot: aorta requires at least one complete report for each applicable reader and validates every report record it sees, but RocJITsu publishes no snapshot-count field, so deletion of an additional whole record is not observable. This removes the stronger claim without pretending the uncheckable case is covered.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
amd-vivekag: Thanks for checking this against the exact bundle. I agree there is an observability trade-off, but I am leaving saturation non-fatal and without a finding in this change: every `Finding` currently makes the ConSan verdict `fail` regardless of severity, so a “warn” record would turn the clean baseline red rather than merely annotate narrowed sensitivity.

The explicit non-fatal test pins upstream's distinction between bounded saturation and true evidence loss. If we want a visible advisory, it needs a separate persisted diagnostic channel whose presence does not change `pass` to `fail`; that is broader than this parser-hardening follow-up. Happy to track that separately if you want it surfaced on the dashboard.

src/aorta/instrumentation/rocjitsu_sanitizers/consan.py
amd-vivekag: Thanks for mutation-checking the narrow boundary. I agree this is benign: `auto report reader` still excludes every meaningful sibling (`plan`, `buffer`, `cleanup`), and the current schema has no other field beginning with `reader`. I am keeping the stricter `reader=` spelling because it states the grammar precisely, but I have not added a test whose only observable is accepting a hypothetical `readerSomething` field that upstream does not emit.

PR #507, opened 2026-09-18
fix(chat): keep every job in a decision record joinable, and its verdict its own

vivekkhandelwal1:
`decision_log.py` does what it says, and the parts that are hard to get right — off by default, the null `resolution` as an attachment point, scrubbing the selector's reason whatever `redact` says, `O_NOFOLLOW` and a digested filename — are all there. This is a small follow-up on top of it: two defects in the CIA extraction, and four fields.

Both defects came out of running `_cia_results` against real tool output rather than reading it, which is most of why I trust them.

### A result naming two jobs gives the second one the first one's verdict

This is the one that matters. `_CATEGORY` and `_CONFIDENCE` are searched across the whole output and then attached to every id found in it:

```
>>> _cia_results(
...   "Job cia-…-4fd02e (slurm 1)\n  category:   gpu_race\n  confidence: 0.82\n"
...   "Job cia-…-9ab31c (slurm 2)\n  category:   numeric_instability\n  confidence: 0.41")
[{'job_id': 'cia-…-4fd02e', 'category': 'gpu_race', 'confidence': 0.82},
 {'job_id': 'cia-…-9ab31c', 'category': 'gpu_race', 'confidence': 0.82}]
```

A job whose verdict is missing is simply absent from whatever reads the log. A job wearing another job's category is an answer, and a wrong one. They are now read from the span between one id and the next.

### A job listing yields no ids at all

`_JOB_ID` wants a `Job ` prefix or a `"job_id":` key, and `list_cluster_jobs` prints neither:

```
>>> _cia_results(
...   "  cia-…-4fd02e  recipe=kernel  status=completed  verdict=gpu_race @ 0.82\n"
...   "      bundle: /home/me/cia-jobs/cia-…-4fd02e/bundle")
[]
```

It is the tool most likely to have run, since reading past jobs needs no `allow_cluster_jobs`, so those turns record a tool call with nothing joinable in it. The pattern now accepts a bare id at the start of a line as well — anchored rather than free-floating, because the bundle path on the following line repeats the same id and would otherwise read as a second job. Its `verdict=x @ y` rendering is still not parsed, so the category stays null there, and that is deliberate: the id is what makes the row joinable, and the report under `jobs_root` is a better source for a verdict than a rendered line.

### Four fields, and what each is for

- **`jobs_root`** — beside the ids on the tool event, and only where there are ids. Without it a recorded id resolves against nothing: the bundle, the autopsy report and the probe cells that would say what actually happened all live under that root, and no other field names it. It is what an empty `resolution` is waiting on.
- **`duration_seconds`** — on the event that ends a turn, and on a failure. A four-minute turn and a four-second one are different decisions even when they record the same ones.
- **`front_door`** — on the envelope, so a browser demo and a scripted CLI run are told apart rather than averaged together. Null unless the caller says, since nothing below the entry point can tell a browser from a terminal.
- **`iterations`** — on the critic event. `accepted` says the answer was taken; this says whether it was taken first time or on the third attempt.

### Tests

12 added. Eleven fail on `a78c154` and pass here, including both reproductions above; the twelfth guards the de-duplication that already worked. `tests/chat` goes from 2109 to 2121 passing with the same nine failures either way — all optional extras missing from my environment rather than anything in the code.

One of your tests changed: `test_ask_once_forwards_the_session_and_turn` asserts the forwarded kwargs exactly, and there is now a third one.

### Taking it

Built on `a78c154`. Take it, rewrite it, or cherry-pick the two fixes and drop the fields — the fixes are yours to own however you prefer, and the four fields are for our side of this, so they are the part to push back on. Happy to split it into two PRs if that reviews better.

Unrelated and not a request: `tests/chat/test_triage_threads.py:172` still hard-codes `/apps/avsharma/aorta/src` into the subprocess path. `0f1f38f` fixed the triage CLI tests; this one is still machine-specific and would fail for anyone off that mount.


src/aorta/chat/decision_log.py
```diff
@@ -172,26 +178,76 @@ def _parse_tool_trace(entry: Any) -> tuple[str, str, str]:
...
+        job_id = match.group("id")
+        if job_id in seen:
+            continue
+        seen.add(job_id)
+        following = (
+            matches[position + 1].start()
+            if position + 1 < len(matches)
+            else len(output)
+        )
+        segment = output[match.end() : following]
```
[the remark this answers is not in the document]
vivekkhandelwal1: Confirmed by running the cited input rather than reading it, and fixed in e18ef0d. On

```
Job cia-20260918-064455-4fd02e started
Job cia-20260918-064455-4fd02e done
  category: gpu_race
```

`_cia_results` returned `{"job_id": "...4fd02e", "category": None, "confidence": None}`. The verdict is in the output, belongs to that job, and was discarded — the segment ended at the job's own second mention and the `seen` set then skipped it. The existing test asserted only the `job_id`, so it passed while the answer was lost. Squarely in scope: this PR exists to stop a verdict reaching the wrong job, and dropping one is the same defect facing the other way. The test now asserts the verdict, as you asked.

**Fixed slightly wider than suggested, and the difference is measured rather than argued.** A job now gets the union of every span that follows one of *its own* mentions, each still bounded at the next mention of any job — rather than "stop at the next distinct ID". The distinguishing input is an interleaved log:

```
Job A started
Job B started
Job A done
  category: gpu_race
```

Here the verdict follows a mention of A that is not A's first, so the next-distinct-ID rule bounds A's span at `Job B` and loses it again. I implemented that rule as a mutation to check rather than assume: it passes every other test in the file and fails only `test_a_job_named_again_after_another_job_keeps_its_verdict`.

**Two things found next to the reported site, which is what makes the wider rule worth the lines.**

1. **It needs no duplicate *announcement* at all.** `_JOB_ID` also matches a bare id at the start of a line — deliberately, because that is how `list_cluster_jobs` prints one — and the bundle path under a listing repeats the id on its own line. So a single, singly-announced job could truncate its own evidence, which makes this more than a property of the input you quoted. Covered by `test_a_bundle_path_repeating_the_id_does_not_truncate_the_verdict`.
2. **The change is strictly additive, and that is now pinned rather than claimed.** Spans are read in document order and the *first* value wins, so a job that already reported a verdict reports the same one: this can turn a `None` into a verdict and can never turn one verdict into another.

The narrowness control matters most here, since widening a read is how misattribution gets reintroduced: a span only ever begins at a mention of its own job and ends before the next job is named, so text following a mention of B is B's however many times A was named earlier. `test_widening_the_read_does_not_leak_a_verdict_backwards` fails if that stops holding.

**Evidence.** `tests/chat` 2130 passed / 132 skipped / 2 xfailed → **2134 / 132 / 2**, same skips and xfails. Four mutations, all caught: the shipped parser (3 tests), the next-distinct-ID rule (1), last-value-wins (1), and spans running to the end of the output (2). Worth recording that last-value-wins was **not** caught on the first pass — the additive-property test used the same verdict twice, which cannot tell first from last apart. It now uses two different ones.

**One thing for you rather than for me, since it is about the base branch.** `feat/cia-reporting` has moved from `a78c154` to `290d4e2c` since this PR's merge base, and it touches `decision_log.py`: `selection.reason` changed from the ungated `_scrub_reason(...)` to the mode-gated `_content(..., mode)`. I checked a three-way merge of both files read-only rather than rebasing — **it is clean, and semantically coherent**: the base's `_content` wins, `_scrub_reason` and its `scrub_text` import drop out, this PR's `front_door` argument survives on the adjacent `_base(...)` line, and no stale comment is left behind. Nothing to do; noting it because the collision would have been invisible in a file list.

src/aorta/chat/decision_log.py
```diff
@@ -37,8 +37,14 @@
...
 _JOB_ID = re.compile(
-    r"(?:\bJob\s+|[\"']job_id[\"']\s*:\s*[\"'])"
-    r"(?P<id>(?:cia|nan)-[A-Za-z0-9._-]+)"
+    # A bare id at the start of a line is how ``list_cluster_jobs`` prints one,
+    # and that tool needs no extra permission, so it is the one most likely to
+    # have run. Still anchored rather than free-floating: the bundle path on
+    # the following line repeats the same id, and an unanchored match would
+    # read it as a second job.
+    r"(?:^[ \t]*|\bJob\s+|[\"']job_id[\"']\s*:\s*[\"'])"
+    r"(?P<id>(?:cia|nan)-[A-Za-z0-9._-]+)",
```
amd-vivekag: **[High]** The bare line-start branch invents a `nan-trap` job out of rocGDB output.

`(?:cia|nan)-[A-Za-z0-9._-]+` at a line start matches `nan-trap value mean_sq=0`, which `src/aorta/cia/autopsy/adapters/rocgdb.py` emits. Run against the PR head:

```
>>> _cia_results("nan-trap value mean_sq=0\nnan-trap value max_abs=inf\n")
[{'job_id': 'nan-trap', 'category': None, 'confidence': None}]

>>> _cia_results("  nan-trap value mean_sq=0\n  category: gpu_race\n  confidence: 0.9\n")
[{'job_id': 'nan-trap', 'category': 'gpu_race', 'confidence': 0.9}]
```

The second one is a phantom job wearing a real verdict, which is the error direction this PR says is the one worth spending code on. It also puts `jobs_root` on a tool event that named no job, so the "only where there are ids" rule the new `configuration.md` note relies on no longer holds. Copilot flagged this too.

Fix: constrain the bare branch to the real id shape (`new_job_id()` is `cia-%Y%m%d-%H%M%S-<6 hex>`) with a lookahead, leaving the prefixed and JSON forms loose. I ran this against every input in `TestEveryNamedJobIsFoundAndKeepsItsOwnVerdict` and all of them still pass, and `nan-trap` returns `[]`.

```suggestion
    r"(?:^[ \t]*(?=(?:cia|nan)-\d{8}-\d{6}-[0-9a-f]{6}\b)"
    r"|\bJob\s+|[\"']job_id[\"']\s*:\s*[\"'])"
    r"(?P<id>(?:cia|nan)-[A-Za-z0-9._-]+)",
```

src/aorta/chat/decision_log.py
```diff
@@ -172,26 +178,107 @@ def _parse_tool_trace(entry: Any) -> tuple[str, str, str]:
...
+    next *distinct* id instead would fix the announce-then-finish case and
+    still lose an interleaved one (``Job A`` / ``Job B`` / ``Job A done``),
+    where the answer sits after a mention of A that is not A's first.
+
+    A verdict that is genuinely absent stays ``None``. The job id is what makes
+    the row joinable; ``jobs_root`` on the same event is what the id resolves
+    against, and the report on disk is a better source for a verdict than a
+    rendered string in any case.
+    """
+    matches = list(_JOB_ID.finditer(output))
```
amd-vivekag: **[Medium]** A verdict that appears *before* its job id is now lost, and with two jobs it is misattributed — the exact failure this PR removes.

Spans start at an id, so nothing before the first id is read. `_CATEGORY`, `_CONFIDENCE` and `_JOB_ID` all carry quoted-key alternatives for JSON renderings, and `json.dumps(..., sort_keys=True)` puts `category` and `confidence` ahead of `job_id`. On the PR head:

```
>>> _cia_results(json.dumps([{'job_id':'cia-a1','category':'gpu_race','confidence':0.82},
...                          {'job_id':'cia-b2','category':'numeric_instability','confidence':0.41}],
...                         sort_keys=True, indent=2))
[{'job_id': 'cia-a1', 'category': 'numeric_instability', 'confidence': 0.41},
 {'job_id': 'cia-b2', 'category': None, 'confidence': None}]
```

`cia-a1` gets `cia-b2`'s verdict. The old whole-output search got `cia-a1` right. I could not find an in-tree tool that emits this today — `report.json` keeps `bundle.job_id` ahead of `category` and `json.dumps` there has no `sort_keys` — so this is a latent regression rather than a live one, but the quoted-key branches exist because that rendering is expected.

Fix: when exactly one id was found and its spans yield no verdict, fall back to searching the whole output. That recovers the single-object JSON case and cannot misattribute, because there is only one job to attribute to.

docs/chat/configuration.md
```diff
@@ -85,15 +85,39 @@ they capture the same decisions without changing the CLI to a streaming path:
...
   scrubbed regardless of `--no-redact`;
-- tool execution order and CIA job ID/category/confidence;
-- whether the critic accepted the answer; and
+- tool execution order, and one row per CIA job with its ID, category and
+  confidence — a turn that starts three jobs records three rows, each carrying
+  its own verdict;
+- `front_door`, which entry point the turn came through (`cli` or `ui`), so a
+  browser demo and a scripted run are told apart rather than averaged;
+- `duration_seconds` for the whole turn, and `iterations` for how many critic
+  rounds the answer took; and
```
amd-vivekag: **[Low]** This rewrite dropped `accepted` from the list of what summary mode retains.

`turn_events` still writes `"accepted": accepted` on the critic event, but the old bullet "whether the critic accepted the answer" is gone and `iterations` replaced it rather than joining it.

Fix:
```suggestion
- `duration_seconds` for the whole turn, `iterations` for how many critic
  rounds the answer took, and whether the critic accepted the answer; and
```

tests/chat/test_decision_log.py
```diff
@@ -413,6 +413,341 @@ async def __call__(self, _node: str, _delta: dict) -> None:
...
+        """
+        monkeypatch.setenv("AORTA_CHAT_JOBS_PATH", str(tmp_path / "cia-jobs"))
+        from aorta.chat import config
+
+        config.reset_settings()
+
+        tool_event = self._one(monkeypatch, "tool")
+
+        assert tool_event["jobs_root"] == str(tmp_path / "cia-jobs")
+        config.reset_settings()
```
amd-vivekag: **[Low]** Cleanup runs after the assert, so a failure here leaks the patched settings singleton into every later test.

`test_the_root_is_raw_in_summary_mode_and_the_docs_say_so` right below already uses `try`/`finally` for the same reason.

Fix:
```suggestion
        try:
            tool_event = self._one(monkeypatch, "tool")
        finally:
            config.reset_settings()

        assert tool_event["jobs_root"] == str(tmp_path / "cia-jobs")
```

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

The span-per-job rewrite is the right shape and the test class around it is unusually good — the interleaved `A / B / A done` case and the first-wins narrowness control are the two tests that actually decide the rule, and both are there. The `jobs_root` disclosure note is the right call. Two problems in the widened `_JOB_ID`, both found by running it rather than reading it.

**Findings:** 1 High · 1 Medium · 2 Low

**High — must fix before merge**
- `src/aorta/chat/decision_log.py:45` — the bare line-start branch matches `nan-trap value mean_sq=0` from the rocGDB adapter and records a phantom job `nan-trap`, which will pick up a nearby `category:`/`confidence:` and also drags `jobs_root` onto tool events that named no job. Suggestion inline, verified against all of this PR's own test inputs. (Copilot's "previously missed" item — confirmed.)

**Also**
- Medium: a verdict rendered *before* its id (sorted-key JSON) is now lost, and with two jobs in one blob the first job gets the second's category. The old code got that first job right.
- Low: `docs/chat/configuration.md` no longer lists `accepted`, which is still written.
- Low: a test leaks patched settings on failure.

Nothing to say about the privacy reasoning on `jobs_root` — the note is clear about what is retained and why a digest would not do, and the test pinning the code and the disclosure together is a good pattern. Worth fixing the High before that note becomes inaccurate: a phantom job means `jobs_root` lands on events the note says it will not.

src/aorta/chat/decision_log.py
vivekkhandelwal1: Fixed in f7c749bd, taking your lookahead as written.

The bare branch now requires the full `new_job_id()` shape — `(?=(?:cia|nan)-\d{8}-\d{6}-[0-9a-f]{6}\b)` — while the `Job ` and `"job_id":` branches stay loose, for the reason you gave: there the surrounding text is the evidence that an id follows, so a shortened or hand-written id is still a job. At a line start there is no such context, so the token has to look like the generator produced it.

Your two checks reproduce: `nan-trap value mean_sq=0` with a category and confidence under it now returns `[]`, and every input in `TestEveryNamedJobIsFoundAndKeepsItsOwnVerdict` still passes.

Three tripwires, in `TestOnlyARealIdCountsAsAJobBeingNamed`:
- `test_a_line_that_merely_starts_with_nan_is_not_a_job` — your input, asserting `[]`.
- `test_a_bare_id_at_a_line_start_is_still_a_job` — the narrowness control. A listing prints its ids bare and indented; if the lookahead were slightly too strict that whole shape would go dark, which is the defect the line-start branch was added for in the first place.
- `test_a_prefixed_id_that_is_not_the_generated_shape_is_still_read` — pins that the lookahead is the price of a *bare* token and not of every id, so a future tidy-up does not "simplify" it onto the other two branches.

Mutation-tested: reverting the lookahead to `r"(?:^[ \t]*"` fails `test_a_line_that_merely_starts_with_nan_is_not_a_job` and nothing else.

Worth naming the second half of what you found, since it is the part that reaches outside this function: the phantom job also put `jobs_root` on a tool event that named no job, so the "only where there are ids" rule `configuration.md` describes had stopped holding. That is now a consequence of the id test rather than a separate guard.

src/aorta/chat/decision_log.py
vivekkhandelwal1: Fixed in f7c749bd — and deliberately wider than you suggested, so here is the reasoning in case you disagree with the widening.

Your fallback is right and it is in: when exactly one id is found and its forward spans yield no verdict, the whole output becomes the span. It cannot misattribute because there is one job to attribute to.

But it only covers half of what you reported. Your own note says the two-job sorted dump is *misattributed* — both jobs take the first verdict — and a whole-output fallback cannot fix that, because with one text and two jobs there is nothing to tell whose verdict is whose. Widening the read there would reintroduce exactly the misattribution this PR removes. So the ordering problem is solved structurally instead: `_cia_results_from_json` parses the output and takes each object carrying a valid `job_id` whole. Key order stops mattering because nothing is being scanned for.

Four properties it holds, each pinned:
- An object with an id is taken whole and never merged with its parent's fields, so a summary `category` above a `results` list is not copied onto every job under it.
- `category` and `confidence` go through the same vocabulary the line patterns accept. Being JSON says the output is machine-written, not that `{"confidence": "high"}` is a confidence.
- Ids are deduplicated, matching the line reader.
- It returns `None`, not `[]`, when the output is not JSON-with-ids — otherwise any JSON output would suppress the line reader, and a tool returning `{"stdout": "<rendered log>"}` has its ids inside a string where only the line reader can see them. That one had no tripwire until a mutation found it; it has one now.

Your fallback stays underneath for the renderings that are not JSON at all, restricted to the single-job case.

Six tests in `TestAVerdictPrintedBeforeItsIdStillLands`, covering the sorted single object, the sorted two-object dump, nesting, wrong-typed values, the plain-text fallback, and the bound on it (two plain-text jobs with both verdicts above both ids report `None` twice rather than guessing).

Mutations, all red at the named test:
- pre-pass removed → `test_two_jobs_dumped_with_sorted_keys_do_not_swap_verdicts`
- fallback removed → `test_one_job_in_plain_text_recovers_a_verdict_printed_above_it`
- fallback widened to any job count → `test_two_jobs_in_plain_text_do_not_fall_back`, plus two of the existing misattribution tests
- JSON value vocabulary dropped → `test_a_json_field_of_the_wrong_type_reads_as_no_verdict`
- object merged with parent → `test_a_nested_object_does_not_inherit_a_parent_verdict`
- `return results` instead of `return results or None` → `test_json_that_names_no_job_does_not_suppress_the_line_reader`

One result worth reporting because it is evidence for your position and not mine: with the pre-pass removed, the *single*-object sorted dump still passes, because your fallback recovers it. The two-job test is the only thing that makes the pre-pass load-bearing.

docs/chat/configuration.md
vivekkhandelwal1: Fixed in f7c749bd. `accepted` is back in the summary-mode bullet list, beside `duration_seconds` and `iterations`:

> `duration_seconds` for the whole turn, `iterations` for how many critic rounds the answer took, and whether the critic accepted the answer; and

You are right that it mattered rather than being a tidy-up: the list is what a reader checks before turning summary mode on, and a field the log records but the list omits is an undisclosed retention — the same direction of error the rest of this PR is about, just in prose. It went missing in the conflict resolution against main, which is the merge this commit also carries.

tests/chat/test_decision_log.py
vivekkhandelwal1: Fixed in f7c749bd, taken as written — the `reset_settings()` call is now in a `finally`, matching the test below it.

One correction to the severity, in your favour on the principle and against it on the evidence. I tried to reproduce the cascade you describe: broke the assert, reverted the cleanup to the trailing-call form, and ran the whole of `tests/chat/`. The result was `1 failed, 2236 passed` — exactly the one broken test, no cascade. The reason is that every test in the class calls `config.reset_settings()` before it reads anything, so a leaked cache is overwritten before it can be observed.

So the hazard is latent, not live. I have still made the change, because the property you are describing — cleanup that runs only on the happy path — does not depend on today's test ordering holding, and the next test added to the file is not obliged to reset first. But I did not want to write a commit message claiming I had prevented a failure cascade when I could not produce one.

src/aorta/chat/decision_log.py
```diff
@@ -160,26 +189,215 @@ def _parse_tool_trace(entry: Any) -> tuple[str, str, str]:
...
+        if not isinstance(node, dict):
+            return
+        raw_id = node.get("job_id")
+        if not isinstance(raw_id, str) or not _JOB_ID_VALUE.fullmatch(raw_id):
+            for value in node.values():
+                walk(value)
+            return
+        if raw_id in seen:
+            return
+        seen.add(raw_id)
```
amd-vivekag: **[Medium]** The JSON pre-pass records Autopsy `report.json` as a job with no verdict.

`build_report()` puts the id under `bundle` and the verdict beside it. Walking that object as the result row takes `job_id`, leaves `category`/`confidence` as None, and never falls through to the regex — so a raw (or `sort_keys`) dump of the in-tree report is joinable and wrong. Prefixed `read_autopsy_report` output still works because `json.loads` fails on the header; this is the gap in the walker added for the last Medium. Copilot's latest pass did not catch it.

Fix: skip a dict that only looks like bundle metadata (`root` present, no verdict keys) and keep walking, so the regex (or a later object that actually carries the verdict) can see it. Pin it with `json.dumps(build_report(...))`.

```suggestion
        if raw_id in seen:
            return
        # report.json keeps the id under bundle and the verdict beside it.
        # Treating this dict as the result row records the id, drops the
        # sibling category/confidence, and never falls through to the regex.
        if "root" in node and "category" not in node and "confidence" not in node:
            for value in node.values():
                walk(value)
            return
        seen.add(raw_id)
```

amd-vivekag (commented):
## Review summary

**Verdict: fix-before-ship**

Re-review of f7c749bd. The last round's High and both Lows are fixed correctly, and the same-object `sort_keys` JSON case is too. The JSON pre-pass still treats Autopsy `report.json` as a job with no verdict: it records `bundle.job_id` and never reads the sibling `category` / `confidence` that `build_report()` actually writes.

**Prior comments**
- High (`nan-trap` bare id) — Fixed (correct)
- Medium (verdict before id) — Fixed (partial): same-object JSON is recovered; the in-tree report shape is not
- Low (`accepted` in the summary-mode list) — Fixed (correct)
- Low (`reset_settings` in `finally`) — Fixed (correct)

Copilot's latest pass recommended approval with no findings. The `report.json` miss is independent of that.

**Still open from last review:** the JSON walker added for the Medium still drops the verdict on the one JSON document this product emits.

src/aorta/chat/decision_log.py
amd-vivekag: Resolved — lookahead matches `new_job_id()` and `nan-trap` now returns `[]`. Thanks.

src/aorta/chat/decision_log.py
amd-vivekag: Partial — same-object `sort_keys` JSON is recovered, but Autopsy `report.json` still lands as a job with no verdict. Raised as a new Medium on the walker.

docs/chat/configuration.md
amd-vivekag: Resolved — `accepted` is back in the summary-mode list.

tests/chat/test_decision_log.py
amd-vivekag: Resolved — cleanup is in `finally`. Thanks for checking the cascade was latent.

src/aorta/chat/decision_log.py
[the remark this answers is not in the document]
amd-vivekag: Resolved — union of per-mention spans is the right rule, and the interleaved test pins it.

src/aorta/chat/decision_log.py
vivekkhandelwal1: Fixed as suggested in `64b0b4c9`.

Reproduced first: a real `build_report(...)` dump loses `gpu_race`/`0.82` in every render order, `sort_keys=True` included, because the claim happens at the first object that names a job rather than at the best one. `bundle` sorting before `category` is not what does it — renaming the locator key so it sorts *after* the verdict keys drops the verdict just the same.

Took the suggested guard verbatim. Declining and walking on, rather than merging the locator into its parent, because `test_a_nested_object_does_not_inherit_a_parent_verdict` exists to forbid that merge, and a fix that has to delete a guard is the guard being right.

Pinned with `json.dumps(build_report(...))` as asked, plus the narrowness controls for each conjunct, since a guard this shape is one careless widening away from swallowing real rows:

* `test_a_raw_report_dump_keeps_its_verdict` — parametrised over both key orders.
* `test_a_locator_key_that_sorts_after_the_verdict_still_lands` — `bundle` renamed to `source`, so the fix cannot be passing on key order.
* `test_the_locator_does_not_claim_the_output_from_the_line_reader` — the prefixed `read_autopsy_report` path you note still works; asserted rather than assumed.
* `test_a_bare_id_in_a_results_list_is_still_taken_verdictless` — `root` is what separates a locator from a genuinely verdictless result row, which stays verdictless.
* `test_locators_that_carry_their_own_verdicts_are_still_results` — a locator that grew a verdict is read as the row it has become.

That last one is worth flagging: my first version used one job and passed a mutant that dropped both verdict conjuncts (`if "root" in node:` alone), because with one job the whole-output fallback in `_cia_results` rescues a walk that declines everything. Two jobs disables the fallback, and the mutant now fails. All three conjuncts have a test that dies without them.

`tests/chat/test_decision_log.py`: 60 → 65 passing; full `tests/chat` green (2243 passed, 122 skipped, 2 xfailed).

vivekkhandelwal1:
Both findings from the "previously missed" review are fixed in eaf7dcc1.

**A job named more than once lost half its verdict.** `_cia_results_from_json`
deduplicated on `job_id` and returned early, so a run that printed
`{"job_id": "x", "category": "hang"}` and later `{"job_id": "x", "confidence": 0.9}`
reported the category with no confidence — both halves present in one output,
the later one discarded. The walker now fills: an id already recorded keeps
whatever the earlier object answered, and only a half that object left `null`
is taken from a later one.

Fill rather than replace, deliberately — the earlier object is the one the log
already reports, and a later partial restatement is not a retraction of it, so
replacing would let a trailing `{"job_id": "x", "confidence": null}` erase a
confidence the run had actually measured. The row also keeps the position its
first mention earned (`results` is ordered and read in order), and the locator
branch is still walked through for an already-recorded job so a second mention
still contributes its children. Four tests, one per property, and each was
mutation-tested by reintroducing the defect.

**`configuration.md` described a scrubber that does not exist.** The
selector-reason bullet claimed paths and IP addresses were "always scrubbed
regardless of `--no-redact`". False in both directions: full mode writes the
reason verbatim, and `--no-redact` governs what is sent *to the model*, not
what is written to the log. An operator reading that bullet would have treated
a full-mode log as sanitised. The bullet now says what the code does, and the
claim is pinned by a test that writes a reason containing a path and an IP,
asserts the record holds it byte for byte, and asserts the doc no longer
carries the old wording — checked against the doc joined into one line, since
the claim was split across a wrap and a re-wrap must not slip it past.

`tests/chat/test_decision_log.py`: 65 → 70. Full `tests/chat`: 2248 passed,
122 skipped, 2 xfailed.

src/aorta/chat/decision_log.py
```diff
@@ -160,26 +189,269 @@ def _parse_tool_trace(entry: Any) -> tuple[str, str, str]:
...
+    object with an id is taken whole rather than merged with its parent's.
+
+    Validated to the same vocabulary the patterns accept, rather than taken on
+    trust -- ``{"job_id": 3}`` or ``{"confidence": "high"}`` is not a verdict
+    this record can carry, and writing it through would put a shape into the
+    log that nothing downstream expects.
+
+    **A job named more than once keeps its verdict, half by half**, which is
+    the rule :func:`_cia_results_from_text` already follows and this side did
+    not. Announce-then-finish is the ordinary shape of a progress log in JSON
```
amd-vivekag: **[Low]** Docstring points at `_cia_results_from_text`, which does not exist. Copilot flagged this too.

Fix:
```suggestion
    **A job named more than once keeps its verdict, half by half**, which is
    the rule :func:`_first_value` already follows on the text path and this
    side did not. Announce-then-finish is the ordinary shape of a progress log in JSON
```

amd-vivekag (commented):
## Review summary

**Verdict: ship**

Re-review of eaf7dcc. Last round's Medium is fixed correctly: a raw `build_report()` dump now keeps `gpu_race` / `0.82`, and `_cia_results_from_json` answers `None` so the line reader (or a later result object) can see the verdict. Half-by-half JSON merge is the right twin of `_first_value`, and the docs no longer promise a scrubber the log does not run.

**Prior comments**
- High (`nan-trap`) — Fixed (correct)
- Medium (verdict before id / `sort_keys` JSON) — Fixed (correct)
- Medium (`report.json` locator) — Fixed (correct)
- Low (`accepted` docs, `finally` cleanup) — Fixed (correct)

**Copilot (latest pass)**
- Confirmed: dangling `:func:`_cia_results_from_text`` (Low, below).
- Dismissed: per-half whole-output fill for one job — `_any_value` documents that trade-off on purpose.
- Dismissed: mixed JSON array of a full report plus a result row — no in-tree tool emits that, and declining the locator is the guard the last Medium required.

src/aorta/chat/decision_log.py
amd-vivekag: Resolved — a raw `build_report()` dump now keeps the verdict, and `_cia_results_from_json` returns `None` so the line reader can see it. Thanks.

src/aorta/chat/decision_log.py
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed in 34351414. `_cia_results_from_text` has never existed — the text path is `_cia_results` itself, reading spans through `_first_value`.

Reworded rather than dropped to plain text: the sentence is a claim that two reads follow one rule, and the link is how a reader checks it. So it now names the function that does implement it, and the paragraph below still cites `_first_value`, which is where the rule actually lives.

```
    **A job named more than once keeps its verdict, half by half**, which is
    the rule the text path in :func:`_cia_results` already follows and this
    side did not.
```

This is worth a tripwire rather than a one-line fix, because the failure mode is exactly why it survived review: Sphinx renders an unresolved `:func:` as plain text, so the docstring read correctly to everyone reading the source and pointed at nothing to anyone reading the docs. `test_every_func_reference_in_the_module_resolves` now resolves every same-module `:func:`_name`` in `decision_log.py` against the module's own `def`s. Scoped to same-module references on purpose — a dotted target cannot be resolved without importing it, and guessing at that is how a link checker starts reporting noise. Mutation-tested by restoring the dangling name and confirming it goes red.

`tests/chat/test_decision_log.py`: 70 → 71.

vivekkhandelwal1: Fixed in 34351414. `_cia_results_from_text` has never existed — the text path is `_cia_results` itself, reading spans through `_first_value`.

Reworded rather than dropped to plain text: the sentence is a claim that two reads follow one rule, and the link is how a reader checks it. So it now names the function that does implement it, and the paragraph below still cites `_first_value`, which is where the rule actually lives.

```
    **A job named more than once keeps its verdict, half by half**, which is
    the rule the text path in :func:`_cia_results` already follows and this
    side did not.
```

This is worth a tripwire rather than a one-line fix, because the failure mode is exactly why it survived review: Sphinx renders an unresolved `:func:` as plain text, so the docstring read correctly to everyone reading the source and pointed at nothing to anyone reading the docs. `test_every_func_reference_in_the_module_resolves` now resolves every same-module `:func:`_name`` in `decision_log.py` against the module's own `def`s. Scoped to same-module references on purpose — a dotted target cannot be resolved without importing it, and guessing at that is how a link checker starts reporting noise. Mutation-tested by restoring the dangling name and confirming it goes red.

`tests/chat/test_decision_log.py`: 70 → 71.

amd-vivekag (approved):

src/aorta/chat/decision_log.py
amd-vivekag: Resolved — `_cia_results` exists, and the new test pins every same-module `:func:`_name``. Thanks.

src/aorta/chat/decision_log.py
```diff
@@ -160,26 +189,269 @@ def _parse_tool_trace(entry: Any) -> tuple[str, str, str]:
...
+        known = rows.get(raw_id)
+        if known is not None:
+            # Fill, never replace: the earlier object's answer is the reported
+            # one, and only a half it left empty is still open.
+            for half, value in verdict.items():
+                if known[half] is None:
+                    known[half] = value
+            return
+        rows[raw_id] = row = {"job_id": raw_id, **verdict}
+        results.append(row)
```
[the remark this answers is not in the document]
vivekkhandelwal1: Confirmed. After taking a row, the walker returned without visiting the object's values, so `{"job_id": "cia-parent", ..., "children": [{"job_id": "cia-child", ...}]}` came back with only the parent. Because the JSON pass had claimed the output, the line reader never ran to recover the child. #507 merged before this could land, so the fix is in the follow-up #524 (515f5a5c). The walk now descends into an ID-bearing object after taking it, on both the first-row branch and the repeated-id fill branch. Each nested object is still read only for its own keys, so a parent's verdict can't reach a child. Tripwires: `test_a_job_nested_inside_a_job_gets_its_own_row`, `test_a_repeated_jobs_nested_jobs_are_read_too`, plus the narrowness test `test_a_nested_job_without_a_verdict_does_not_take_its_parents`. 3 of 3 mutations are caught.

